PR #3, opened 2021-07-13
Usf scheduling

Saeid-Solei:
This is the initial scheduling algorithm commit!!!


paulbourelly999:
@Saeid-Solei Change class names without version number and change package name scheduling-service.

dan-du-car:
- No Camel case for file names, variable names, class name, updating them to lower cases words separated by underscore. eg: csCtseClassD_v1_0.cpp rename to cs_ctse_class.cpp.

- Split one header file to multiple smaller header files to only include relevant information in each header file.

paulbourelly999 (changes requested):
File structure should be something like scheduling-service/include/<header files>
                                                                                             /src/<cpp files>
                                                                                            /test/test_cases.cpp

PR #6, opened 2021-07-26
Fix the core dump issue

dan-du-car:
# PR Details
## Description
After intensive testing with publishing data  at higher rate (tested both 100HZ and 10HZ), it turns out the delete operator is problematic.

## Related Issue
https://github.com/usdot-fhwa-stol/carma-streets/issues/7

## Motivation and Context
Kafka clients support

## How Has This Been Tested?
Local integration testing

## Types of changes

## Checklist:

### DIFF
--- a/kafka_clients/src/kafka_consumer_worker.cpp
+++ b/kafka_clients/src/kafka_consumer_worker.cpp
@@ -105,13 +105,12 @@ namespace kafka_clients
 
     void kafka_consumer_worker::stop()
     {        
-        _consumer->close();
-
         if (_topic) {
             delete _topic;
         }
 
         if (_consumer) {
+             _consumer->close();
             delete _consumer;
         }
         _run=false;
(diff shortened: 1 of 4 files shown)


kafka_clients/src/kafka_consumer_worker.cpp
```diff
@@ -139,7 +138,7 @@ namespace kafka_clients
         RdKafka::Message *msg = NULL;
         msg = _consumer->consume(timeout_ms);      
         const char* msg_str= msg_consume(msg, NULL);
-        delete msg;
+        free(msg);
```
paulbourelly999, line 141: @dan-du-car I think this should stay as delete. In the example rdkakfa_complex_consumer_example.cpp it is delete. Also if you look at the consume method : 
```
RdKafka::Message *RdKafka::KafkaConsumerImpl::consume (int timeout_ms) {
  rd_kafka_message_t *rkmessage;

  rkmessage = rd_kafka_consumer_poll(this->rk_, timeout_ms);

  if (!rkmessage)
    return new RdKafka::MessageImpl(NULL, RdKafka::ERR__TIMED_OUT);

  return new RdKafka::MessageImpl(rkmessage);

}
```
It uses new to allocate the memory for the message so according to [https://www.geeksforgeeks.org/g-fact-30/](delete vs free) this should be delete right?

paulbourelly999 (changes requested):
See kafka_consumer_worker.cpp comment. Other than that everything looks good so once this is addressed I can approve the PR

paulbourelly999 (approved):
Looks good. @dan-du-car Did you create an issue with librdkafka and should we link it to this somehow?

Issue #13, opened 2021-09-27
Message correlation service 


dan-du-car:
### Types of Issue

- [ ] Anomaly report (something appears to not work correctly)
- [x] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary
Design Message Correlation Library/Service including how we will create/use dependent libraries and how it will interface with the Scheduling Service. Some design questions that need to be answered:

What form will the Message Correlation part of CARMA-Streets take ( Separate Service/Docker Image, Library dependency for other Services to use)?

How will we persist/publish correlated updates ( In memory DB, Kafka)?

How will we correlate BSM, MobilityOperation and MobilityPath ( static vehicle ID, BSM ID, sequence ID)

What dependent libraries are necessary ( lanelet2 and extensions, Map Library )

### Expected behavior
Design and initial implementation for message service to satisfy those answer the above questions.
### Actual behavior

N/A
### Steps to reproduce the actual behavior
N/A

dan-du-car:
It can be closed as it is addressed by the above PRs

PR #14, opened 2021-09-27
Message service 

dan-du-car:
# PR Details
## Related Issue
https://github.com/usdot-fhwa-stol/carma-streets/issues/13

## Motivation and Context
Cooperative stop 

## How Has This Been Tested?
unit test and local integration test

## Types of changes

## Checklist:

### DIFF
--- a/kafka_clients/include/kafka_client.h
+++ b/kafka_clients/include/kafka_client.h
@@ -1,3 +1,4 @@
+#pragma once
 #include "kafka_producer_worker.h"
 #include "kafka_consumer_worker.h"
 #include <boost/thread.hpp>
--- a/message_services/include/mobilityoperation_worker.h
+++ b/message_services/include/mobilityoperation_worker.h
@@ -1,0 +1,26 @@
+#pragma once
+#include <iostream>
+#include "mobilityoperation.h"
+#include "kafka_client.h"
+#include "base_worker.h"
+
+namespace message_services
+{
+    namespace workers
+    {
+        class mobilityoperation_worker : public base_worker
+        {
+        private:
+            std::deque<message_services::models::mobilityoperation> mobilityoperation_v;
+
+        public:
+            mobilityoperation_worker();
+            ~mobilityoperation_worker();
+
+            std::mutex worker_mtx;
+            std::deque<models::mobilityoperation>& get_curr_list();
+            void process_incoming_msg(const std::string json_str);
+            void pop_cur_element_from_list(long element_position);
+        };
+    }
+}
--- a/message_services/src/models/bsm.cpp
+++ b/message_services/src/models/bsm.cpp
@@ -1,0 +1,193 @@
+#include <iostream>
+#include <vector>
+#include <string.h>
+#include "rapidjson/writer.h"
+#include "rapidjson/stringbuffer.h"
+#include "bsm.h"
+
+namespace message_services
+{
+
+    namespace models
+    {
+
+        bsm::bsm() : core_data(), header()
+        {
+        }
+
+        bsm::~bsm()
+        {
+        }
+
+        void bsm::fromJsonObject(const rapidjson::Value &obj)
+        {
+            if (obj.IsObject())
+            {
+                for (rapidjson::Value::ConstMemberIterator obj_itr = obj.MemberBegin(); obj_itr != obj.MemberEnd(); obj_itr++)
+                {
+                    if (obj_itr->value.IsObject() && boost::iequals(std::string(obj_itr->name.GetString()), std::string("header")))
+                    {
+                        header_t header;
+
+                        for (rapidjson::Value::ConstMemberIterator header_itr = obj_itr->value.MemberBegin(); header_itr != obj_itr->value.MemberEnd(); header_itr++)
+                        {
+                            if (std::string(header_itr->name.GetString()) == std::string("year"))
+                            {
+                                header.year = header_itr->value.GetUint64();
+                            }
+
+                            if (std::string(header_itr->name.GetString()) == std::string("timestamp"))
+                            {
+                                header.timestamp = (header_itr->value.GetUint64());
+                            }
+
+                            if (std::string(header_itr->name.GetString()) == std::string("sec_mark"))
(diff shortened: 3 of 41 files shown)


message_services/src/services/vehicle_status_intent_service.cpp
```diff
@@ -0,0 +1,389 @@
...
+                for (auto mp_v_item : mp_w_ptr->get_curr_list())
+                {
+                    std::string mp_vehicle_id = mp_v_item.getHeader().sender_id;
+                    uint64_t mp_timestamp = mp_v_item.getHeader().timestamp;
+                    if (mo_vehicle_id == mp_vehicle_id && std::abs((long)mo_timestamp - (long)mp_timestamp) < 100)
+                    {
+                        mp_ptr->setHeader(mp_v_item.getHeader());
+                        mp_ptr->setTrajectory(mp_v_item.getTrajectory());
+
+                        mp_w_ptr->pop_cur_element_from_list(mp_pos);
```
paulbourelly999, line 289: Is this correct. It seems possible that the mp_v_item could be an index other that 0 since we are looping through. If you get values from one index and then pop zero doesn't that cause an issue

message_services/src/services/vehicle_status_intent_service.cpp
```diff
@@ -0,0 +1,389 @@
...
+                    }
+                }
+
+                //checking timestamp for this bsm_id
+                long bsm_pos = 0;
+                for (auto bsm_v_item : bsm_w_ptr->get_curr_list())
+                {
+                    std::string bsm_id = bsm_v_item.getCore_data().temprary_id;
+                    uint64_t bsm_timestamp = bsm_v_item.getHeader().timestamp;
+                    if (mo_bsm_id == bsm_id && std::abs((long)mo_timestamp - (long)bsm_timestamp) < 100)
```
paulbourelly999, line 300: we need to make sure that bsm_v_item has the index of bsm_pos right? Otherwise we could pop the wrong value

message_services/src/services/vehicle_status_intent_service.cpp
```diff
@@ -0,0 +1,389 @@
...
+        void vehicle_status_intent_service::identify_latest_mapping_bsm_mo_mp(std::shared_ptr<workers::bsm_worker> bsm_w_ptr,
+                                                                              std::shared_ptr<workers::mobilityoperation_worker> mo_w_ptr,
+                                                                              std::shared_ptr<workers::mobilitypath_worker> mp_w_ptr,
+                                                                              std::shared_ptr<models::bsm> bsm_ptr,
+                                                                              std::shared_ptr<models::mobilityoperation> mo_ptr,
+                                                                              std::shared_ptr<models::mobilitypath> mp_ptr)
+        {
+            //Identify which vehicle by vehicle id to find mobililityoperation
+            std::deque<models::mobilityoperation>::iterator itr;
+            long mo_pos = 0;
```
paulbourelly999, line 266: Will this always be zero. If so what is the point of defining it. If not where is it incremented?

paulbourelly999 (approved):
Any comments left in the vehilce_status_intent_service can be left unadress since this service exceeds that scope of this issue

message_services/src/services/vehicle_status_intent_service.cpp
dan-du-car: Addressed

message_services/src/services/vehicle_status_intent_service.cpp
dan-du-car: yeah, by comparing the bsm_v_item's bsm_id and timestamp, if return true, we know that this bsm_v_item exit. The bsm_pos is to keep track of the position of the bsm_v_item. so it should be the right value.

message_services/include/mobilityoperation_worker.h
```diff
@@ -0,0 +1,26 @@
...
+        {
+        private:
+            std::deque<message_services::models::mobilityoperation> mobilityoperation_v;
+
+        public:
+            mobilityoperation_worker();
+            ~mobilityoperation_worker();
+
+            std::mutex worker_mtx;
+            std::deque<models::mobilityoperation>& get_curr_list();
```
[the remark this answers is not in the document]
dan-du-car: Addressed

message_services/src/models/bsm.cpp
```diff
@@ -0,0 +1,193 @@
...
+                std::cout << "BSM model: asJsonObject error.: " << ex.what() << std::endl;
+                return false;
+            }
+        }
+
+        std::ostream &operator<<(std::ostream &out, bsm &bsm_obj)
+        {
+            std::cout << std::fixed;
+            std::cout << std::setprecision(6);
+            std::cout << "bsm model: year =  " << bsm_obj.header.year
```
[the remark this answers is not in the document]
dan-du-car: As discussed, any std::cout inside the operator method is an override to std::out. It does not apply to spd logger.

message_services/src/services/vehicle_status_intent_service.cpp
dan-du-car: mo_pos is now increased.

message_services/src/models/bsm.cpp
```diff
@@ -0,0 +1,193 @@
...
+                writer->String(this->getCore_data().transmission.c_str());
+                writer->String("speed");
+                writer->String(std::to_string(this->getCore_data().speed).c_str());
+                writer->EndObject();
+                writer->EndObject();
+                return true;
+            }
+            catch (std::exception &ex)
+            {
+                std::cout << "BSM model: asJsonObject error.: " << ex.what() << std::endl;
```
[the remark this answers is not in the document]
dan-du-car: updated

kafka_clients/include/kafka_client.h
```diff
@@ -1,3 +1,4 @@
+#pragma once
```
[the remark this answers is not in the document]
dan-du-car: updated

itabrl (approved):

Issue #12, opened 2021-09-24
Create Map Library


paulbourelly999:
### Types of Issue

- [ ] Anomaly report (something appears to not work correctly)
- [ x ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

Create a library that is includes lanelet2 and the autware.io extensions to consume a lanelet2 osm file. This component will be responsible for translating long,lat location information into lane specific information and providing calculation about lane conflicts, distance to end of lane, and regulatory elements. 
### Steps to reproduce the actual behavior

1. Do this
1. Then do this...

dan-du-car:
This issue can be closed per the above PR merged

PR #16, opened 2021-10-07
Intersection Model

paulbourelly999:
# PR Details
## Description
This library will read and lanelet2 osm map and take in a configurable bounding box to mark the intersection being managed. The library will store the appropriate lanelet2 objects inside this bound box including entry lanes into the intersection, linking lanes through the intersection and any regulatory elements (stop bars, stop signs, traffic lights, speed limits) inside the intersection. The intersection_model had methods to translate GPS location data into lane specific information like lanelet id and distance to the end of the lane for services like the message_service to provide lane. The intersection_model will also return the intersection lanelet2 objects to services that need them to make sense of the lane specific updates the message_service provides.

I have added methods for the library to load a lanelet2 osm map and create the required projector. I have also included methods to return Lanelet objects given GPS location. I have implemented some stub methods and fields to return all link and entry Lanelet objects for a intersection given a bounding box to mark included objects

## Related Issue
#12 

## How Has This Been Tested?
Unit Testing 

## Types of changes

- [ x ] New feature (non-breaking change that adds functionality)

## Checklist:

- [ x ] I have read the **CONTRIBUTING** document.
- [ x ] I have added tests to cover my changes.


intersection_model/include/intersection_model.h
```diff
@@ -0,0 +1,66 @@
...
+         * @param filename path to lanelet2 osm map
+         */
+        intersection_model( std::string filename );
+        
+        /**
+         * Getter for lanelet2 map.
+         * 
+         * @return lanelet map object
+         */ 
+        lanelet::LaneletMapPtr get_lanelet2_map();
```
dan-du-car, line 29: map object shall be private and only be manipulated by this model itself.

intersection_model/include/intersection_model.h
```diff
@@ -0,0 +1,66 @@
...
+         * @param filename path to lanelet2 osm map
+         */
+        void read_lanelet2_map( std::string filename );
+
+        /**
+         * Method to set intersection
+         * 
+         * @param bounds bounding box for intersection
+         */
+        void set_intersection( lanelet::BoundingBox2d bounds );
```
dan-du-car, line 64: Will need a design for this. Please remove this for now.

intersection_model/include/intersection_model.h
paulbourelly999: When we spoke with Mike and Kyle I remember the consensus being that our library doesn't need to obscure lanelet2 objects since then we would just be rewriting functionality that already exists in the lanelet2 library. I figured if we still return lanelet2 objects and at some point need there functionality then still have access to them. I understand this kind of goes against some of the normal practices for designing a model. If we want to talk about this in more detail I am willing to.

PR #19, opened 2021-10-14
Intersection Model: Add generated server code and update intersection library

dan-du-car:
# PR Details
## Description
Add generated server code and update intersection library.

## Related Issue
https://github.com/usdot-fhwa-stol/carma-streets/issues/12

## How Has This Been Tested?
Local integration testing
Unit tests

## Types of changes

## Checklist:


intersection_model/readme.md
```diff
@@ -6,7 +6,7 @@ This library will read and lanelet2 osm map and take in a configurable bounding
 ## Steps to Build intersection_model locally
 
 ```
-apt-get update  && apt-get install -y cmake libboost1.65-all-dev git autotools-dev automake sqlite3 libsqlite3-dev curl libpugixml-dev libgeographic-dev
+apt-get update  && apt-get install -y cmake libboost1.65-all-dev git autotools-dev automake sqlite3 libsqlite3-dev curl libpugixml-dev libgeographic-dev qtbase5-dev qtbase5-dev-tools libqhttpengine-dev
```
paulbourelly999, line 9: missing libssl-dev

docker-compose.yml
```diff
@@ -84,6 +84,17 @@ services:
       DOCKER_HOST_IP:  ${DOCKER_HOST_IP}
     volumes: 
       - ./message_services/manifest.json:/home/carma-streets/message_services/manifest.json
+  intersection_model:
```
paulbourelly999, line 87: Why do we have an image for the intersection_model. This is just suppose to be a library that the message service uses

intersection_model/Dockerfile
```diff
@@ -128,7 +127,7 @@ RUN source /opt/ros/melodic/setup.bash && source /home/carma-streets/carma_lanel
 RUN source /opt/ros/melodic/setup.bash && source /home/carma-streets/carma_lanelet2/install/setup.bash &&  make
 
 # Source /opt/ros/melodic/setup.bash and /home/carma-streets/carma_lanelet2/install/setup.bash
-ENTRYPOINT [ "/home/carma-streets/intersection_model/entrypoint.sh" ]
```
paulbourelly999, line 131: In order to run the intersection_model executable in a shell that has the ros medolic setup.bash sourced and the lanelet2 install setup.bash sourced this line was necessary. Why is it no longer necessary? Have you tried running the docker image to see if it can still find the lanelet2 depedencies

intersection_model/include/intersection_model.h
```diff
@@ -1,59 +1,223 @@
...
+            uint32_t get_speed_limit_by_lanelet(const lanelet::ConstLanelet& subj_lanelet);
+
+            /**
+             * @brief When an intersection determines whether there are conflict for vehicles entering the intersection at the same time,
+             * it checks all the vehicles entering from different directions. If other vehicles entering or come behind from the same entry lane,
+             * This vehicle shall ignore the conflicts associate with this entry lane. It is assuming that the subject vehicle is the first vehicle
+             * at the entering lane and is waiting to enter the intersection. 
+             * @return A vector of link lanelet ids
+             **/
+            std::vector<int> get_shared_entry_link_lanelet_ids(int64_t subject_link_lanelet_id);
```
paulbourelly999, line 165: Not sure I understand what this method is for

intersection_model/intersection_model_v1.yaml
```diff
@@ -0,0 +1,218 @@
...
+                              {"id": 233, "speed_limit": 12.2},
+                              {"id": 234, "speed_limit": 12.2},
+                              {"id": 227, "speed_limit": 12.2},
+                              {"id": 232 "speed_limit": 12.2},
+                            ]
+                          },
+                          "name": "West Intersection"
+                        }'
+        default:
+          description: Unexpected error
```
paulbourelly999, line 55: I think the correct way to reflect errors would be with http status code like :
`responses:
        '201':
            description: Created   
        '400':
            description: Bad Request  `

intersection_model/src/server/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+        libgcc  \
+        libstdc++  \
+        qt5-qtbase \
+        openssl
+
+WORKDIR /usr/server
+COPY --from=build /usr/server/build/src/cpp-qt-qhttpengine-server ./build/src/
+COPY --from=build /usr/server/external/ ./external
+EXPOSE 8080/tcp
+ENTRYPOINT ["/usr/server/build/src/cpp-qt-qhttpengine-server"]
```
paulbourelly999, line 29: Why does the REST server have a separate Dockerfile. It shouldn't be a separate service right?

intersection_model/src/server/Makefile
```diff
@@ -0,0 +1,29 @@
+BUILD_DIR=build
```
paulbourelly999, line 1: Should this be committed??

intersection_model/src/server/external/lib/cmake/qhttpengine/qhttpengineConfig.cmake
```diff
@@ -0,0 +1,99 @@
+# Generated by CMake
+
```
paulbourelly999, line 2: This should no be committed

intersection_model/src/server/external/lib/cmake/qhttpengine/qhttpengineConfig-noconfig.cmake
```diff
@@ -0,0 +1,19 @@
+#----------------------------------------------------------------
```
paulbourelly999, line 1: This should not be committed

intersection_model/src/server/external/lib/cmake/qhttpengine/qhttpengineConfigVersion.cmake
```diff
@@ -0,0 +1,70 @@
+# This is a basic version file for the Config-mode of find_package().
```
paulbourelly999, line 1: This should not be committed

intersection_model/src/server/external/lib/libqhttpengine.so
```diff
@@ -0,0 +1 @@
+libqhttpengine.so.1
```
paulbourelly999, line 1: This should not be committed right ?

intersection_model/src/server/external/lib/libqhttpengine.so.1
```diff
@@ -0,0 +1 @@
+libqhttpengine.so.1.0.1
```
paulbourelly999, line 1: This should not be committed right

intersection_model/test/test_intersection_model.cpp
```diff
@@ -1,17 +1,157 @@
...
-    intersection_model::intersection_model *model = new intersection_model::intersection_model("../../osm/AOI_1_TFHRC_faster_pretty.osm");
+    intersection_model::intersection_model *model = new intersection_model::intersection_model("../vector_map.osm");
     lanelet::LaneletMapPtr map = model->get_lanelet2_map();
     ASSERT_TRUE(map->laneletLayer.size() > 0);
+}
+
+TEST(intersection_model_test, get_lanelet)
+{
+    intersection_model::intersection_model *model = new intersection_model::intersection_model("../vector_map.osm");
+    lanelet::Lanelet lane = model->get_lanelet(38.9551104, -77.1498214);
```
paulbourelly999, line 14: I was having trouble getting this method to return the correct lanelet when using a point inside the lanelet. I would often get the adjacent lanelet. Did you change the logic for the get_lanelet at all.

docker-compose.yml
dan-du-car: It is itself also a REST service.

intersection_model/Dockerfile
dan-du-car: Line 128 has already sourced the setup.bash.

intersection_model/include/intersection_model.h
dan-du-car: It gets the list of link lanelets that vehicles enter from the same entry lane.

intersection_model/readme.md
dan-du-car: Addressed

intersection_model/src/server/external/lib/cmake/qhttpengine/qhttpengineConfigVersion.cmake
dan-du-car: removed

paulbourelly999 (changes requested):

intersection_model/Dockerfile
dan-du-car: yes, I can build docker image and run as container without this ENTRYPOINT line.

intersection_model/src/server/src/handlers/OAIApiRouter.h
```diff
@@ -0,0 +1,114 @@
...
+    virtual void process(QHttpEngine::Socket *socket, const QString &path){
+        Q_UNUSED(path);
+
+        // If the slot requires all data to be received, check to see if this is
+        // already the case, otherwise, wait until the rest of it arrives
+        if (socket->bytesAvailable() >= socket->contentLength()) {
+            emit requestReceived(socket);
+        } else {
+            // connect(socket, &Socket::readChannelFinished, [this, socket, m]() {
+            //     emit requestReceived(socket);
```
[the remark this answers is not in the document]
dan-du-car: It generated errors. Yeah, it is generated code, and I had to comment out to get it to work.

intersection_model/Dockerfile
paulbourelly999: So if you tell me it works without that is fine, but I know for a fact that dockerfiles can have both entrypoint and CMD at the same time. CARMA-Platform has entrypoint in its base images and them CMD in its image. Also I tested this dockerfile and it did build. But if you say the docker image runs without it then that is fine to. We probably want to remove the scripts then if it runs without this. My experience was that the docker image would fail to run if we did not have this line and the entrypoint.sh and init-env.sh scripts.

paulbourelly999: If it builds and runs without we need to remove the entrypoint and init-env.sh scripts since they are no longer necessary.

PR #28, opened 2021-11-03
Add length for the lanelet

dan-du-car:
# PR Details
## Description
Add length for  each intersection lanelet

## Related Issue
N/A

## Motivation and Context
TSMO use case one

## How Has This Been Tested?
Local integration testing

## Types of changes

## Checklist:

### DIFF
--- a/intersection_model/include/intersection_model.h
+++ b/intersection_model/include/intersection_model.h
@@ -51,7 +51,8 @@ namespace intersection_model
     struct lanelet_info_t
     {
         int64_t id           = 0;    //Uniquely identify the lanelet id within an OSM file
-        uint32_t speed_limit = 0.0;  //Unit of measure MPH
+        uint32_t speed_limit = 0;    //Unit of measure MPH
+        double length        = 0.0;  //Unit of meter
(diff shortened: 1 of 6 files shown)


intersection_model/include/intersection_model.h
```diff
@@ -51,7 +51,8 @@ namespace intersection_model
     struct lanelet_info_t
     {
         int64_t id           = 0;    //Uniquely identify the lanelet id within an OSM file
-        uint32_t speed_limit = 0.0;  //Unit of measure MPH
+        uint32_t speed_limit = 0;    //Unit of measure MPH
```
Saeid-Solei, line 54: Can we change the unit to meter/sec? if not, I have to make the conversion in the scheduling service.

dan-du-car: It is the original value from OSM map. If you look at the .osm file, it has MPH along with the speed limit value. intersection model is only responsible reading and publish the data from map, and minimize any modification if possible.

Saeid-Solei (approved):

PR #53, opened 2021-12-09
adding an expiration time for the vehicle status and intent update

Saeid-Solei:
# PR Details
## Description

An expiration time has been added to the received vehicle status and intent update. If the vehicle update is older than a configurable time duration, the scheduling service will not include the vehicle in the schedule. 

## Types of changes

## Checklist:

### DIFF
--- a/scheduling_service/include/configuration.h
+++ b/scheduling_service/include/configuration.h
@@ -19,6 +19,11 @@ class configuration{
 		/* the time point when the current schedule started */
 		double cur_schedule_start_time;
 
+		/* If the last update of the vehicle is update_expiration_delta seconds older than the current time, the vehicle will not be included in the schedule
+		*  Unit: second
+		 */
+		double update_expiration_delta = 300;
+
 	public:
 
 		/* get the scheduling time interval */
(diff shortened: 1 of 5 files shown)


scheduling_service/src/scheduling.cpp
```diff
@@ -11,157 +11,164 @@ scheduling::scheduling(unordered_map<string, vehicle> list_veh, set<string> list
...
-				list_veh_confirmation.erase(veh_id.back());
-			}
-			/* if CARMA Streets did not receive the confirmation from the vehicle updates yet, it will consider the vehicle as a DV for now */
-			else{
-				state[state.size() - 1] = "DV";
-				et[et.size() - 1] = config.get_curSchedulingT();
-				access[access.size() - 1] = true;
-			}
-		}
+		if (config.get_curSchedulingT() - element.second.get_curTime() <= config.get_expDelta()){
```
dan-du-car, line 14: Could you add comment before this line to explain what this if statement with expDelta() works?

scheduling_service/include/configuration.h
```diff
@@ -19,6 +19,11 @@ class configuration{
 		/* the time point when the current schedule started */
 		double cur_schedule_start_time;
 
+		/* If the last update of the vehicle is update_expiration_delta seconds older than the current time, the vehicle will not be included in the schedule
+		*  Unit: second
+		 */
+		double update_expiration_delta = 300;
```
dan-du-car, line 25: Can we update this value from the manifest.json so that we do not need to rebuild the docker image when we adjust this expiration?

Issue #55, opened 2021-12-13
scheduling_service integration testing with one vehicle issues


dan-du-car:
### Types of Issue

- [x] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

Vehicle not able to stop at the intersection or the scheduling plan is not updated when the vehicle is stopped at the stop bar.
The schedule plan transit from null directly to access = 1, the vehicle does not stop. The vehicle only stops when it receives the updated schedule with access = 0, otherwise it will cruise as normal.
### Expected behavior
The scheduling plan shall be updated when the vehicle stop at the stop bar
### Steps to reproduce the actual behavior

NA

dan-du-car:
Issue can be closed due to merged PRs.

PR #60, opened 2021-12-15
adding a check before processing the status and intent message

Saeid-Solei:
# PR Details
## Description

Add a check to the scheduling_service to make sure the received status and intent message has a valid cur_lane_id.

## Related Issue
https://github.com/usdot-fhwa-stol/carma-streets/issues/55

## Types of changes

## Checklist:

### DIFF
--- a/scheduling_service/src/scheduling.cpp
+++ b/scheduling_service/src/scheduling.cpp
@@ -6,7 +6,7 @@ using namespace std;
 
 
 /* */
-scheduling::scheduling(unordered_map<string, vehicle> list_veh, set<string> list_veh_confirmation, intersection_client& localmap, configuration& config){
+scheduling::scheduling(unordered_map<string, vehicle> list_veh, set<string>& list_veh_confirmation, intersection_client& localmap, configuration& config, set<string>& list_veh_removal){
 
 	index_EVs.resize(localmap.get_laneIdEntry().size());
 	for (auto& element : list_veh){
@@ -181,6 +181,8 @@ scheduling::scheduling(unordered_map<string, vehicle> list_veh, set<string> list
 			if (list_veh_confirmation.find(element.first) != list_veh_confirmation.end()){
 				list_veh_confirmation.erase(element.first);
 			}
+			spdlog::info("Vehicle {0} is not added to the schedule as its update is more than {1} seconds old!", element.first, config.get_expDelta());
+			list_veh_removal.insert(element.first);
 		}
 	}
 
--- a/scheduling_service/src/vehicle.cpp
+++ b/scheduling_service/src/vehicle.cpp
@@ -291,10 +291,10 @@ bool vehicle::message_hasError(const Document& message, intersection_client& loc
 			return false;
 		}
 	} 
-	/* if it is not the first time processing the vehicle update, the vehicle cur_lane_id must be the same as entry_lane_id, link_lane_id, or depart_lane_id . */
+	/* if a vehicle update has been succesfully processed before, the vehicle cur_lane_id must be the same as entry_lane_id, link_lane_id, or depart_lane_id . */
 	else{
 		string veh_id = message["payload"]["v_id"].GetString();
-		if (to_string(message["payload"]["cur_lane_id"].GetInt()) != entryLane_id && to_string(message["payload"]["cur_lane_id"].GetInt()) != link_id && to_string(message["payload"]["cur_lane_id"].GetInt()) != exitLane_id){
+		if (to_string(message["payload"]["cur_lane_id"].GetInt()) != entryLane_id && to_string(message["payload"]["cur_lane_id"].GetInt()) != link_id && to_string(message["payload"]["cur_lane_id"].GetSInt()) != exitLane_id){
 			
 			spdlog::critical("the current lane id of Vehicle {0} is not correct! entry_lane_id: {1}, link_lane_id: {2}, exit_lane_id: {3}, cur_lane_id: {4}", veh_id, entryLane_id, link_id, exitLane_id, to_string(message["payload"]["cur_lane_id"].GetInt()));
 			return true;
--- a/scheduling_service/src/vehicle.cpp
+++ b/scheduling_service/src/vehicle.cpp
@@ -15,22 +15,22 @@ void vehicle::update(const rapidjson::Document& message, intersection_client& lo
 
 	if (message["metadata"].HasMember("timestamp") && (double)message["metadata"]["timestamp"].GetInt64() / 1000.0 >= timestamp){
 		
-		/* the unit of the received speed from the message is centimeter per second
+		/* the unit of the received speed from the message is 0.02 of meter per second
 		*  the unit of the speed defined in the vehicle class is meter per second. 
 		*/
 		if (message["payload"].HasMember("cur_speed")){
-			if (message["payload"]["cur_speed"].GetDouble() / 100.0 > config.get_maxValidSpeed() && id != "" && message["payload"].HasMember("cur_ds") && message["payload"].HasMember("cur_lane_id")){
+			if (message["payload"]["cur_speed"].GetDouble() * 0.02 > config.get_maxValidSpeed() && id != "" && message["payload"].HasMember("cur_ds") && message["payload"].HasMember("cur_lane_id")){
 				if (to_string(message["payload"]["cur_lane_id"].GetInt()) == lane_id){
 					speed = (distance - message["payload"]["cur_ds"].GetDouble()) / (((double)message["metadata"]["timestamp"].GetInt64() / 1000.0) - timestamp);
 				}
 				else{
 					speed = (distance + (localmap.get_laneLength(to_string(message["payload"]["cur_lane_id"].GetInt())) - message["payload"]["cur_ds"].GetDouble())) / (((double)message["metadata"]["timestamp"].GetInt64() / 1000.0) - timestamp);
 				}
-				double invalid_speed = message["payload"]["cur_speed"].GetDouble() / 100.0;
+				double invalid_speed = message["payload"]["cur_speed"].GetDouble() * 0.02;
 				spdlog::info("Invalid speed in the message received from {0}: {1}", veh_id, invalid_speed);
 			}
 			else{
-				speed = message["payload"]["cur_speed"].GetDouble() / 100.0;
+				speed = message["payload"]["cur_speed"].GetDouble() * 0.02;
 				spdlog::info("valid speed in the message received from {0}: {1}", veh_id, speed);
 			}
 		} else{
--- a/scheduling_service/src/vehicle.cpp
+++ b/scheduling_service/src/vehicle.cpp
@@ -19,7 +19,20 @@ void vehicle::update(const rapidjson::Document& message, intersection_client& lo
 		*  the unit of the speed defined in the vehicle class is meter per second. 
 		*/
 		if (message["payload"].HasMember("cur_speed")){
-			speed = message["payload"]["cur_speed"].GetDouble() / 100.0;
+			if (message["payload"]["cur_speed"].GetDouble() / 100.0 > config.get_maxValidSpeed() && id != "" && message["payload"].HasMember("cur_ds") && message["payload"].HasMember("cur_lane_id")){
+				if (to_string(message["payload"]["cur_lane_id"].GetInt()) == lane_id){
+					speed = (distance - message["payload"]["cur_ds"].GetDouble()) / (((double)message["metadata"]["timestamp"].GetInt64() / 1000.0) - timestamp);
+				}
+				else{
+					speed = (distance + (localmap.get_laneLength(to_string(message["payload"]["cur_lane_id"].GetInt())) - message["payload"]["cur_ds"].GetDouble())) / (((double)message["metadata"]["timestamp"].GetInt64() / 1000.0) - timestamp);
+				}
+				double invalid_speed = message["payload"]["cur_speed"].GetDouble() / 100.0;
+				spdlog::info("Invalid speed in the message received from {0}: {1}", veh_id, invalid_speed);
+			}
+			else{
+				speed = message["payload"]["cur_speed"].GetDouble() / 100.0;
+				spdlog::info("valid speed in the message received from {0}: {1}", veh_id, speed);
+			}
 		} else{
 			spdlog::critical("the current speed of Vehicle {0} is missing in the received update!", veh_id);
 		}
@@ -164,7 +177,7 @@ void vehicle::update(const rapidjson::Document& message, intersection_client& lo
 			} else if (state == "RDV"){
 				if (access == false){
 					lane_id = entryLane_id;
-					distance = 0.1;
+					// distance = 0.1;
 				} else{
 					if (lane_id != link_id){
 						lane_id = link_id;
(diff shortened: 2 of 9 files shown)


scheduling_service/src/vehicle.cpp
```diff
@@ -144,8 +150,10 @@ void vehicle::update(const rapidjson::Document& message, intersection_client& lo
 			/* vehicle state update */
 			if (state == "EV"){
 				if (lane_id != entryLane_id){
+					if (lane_id == link_id){
+						distance = 0.1;
```
dan-du-car, line 154: Why do we update the distance to 0.1 when the lane id is link id?

scheduling_service/src/vehicle.cpp
```diff
@@ -144,8 +150,10 @@ void vehicle::update(const rapidjson::Document& message, intersection_client& lo
...
+					}
 					lane_id = entryLane_id;
```
dan-du-car, line 156: `if (lane_id != entryLane_id)`

scheduling_service/src/vehicle.cpp
```diff
@@ -246,6 +254,67 @@ void vehicle::update(const rapidjson::Document& message, intersection_client& lo
...
+	if (message["payload"].HasMember("cur_lane_id")){
+		cur_lane_id = to_string(message["payload"]["cur_lane_id"].GetInt());
+	}
+	else{
+		spdlog::critical("the current lane id of Vehicle {0} is missing in the received update!", veh_id);
+		return true;
+	}
+		
+	if (message["payload"].HasMember("entry_lane_id")){
+		entry_lane_id = to_string(message["payload"]["entry_lane_id"].GetInt());
```
dan-du-car, line 286: If the vehicle exit or in link lane, do we still care about the entry_lane_id exist or not?

scheduling_service/src/vehicle.cpp
```diff
@@ -246,6 +254,67 @@ void vehicle::update(const rapidjson::Document& message, intersection_client& lo
...
+	}
+	else{
+		spdlog::critical("the entry lane id of Vehicle {0} is missing in the received update!", veh_id);
+		return true;
+	}
+
+	if (message["payload"].HasMember("dest_lane_id")){
```
dan-du-car, line 293: If the vehicle is still far away and have not enter the entry lane, it will not show dest_lane_id. This dest_lane_id only exit when the vehicle is closed to intersection. It is not mandatory.

scheduling_service/src/vehicle.cpp
```diff
@@ -246,6 +254,67 @@ void vehicle::update(const rapidjson::Document& message, intersection_client& lo
...
+		exit_lane_id = to_string(message["payload"]["dest_lane_id"].GetInt());
+	}
+	else{
+		spdlog::critical("the exit lane id of Vehicle {0} is missing in the received update!", veh_id);
```
dan-du-car, line 297: If the vehicle is still far away and have not enter the entry lane, it will not show dest_lane_id. This dest_lane_id only exit when the vehicle is closed to intersection. It is not mandatory.

scheduling_service/src/vehicle.cpp
```diff
@@ -246,6 +254,67 @@ void vehicle::update(const rapidjson::Document& message, intersection_client& lo
...
+		return true;
+	}
+	
+	if (message["payload"].HasMember("link_lane_id")){
+		link_lane_id = to_string(message["payload"]["link_lane_id"].GetInt());
+	}
+	else{
+		spdlog::critical("the link lane id of Vehicle {0} is missing in the received update!", veh_id);
```
dan-du-car, line 305: The same. If vehicle have not enter the entry_lane at all, the link lane id would not show. So it should only be warning and proceed.

scheduling_service/src/vehicle.cpp
```diff
@@ -104,6 +104,12 @@ void vehicle::update(const rapidjson::Document& message, intersection_client& lo
 			} else{
 				spdlog::critical("the entry lane id of Vehicle {0} is missing in the received update!", veh_id);
 			}
+
+			if (message["payload"].HasMember("dest_lane_id")){
+				exitLane_id = to_string(message["payload"]["dest_lane_id"].GetInt());
```
dan-du-car, line 109: Not sure why we need exitLane_id if it seems not used in the logic.

scheduling_service/src/vehicle.cpp
Saeid-Solei: if the vehicle is EV, it shouldn't enter the link before stopping and becoming RDV. If by mistake, the vehicle slightly enters the link and stops, the "distance" will be the distance from the end of the link lane. this will cause issues. In this case, we update the lane_id to entry_lane_id and distance to 0.1 to resolve the issue.

scheduling_service/src/vehicle.cpp
Saeid-Solei: well, it should always be in the message, isn't it?

scheduling_service/src/vehicle.cpp
Saeid-Solei: I agree that the dest_lane_id is not used. So we can remove it.

Saeid-Solei: but anyway, if there is no dest_lane_id, it means the vehicle is not sending the correct message according to the stop-controlled intersection requirements. Therefore, we just don't process the message.

scheduling_service/src/vehicle.cpp
Saeid-Solei: Right now, it gives a warning and does not use the update if the link_lane is missing. the link_id is necessary for processing the message.

scheduling_service/src/vehicle.cpp
dan-du-car: If the vehicle is in exit lane, the cur_lane_id equal to exit lane id. But the entry , exit, and link lane ids from the message is 0. Because we can only identify entry, link and exit lane ids when vehicle moves into entry lane, and link lane.

dan-du-car:
https://github.com/usdot-fhwa-stol/carma-streets/issues/55

scheduling_service/src/vehicle.cpp
dan-du-car: I think I understand this issue. Why value 0.1 not 0 or 0.1?

scheduling_service/src/vehicle.cpp
```diff
@@ -246,6 +255,56 @@ void vehicle::update(const rapidjson::Document& message, intersection_client& lo
...
+	else if (!message["payload"].HasMember("v_id")){
+		spdlog::critical("vehicle id is missing in the received status and intent update!");
+		return true;
+	}
+	else if (!message["payload"].HasMember("cur_lane_id") || message["payload"]["cur_lane_id"].GetInt() == 0){
+		string veh_id = message["payload"]["v_id"].GetString();
+		spdlog::critical("the current lane id of Vehicle {0} is missing in the received update!", veh_id);
+		return true;
+	}
+	/* if it is the first time processing the vehicle update, the entry_lane_id and link_lane_id must be valid. */
```
dan-du-car, line 275: It requires both entry_lane_id and link_lane_id to be valid in order to process the the vehicle update successfully.  The id is always "" (empty) until the process succeed.

dan-du-car: Maybe update to the above to help clarify why id = ""?

scheduling_service/src/vehicle.cpp
```diff
@@ -246,6 +255,56 @@ void vehicle::update(const rapidjson::Document& message, intersection_client& lo
...
+		string veh_id = message["payload"]["v_id"].GetString();
+		if (!message["payload"].HasMember("entry_lane_id") || message["payload"]["entry_lane_id"].GetInt() == 0){
+			spdlog::critical("the entry lane id of Vehicle {0} is missing in the received update!", veh_id);
+			return true;
+		}
+		else if (!message["payload"].HasMember("link_lane_id") || message["payload"]["link_lane_id"].GetInt() == 0){
+			spdlog::critical("the link lane id of Vehicle {0} is missing in the received update!", veh_id);
+			return true;
+		}
+		else if (message["payload"]["cur_lane_id"].GetInt() != message["payload"]["entry_lane_id"].GetInt() && message["payload"]["cur_lane_id"].GetInt() != message["payload"]["link_lane_id"].GetInt()){
```
dan-du-car, line 286: I do not think it can reach the last else if statement?

scheduling_service/src/vehicle.cpp
```diff
@@ -246,6 +255,56 @@ void vehicle::update(const rapidjson::Document& message, intersection_client& lo
...
+			spdlog::critical("Vehicle {0} is not in the communication area yet!", veh_id);
+			return true;
+		}
+		else{
+			return false;
+		}
+	} 
+	/* if it is not the first time processing the vehicle update, the vehicle cur_lane_id must be the same as entry_lane_id, link_lane_id, or depart_lane_id . */
```
dan-du-car, line 294: Maybe change to "if not the first successful process of the vehicle update"?

scheduling_service/src/scheduling.cpp
```diff
@@ -181,6 +181,8 @@ scheduling::scheduling(unordered_map<string, vehicle> list_veh, set<string> list
 			if (list_veh_confirmation.find(element.first) != list_veh_confirmation.end()){
 				list_veh_confirmation.erase(element.first);
 			}
+			spdlog::info("Vehicle {0} is not added to the schedule as its update is more than {1} seconds old!", element.first, config.get_expDelta());
```
dan-du-car, line 184: Why intially 3 seconds for expiration?

Saeid-Solei: is the expiration delay set to 3 seconds? I agree it should be more than 3 seconds

scheduling_service/src/vehicle.cpp
Saeid-Solei: There is no specific reason. I just wanted to know when the scheduling service forces a short distance.

scheduling_service/src/vehicle.cpp
Saeid-Solei: well, if the vehicle is EV, its lane id shall be entry_lane_id anyway.

scheduling_service/src/vehicle.cpp
Saeid-Solei: It can. If the cur_lane_id, entry_lane_id, and link_lane_id are valid, then this section checks whether cur_lane_id is the same as entry_lane_id or link_lane_id or not.

scheduling_service/src/vehicle.cpp
Saeid-Solei: sure.

scheduling_service/src/vehicle.cpp
dan-du-car: `{"metadata":{"timestamp":1640037608988},"payload":{"v_id":"DOT-45254","v_length":500,"cur_speed":225.0,"cur_accel":87.0,"react_t":5.0,"max_accel":1.5,"max_decel":-0.6,"min_gap":10.0,"depart_pos":1000,"is_allowed":false,"cur_lane_id":5894,"cur_ds":32.71167895363702,"entry_lane_id":5894,"dest_lane_id":0,"link_lane_id":23016,"direction":"straight","est_paths":[{"id":5894,"ds":32.71167895363702,"ts":1640037608958},{"id":5894,"ds":22.585244874053974,"ts":1640037609958},{"id":5894,"ds":13.772385032539258,"ts":1640037610858},{"id":5894,"ds":4.750002729608827,"ts":1640037611758}]}}`

dan-du-car: The above is a vehicle status and intent message we captured during the integration testing. The entry and link lane ids are both correct. The dest_lane_id is invalid here. Is this record not processed as well?

dan-du-car: `{"metadata":{"timestamp":1640037632486},"payload":{"v_id":"DOT-45254","v_length":500,"cur_speed":205.0,"cur_accel":-57.0,"react_t":5.0,"max_accel":1.5,"max_decel":-0.6,"min_gap":10.0,"depart_pos":1000,"is_allowed":true,"cur_lane_id":12459,"cur_ds":138.96261737449385,"entry_lane_id":0,"dest_lane_id":0,"link_lane_id":0,"direction":"straight","est_paths":[{"id":12459,"ds":138.96261737449385,"ts":1640037632468},{"id":12459,"ds":0.0,"ts":1640037633468}]}}`

dan-du-car: This is another vehicle status and intent message. dest, link, and entry lane ids are all invalid. cur_lane_id indicates that the vehicle is currently in departure lane. This record should not be processed as well?

scheduling_service/src/scheduling.cpp
dan-du-car: sorry, 5 mins = 3000 secs.

dan-du-car: We usually update it to 1 sec or maybe 10 during the integration testing.
Thanks!

scheduling_service/src/vehicle.cpp
dan-du-car: If the link and entry lane ids are valid, the process is successful, and the id is no longer empty/"". So the if (id=="") is false?

scheduling_service/src/vehicle.cpp
```diff
@@ -15,11 +15,24 @@ void vehicle::update(const rapidjson::Document& message, intersection_client& lo
...
 		*  the unit of the speed defined in the vehicle class is meter per second. 
 		*/
 		if (message["payload"].HasMember("cur_speed")){
-			speed = message["payload"]["cur_speed"].GetDouble() / 100.0;
+			if (message["payload"]["cur_speed"].GetDouble() * 0.02 > config.get_maxValidSpeed() && id != "" && message["payload"].HasMember("cur_ds") && message["payload"].HasMember("cur_lane_id")){
+				if (to_string(message["payload"]["cur_lane_id"].GetInt()) == lane_id){
+					speed = (distance - message["payload"]["cur_ds"].GetDouble()) / (((double)message["metadata"]["timestamp"].GetInt64() / 1000.0) - timestamp);
+				}
+				else{
+					speed = (distance + (localmap.get_laneLength(to_string(message["payload"]["cur_lane_id"].GetInt())) - message["payload"]["cur_ds"].GetDouble())) / (((double)message["metadata"]["timestamp"].GetInt64() / 1000.0) - timestamp);
```
dan-du-car, line 27: Not sure I understand this equation here. If cur_lane_id is not lane id in the vehicle class, then (vehicle class distance + total length of current lane - current vehicle distance to the end of current lane) / (current vehicle timestamp - vehicle class timestamp)? Could you provide a situation like this?

scheduling_service/src/vehicle.cpp
```diff
@@ -15,11 +15,24 @@ void vehicle::update(const rapidjson::Document& message, intersection_client& lo
 
 	if (message["metadata"].HasMember("timestamp") && (double)message["metadata"]["timestamp"].GetInt64() / 1000.0 >= timestamp){
 		
-		/* the unit of the received speed from the message is centimeter per second
+		/* the unit of the received speed from the message is 0.02 of meter per second
 		*  the unit of the speed defined in the vehicle class is meter per second. 
 		*/
 		if (message["payload"].HasMember("cur_speed")){
-			speed = message["payload"]["cur_speed"].GetDouble() / 100.0;
+			if (message["payload"]["cur_speed"].GetDouble() * 0.02 > config.get_maxValidSpeed() && id != "" && message["payload"].HasMember("cur_ds") && message["payload"].HasMember("cur_lane_id")){
```
dan-du-car, line 22: not sure why we have restriction for valid speed?

scheduling_service/src/vehicle.cpp
Saeid-Solei: It says if cur_lane_id of the new message is not the same as cur_lane_id from the previous message (i.e., stored lane_id), then the vehicle has move from one lane to another since last update. An example could be a vehicle moving from an entry lane to a link. In this case, the total traveled distance between the last and new updates will be the distance from the previous update plus the difference between the length of the cur_lane_id and the distance in the new update.

scheduling_service/src/vehicle.cpp
Saeid-Solei: No, this check decides whether a process is successful or not. If id="" which means that we haven't successfully process the vehicle update yet, three conditions must be met to be able to process an update for the first time. Those conditions are:
1. the entry_lane_id be valid
2. the link_lane_id be valid
3. the cur_lane_id is the same as entry_lane_id or link_lane_id

dan-du-car (approved):

PR #69, opened 2022-01-04
UC1 integration testing issues

dan-du-car:
# PR Details
## Description
UC1 integration testing issues:
Current distance to end of lanelet calculation sometimes generate the incorrect value.
Remove code that is not needed.

## Related Issue
NA

## Motivation and Context
UC#1

## How Has This Been Tested?
Integration testing with one vehicle

## Types of changes

## Checklist:


message_services/manifest.json
```diff
@@ -11,5 +11,5 @@
     "MOBILITY_PATH_TRAJECTORY_OFFSET_DURATION": "9",
     "VSI_TH_SLEEP_MILLI_SEC": "100",
     "BSM_MSG_EXPIRE_IN_SEC": "1",
-    "CLEAN_QUEUE_IN_SECS": "2"
+    "CLEAN_QUEUE_IN_SECS": "3"
```
paulbourelly999, line 14: What is this parameter for?

message_services/test/test_message_lanelet2_translation.cpp
```diff
@@ -92,21 +92,21 @@ TEST(test_message_lanelet2_translation, distance2_cur_lanelet_end_point)
 
     //Position within the link lanelet with turn direction
     lanelet::Point3d point3d = {lanelet::utils::getId(), {-83.21549512455204, 329.05751672287005, 72}};
-    ASSERT_NEAR(15.033, clt.distance2_cur_lanelet_end(point3d, "right", trajectory), 0.01);
-
+    ASSERT_NEAR(15.286, clt.distance2_cur_lanelet_end(point3d, "right", trajectory), 0.01);
```
paulbourelly999, line 95: What has changed in these unit test.

paulbourelly999 (approved):
Left a couple comments but it looks good. What bug is this meant to fix. Why are those classes/methods no longer necessary?

message_services/test/test_message_lanelet2_translation.cpp
dan-du-car: There is an update to this function distance2_cur_lanelet_end. The unit tests  update for this change.

message_services/manifest.json
dan-du-car: Every 3 seconds, the BSM, MobilityPath queues are cleared, so that any staled data is removed, and also prevent the queue from growing too large.

paulbourelly999: So this means that currently we are unable to consume queues faster than they populate (10 hz )

dan-du-car: We are able to consume the queues, but not able to find the match in time before it grows too large and slow down the application.

PR #102, opened 2022-01-26
Hot fix csv logging

paulbourelly999:
# PR Details
## Description

Transitioning from using custom logging code to using spdlog library for csv logging. The spdlog library offers capability to create a asynchronous rotating file logger, where you can configure max file size, maximum number of files, and file location. This improves on the previous implementation since it is more stable ( popular library vs. custom code ), asynchronous ( can even configure thread pool ), likely higher performance, more configurable, and writes files directly to volume upon creation. The current implementation only writes files to volume when they are cut, either by shutdown or by file size limit.

Also corrected toCSV method for multiple vehicles and added configuration parameters for maximum number of files.

Documentation for spdlog implementation can be found here (https://spdlog.docsforge.com/master/getting-started/)
## Motivation and Context

This improves on the previous implementation since it is more stable ( popular library vs. custom code ), asynchronous ( can even configure thread pool ), likely higher performance, more configurable, and writes files directly to volume upon creation. The current implementation only writes files to volume when they are cut, either by shutdown or by file size limit.

Also corrected toCSV method for multiple vehicles and added configuration parameters for maximum number of files.
## How Has This Been Tested?
Tested locally using python scripts to send vehicle status and intent messages. Tested file creation, file rotation, volume deployment, and incorrect file path error handling.

## Types of changes

- [ x] New feature (non-breaking change that adds functionality)

## Checklist:


dan-du-car (commented):
Is this target branch release? It is hard to see what your changes are.

scheduling_service/src/scheduling.cpp
```diff
@@ -218,21 +218,25 @@ scheduling::scheduling(unordered_map<string, vehicle> list_veh, set<string>& lis
...
-		schedule_info += to_string(acceleration[i]) + ", ";
-		schedule_info += lane_id[i] + ", ";
-		schedule_info += state[i] + ", ";
-		schedule_info += to_string(departurePosition_index[i]) + ", ";
-		schedule_info += to_string(distance[i]) + ", ";
-		schedule_info += to_string(clearance_time[i]) + ", ";
-		schedule_info += to_string(st[i]) + ", ";
-		schedule_info += to_string(et[i]) + ", ";
-		schedule_info += to_string(dt[i]) + ", ";
+		schedule_info += veh_id[i] + ",";
```
dan-du-car, line 223: The local testing result below:
image
Is it possible to add header to add titles to the columes

scheduling_service/src/main.cpp
```diff
@@ -497,8 +503,23 @@ void call_scheduling_thread(){
     double last_schedule;
 
     // Create logger
-    auto logger = std::unique_ptr<csv_logger>(new csv_logger( config.get_scheduleLogPath(), config.get_scheduleLogFilename(), config.get_scheduleLogMaxsize() ));
-    
+    if ( config.isScheduleLoggerEnabled() ) {
+        try{
+            auto logger = spdlog::rotating_logger_mt<spdlog::async_factory>(
+                "csv_logger",  // logger name
+                config.get_scheduleLogPath()+config.get_scheduleLogFilename() +".csv",  // log file name and path
```
dan-du-car, line 510: Ran with the docker-compose up in local. It is able to generate the csv file inside container, but not able to see the logs in host machine?

dan-du-car: Could we add timestamp to the csv file name when the .csv file is about to full/rotate? It is helpful to identify when the logs are generated.

scheduling_service/src/main.cpp
```diff
@@ -83,12 +85,16 @@ void consumer_update(const char* payload){
...
 
     scheduling schedule(list_veh, list_veh_confirmation, localmap, config, list_veh_removal);
     if ( config.isScheduleLoggerEnabled() ) {
-        logger->log_line( schedule.toCSV() ); 
-    }
+        auto logger = spdlog::get("csv_logger");
+        if ( logger != nullptr )
+            logger->info( schedule.toCSV());
+        
+    };
```
dan-du-car, line 96: Remove the semicolon after the curly brace

scheduling_service/src/scheduling.cpp
```diff
@@ -218,21 +218,25 @@ scheduling::scheduling(unordered_map<string, vehicle> list_veh, set<string>& lis
...
+		schedule_info += to_string(distance[i]) + ",";
+		schedule_info += to_string(clearance_time[i]) + ",";
+		schedule_info += to_string(st[i]) + ",";
+		schedule_info += to_string(et[i]) + ",";
+		schedule_info += to_string(dt[i]) + ",";
 		schedule_info += to_string(access[i]);
+		// Added comma to separate vehicle entries
+		if ( i != (int)veh_id.size()-1 ) {
+			schedule_info += ",";
+		}
```
dan-du-car, line 239: Are these data sufficient in the csv file?

dan-du-car: Maybe convert all the timestamp (st, et....) inside the csv file to human readable format? If not, maybe the first column timestamp since it is when the schedule is generated?

paulbourelly999: Not sure if this is a good idea. We can convert all these timestamps to human readable format in post processing but keeping them in this format I think makes calculations with them easiest. (ie time difference ). These are the values @Saeid-Solei wanted in the csv. We can add any retro-actively

scheduling_service/src/main.cpp
paulbourelly999: Looked into this and apparently the only way to achieve this is to use a daily file sink instead of a rotating file sink. Here is the github issue for people requesting this feature for the rotating file sink (https://github.com/gabime/spdlog/issues/580).Both rotating and daily file sinks are able to limit the maximum number of log files we have so we would never have to worry about clearing the logs. I think cutting log files at a certain size is cleaner because it gives us more control of the resource that each service would use.. I could change this to a daily file sink but daily log files may grow larger than we would want. My vote is to keep it like this and maybe rename files based on date when we pull them from the carma-streets VM, but I am open to implement either. (https://spdlog.docsforge.com/v1.x/api/spdlog/sinks/daily_file_sink/)

scheduling_service/src/scheduling.cpp
paulbourelly999: On second thought. The schedule can grow and shrink dynamically in size, number of vehicles accounted for. Because of this I don't think we can add header columns since that can change over the duration of the log multiple times

scheduling_service/src/main.cpp
```diff
@@ -501,8 +509,23 @@ void call_scheduling_thread(){
     double last_schedule;
 
     // Create logger
-    auto logger = std::unique_ptr<csv_logger>(new csv_logger( config.get_scheduleLogPath(), config.get_scheduleLogFilename(), config.get_scheduleLogMaxsize() ));
-    
+    if ( config.isScheduleLoggerEnabled() ) {
+        try{
+            auto logger = spdlog::daily_logger_mt<spdlog::async_factory>(
+                "csv_logger",  // logger name
+                config.get_scheduleLogPath()+config.get_scheduleLogFilename() +".csv",  // log file name and path
```
dan-du-car, line 516: Can we use relative path get_scheduleLogPath()?

paulbourelly999: Resolved this.

scheduling_service/src/main.cpp
```diff
@@ -85,14 +87,20 @@ void consumer_update(const char* payload){
...
-rapidjson::Value scheduling_func(unordered_map<string, vehicle> list_veh, Document::AllocatorType& allocator, std::unique_ptr<csv_logger> &logger, double &last_schedule){
+
+rapidjson::Value scheduling_func(unordered_map<string, vehicle> list_veh, Document::AllocatorType& allocator, double &last_schedule){
+
     std::unique_lock<std::mutex> lck(worker_mtx);
     scheduling schedule(list_veh, list_veh_confirmation, localmap, config, list_veh_removal);
     lck.unlock();
     if ( config.isScheduleLoggerEnabled() ) {
-        logger->log_line( schedule.toCSV() ); 
+        auto logger = spdlog::get("csv_logger");
```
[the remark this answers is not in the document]
paulbourelly999: I think this code smell is in reference to list_veh not the logger variable. I can updated it anyways for clarity.

dan-du-car (approved):

PR #104, opened 2022-01-31
103 adding configurable daily logging

paulbourelly999:
# PR Details
## Description

This PR adds daily rotating file logs to both the scheduling and message services.

## Types of changes

## Checklist:

- [ x ] I have updated the documentation accordingly.
- [ x ] I have read the **CONTRIBUTING** document.

### DIFF
--- a/message_services/src/services/vehicle_status_intent_service.cpp
+++ b/message_services/src/services/vehicle_status_intent_service.cpp
@@ -3,6 +3,8 @@
 
 #include "vehicle_status_intent_service.h"
 
+
+
 namespace message_services
 {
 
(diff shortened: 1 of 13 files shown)


message_services/src/services/vehicle_status_intent_service.cpp
```diff
@@ -35,6 +37,26 @@ namespace message_services
                 // producer topics
                 this->vsi_topic_name = client->get_value_by_doc(doc, "VSI_PRODUCER_TOPIC");
 
+                // Create default multisink daily file logger
+                try {
+                    std::string loglevel = client->get_value_by_doc(doc, "LOG_LEVEL");
+
+                    spdlog::init_thread_pool(8192, 1);
```
dan-du-car, line 44: Is it possible to consolidate this logger code into a library, so that we do not need to repetitively adding the similar code for every component/service?

paulbourelly999: I think we should consolidate reading configurations from a manifest.json file into a library. Maybe as part of that we could add some general configuration implementation like log level but I think the issue here is that for each service we read the same configuration json differently. We need to develop some sort of utility library for reading the manifest json and start coming up with standards for how to name and include new configuration parameters.  I think that is a good idea but maybe not in the scope of this issue.

scheduling_service/src/main.cpp
```diff
@@ -81,6 +85,31 @@ void consumer_update(const char* payload){
...
+/**
+ * Method to set spdlog default logger to a multisink( daily rotating file logger and stdout logger ) asynchronous logger.
+ */ 
+void configure_logger() {
+  // Create default multisink daily file logger
+    std::string loglevel = config.get_loglevel();
+    spdlog::init_thread_pool(8192, 1);
+    try {
+        
+        auto file_sink = std::make_shared<spdlog::sinks::daily_file_sink_mt>("/home/carma-streets/scheduling_service/logs/scheduling_service.log", 23, 3);
```
dan-du-car, line 98: Can we use relative folders and file names, so that we can run this both local and through containers. As the local path to this logs might be different from the container one.

message_services/src/services/vehicle_status_intent_service.cpp
```diff
@@ -35,6 +37,26 @@ namespace message_services
...
+                    auto file_sink = std::make_shared<spdlog::sinks::daily_file_sink_mt>("/home/carma-streets/message_services/logs/message_service.log", 23, 3);
```
dan-du-car, line 45: Can we use relative folders and file names, so that we can run this both local and through containers. As the local path to this logs might be different from the container one.

scheduling_service/src/main.cpp
paulbourelly999: Making this fix now. Great idea

message_services/src/services/vehicle_status_intent_service.cpp
paulbourelly999: Making this fix now. Great idea

message_services/src/services/vehicle_status_intent_service.cpp
```diff
@@ -35,6 +37,26 @@ namespace message_services
                 // producer topics
                 this->vsi_topic_name = client->get_value_by_doc(doc, "VSI_PRODUCER_TOPIC");
 
+                // Create default multisink daily file logger
+                try {
```
dan-du-car, line 41: https://sonarcloud.io/project/issues?id=usdot-fhwa-stol_carma-streets&open=AX7-ZsZtiNHjES7F7cRM&pullRequest=104&resolved=false&types=CODE_SMELL
Could you please resolve this as well?

dan-du-car: @paulbourelly999 Please let me know if we can move this logic to a separate function

paulbourelly999: Great catch, updating this now.

Issue #115, opened 2022-02-17
Create common utility library to parse manifest.json configurations


paulbourelly999:
### Types of Issue

- [ ] Anomaly report (something appears to not work correctly)
- [ x ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

Each service in carma-streets currently parses its manifest json with its own implementation. This are sometimes substantially different. It may be worth will creating a utility library to establish a common way to stand up a carma-streets service and also parse these configuration files. This may help make each carma streets service implementation be more consistent with each other and make the services easier to read.
### Expected behavior
NA
### Actual behavior

NA
### Steps to reproduce the actual behavior

NA

### Related work

#119 #118

PR #122, opened 2022-03-21
115 streets service base lib

paulbourelly999:
# PR Details
## Description

This PR encapsulates the addition of a new CARMA-Streets library which can act as a library that all CARMA-Streets service can depend on. This library has two main pieces of functionality: **Extensible Singleton Implementation**, **Streets Configuration**.

**Extensible Singleton Implementation** : Library offers access to the `streets_singleton`  class. This is a templated, extensible class that stores and offers static retrieval of a single instance of itself, which is lazily initialized (not initialized until retrieved for the first time). To ensure that new instances of this class can not be created the constructors are deleted or hidden using private or protected access.
 
**Streets Configuration** : Library creates `streets_configuration` singleton which standardizes configuration file parsing and configuration parameter retrieval. Extending `streets_singleton` allows `streets_configuration` to be limited to singleton scope ( single instance ) and offer static methods for configuration parameter retrieval. `streets_configuration` also parses some service required configurations like **loglevel** and **service_name** to create and configure a `spdlog::async_loggerr` with a  `spdlog::sinks::daily_file_sink_mt` and a `spdlog::sinks::stdout_color_sink_mt`.

The expected format for `manifest.json` configuration files is :
```
{
    "service_name": "test_service",
    "loglevel": "info",
    "configurations": [
        {
            "name": "param_1",
            "value": 1,
            "description": "Test Parameter 1",
            "type": "INTEGER"
        },
        {
            "name": "param_2",
            "value": "Second Parameter",
            "description": "Test Parameter 2",
            "type": "STRING"
        },
        {
            "name": "param_3",
            "value": 123.2,
            "description": "Test Parameter 3",
            "type": "DOUBLE"
        },
        {
            "name": "param_4",
            "value": true,
            "description": "Test Parameter 3",
            "type": "BOOL"
        }
    ]
}
```

## Related Issue
#115 

## Motivation and Context

Before this each **CARMA-Streets** service parsed configuration parameters from manifest.json files with separate custom methods. This was also the case for creating and configuring spdlog default logger. This library implements both of these things, avoiding code duplication and standardizing the practice among all **CARMA-Streets** services. It also offers static retrieval of configuration parameters which alleviates the need to pass configuration data objects between methods to where the parameters are used. Finally it also provides a extensible singleton class which can be used to implement singleton scope objects without the use of any global variables like in the `scheduling_service`.
## How Has This Been Tested?
Unit Testing

## Types of changes

## Checklist:

- [ x] I have added any new packages to the sonar-scanner.properties file


paulbourelly999:
@msmcconnell still working on CI/CD stuff and related refactoring for all existing services to use this new library but I just wanted to add you to the review now so you could look at my singleton implementation and give me any initial feed back. That's probably the only part of the code a need you to look at however would very much appreciate any additional portions you can review and any feedback you can provide. Still getting acclimated to C++ standards and style. If there is something I am doing weird or incorrect any suggestions would be greatly appreciated. Tried/still trying to document stuff as much as possible to make things clear. Let me know if anything is confusing.

streets_utils/streets_service_base/include/streets_singleton.h
```diff
@@ -0,0 +1,57 @@
...
+             * Protected destructor
+             */        
+            ~streets_singleton();        
+    };
+
+    // Implementation
+
+    template <typename T>   
+    T& streets_singleton<T>::get_singleton() {
+        static T instance;  // Guaranteed to be destroyed.
```
msmcconnell: Per our discussion in the lab we need to confirm this is the same between compilation units in the same executable.

streets_utils/streets_service_base/include/streets_configuration.h
```diff
@@ -0,0 +1,119 @@
...
+
+
+namespace streets_service {
+    /**
+     * @brief Streets Configuration singleton scoped object that parses
+     * manifest.json configuration file and offers static access to the
+     * configuration parameter values. Singleton also configures default
+     * multisink logger file and terminal logger.
+     */ 
+    class streets_configuration : public streets_singleton<streets_configuration> {
```
msmcconnell, line 32: As discussed can you share where you found this approach. I want to understand why its not a circular dependency.

streets_utils/streets_service_base/src/streets_configuration.cpp
```diff
@@ -0,0 +1,193 @@
...
+        }   
+        catch (const spdlog::spdlog_ex& ex)
+        {
+            spdlog::error( "Log initialization failed: {0}!",ex.what());
+        }
+    };
+
+    void streets_configuration::initialize(){
+        streets_configuration::get_singleton();
+    };
```
msmcconnell, line 185: Seems like this function isn't overly valuable since the first call will initialize the variable anyway.

msmcconnell (commented):
Here are my initial comment. I haven't really dug into it, but these and our discussion should cover the singleton approach.

streets_utils/streets_service_base/include/streets_configuration.h
```diff
@@ -0,0 +1,119 @@
...
+        friend class streets_singleton<streets_configuration>;
+
+        private:
+            /* String filepath to manifest.json configuration file */
+            std::string filepath;
+            /* mutex lock for streets_configuration*/
+            std::mutex config_lock;
+            /* Map of configuration names and values*/
+            std::map< std::string, configuration > configuration_map;
```
paulbourelly999, line 41: The following code smell exists on this line (https://sonarcloud.io/project/issues?resolved=false&sinceLeakPeriod=true&types=CODE_SMELL&pullRequest=122&id=usdot-fhwa-stol_carma-streets&open=AX-3TJ7M3wW2MlKvf3di). Not sure I understand it though. Any ideas of what changes are actually necessary for this

streets_utils/streets_service_base/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
+cmake_minimum_required(VERSION 3.10.2)
+project(streets_service_base)
+                    
+link_directories("/usr/lib"  "/usr/local/lib"  )
```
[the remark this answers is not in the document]
paulbourelly999: Probably not necessary, will remove

streets_utils/streets_service_base/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
...
+
+find_package(Boost COMPONENTS system thread REQUIRED)
+find_package(spdlog REQUIRED)
+
+include_directories("/usr/local/include")
```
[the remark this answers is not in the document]
paulbourelly999: Probably not necessary. Will attempt to remove.

streets_utils/streets_service_base/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
...
+
+add_library(${PROJECT_NAME}_lib 
+                src/streets_configuration_exception.cpp
+                src/configuration.cpp
+                src/streets_configuration.cpp
+                )
+target_include_directories( ${PROJECT_NAME}_lib PUBLIC ${PROJECT_SOURCE_DIR}/include
```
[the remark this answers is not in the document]
paulbourelly999: Will attempt to remove. Added a lot of these to attempt to get things to work. May be a remnant from that.

streets_utils/streets_service_base/include/streets_configuration.h
paulbourelly999: @msmcconnell The following code smell exists on this line (https://sonarcloud.io/project/issues?resolved=false&sinceLeakPeriod=true&types=CODE_SMELL&pullRequest=122&id=usdot-fhwa-stol_carma-streets&open=AX-3TJ7M3wW2MlKvf3di). Not sure I understand it though. Any ideas of what changes are actually necessary for this.

streets_utils/streets_service_base/include/streets_singleton.h
```diff
@@ -0,0 +1,57 @@
...
+        public:
+            /**
+             * Method for static singleton instance retrieval
+             * @returns T singleton instance
+             */ 
+            static T& get_singleton() ;
+            // Remove copy constructor 
+            streets_singleton(streets_singleton const &) = delete;
+            // Remove move constructor
+            streets_singleton& operator=(streets_singleton const &) = delete;
```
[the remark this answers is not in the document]
paulbourelly999: Will remove that as well.

streets_utils/streets_service_base/src/streets_configuration.cpp
paulbourelly999: So the only reason I added this is to configuration the spdlogger.

streets_utils/streets_service_base/src/streets_configuration.cpp
```diff
@@ -0,0 +1,193 @@
...
+        if ( doc.FindMember("configurations")->value.IsArray() ) {
+            parse_configurations_array(doc.FindMember("configurations")->value.GetArray());
+        }
+        else {
+            spdlog::warn("No configurations found in manifest.json!");
+        }   
+    };
+
+    void streets_configuration::parse_configurations_array( const rapidjson::GenericArray<true, rapidjson::Value> &arr) {
+        std::unique_lock<std::mutex> lck(config_lock);
```
paulbourelly999, line 64: @msmcconnell @dan-du-car and I were having some discussion about whether using mutex locks inside methods like this that make edits to the underlying data to make it thread safe is the correct/ C++ standard way of making objects thread safe. I guess the alternative would be to rely on users of the library to implement locking mechanisms when calling this methods but having this internal seemed like the safer implementation. What do you think?

streets_utils/streets_service_base/src/streets_configuration.cpp
```diff
@@ -0,0 +1,193 @@
...
+            }
+            else {
+                throw streets_service::streets_configuration_exception("Configuration parameter " + property_name + " not properly formatted!");
+            }  
+        }
+    };
+
+    std::string streets_configuration::get_string_config(const std::string &config_param_name) {
+        // Get Singleton
+        auto &instance = get_singleton();
```
paulbourelly999, line 89: @msmcconnell Another discussion @dan-du-car and I had was whether updating the singleton on get parameters is a valid/good design paradigm for the configuration singleton. I have added a check to see if the configuration file has been modified before rereading it. The alternative would be to provide external methods for services that implement this library to write logic to chose when to update configuration values from the manifest.json. I thought this was is appropriate because then each service does not need to manage this itself and it allows for "live" reading of configuration values so that we don't need to restart a service for configuration changes to hold

streets_utils/streets_service_base/include/streets_configuration.h
paulbourelly999: @msmcconnell Here is a link to the C++ singleton implementation I worked off (https://stackoverflow.com/questions/41328038/singleton-template-as-base-class-in-c).

streets_utils/streets_service_base/include/streets_singleton.h
paulbourelly999: With the log statement of the singleton memory address I integrated this into the scheduling_service . In the scheduling_service multiple different .cpp files ( .o files therefore translational units) call get_instance on the configuration singleton. There was still only a single memory address every printed. Will continue to test further to make sure this is the case.

streets_utils/streets_service_base/test/test_streets_configuration.cpp
```diff
@@ -0,0 +1,136 @@
+#define RAPIDJSON_HAS_STDSTRING 1
```
paulbourelly999, line 1: Enables SetString(std::string) method for updating configurations.

dan-du-car (approved):

streets_utils/streets_service_base/src/streets_configuration.cpp
```diff
@@ -0,0 +1,193 @@
...
+        update_configuration( doc );
+        spdlog::info("Printing Configuration Parameters ---------------------------------------");
+        for(const auto& conf : configuration_map)
+        {
+            spdlog::info("{0} : {1} ", conf.first.c_str(), conf.second.value.c_str());
+        }
+    };
+
+    rapidjson::Document streets_configuration::parse_configuration_file() {
+        std::unique_lock<std::mutex> lck(config_lock);
```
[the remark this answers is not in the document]
paulbourelly999: We need a lock here because we open and read from the json file. I thought opening the same file from two different threads may cause issues so I put the lock here. Will look online to see if this is actually an issue

streets_utils/streets_service_base/src/streets_configuration.cpp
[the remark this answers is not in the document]
paulbourelly999: We definitely need this lock since this method edits the map of configurations stored in the singleton. We do not want two threads to ever execute this method simultaneously since it would cause concurrency issues

PR #130, opened 2022-04-05
129 refactor intersection model

paulbourelly999:
# PR Details
## Description

PR to refactor intersection_model code to use streets_service_base_lib for configuration parsing, storing and retrieval. Also updated intersection_model to use SPDLOG macros to include filename and line number.
## Motivation and Context
All CARMA-Streets should use a common library to parse and store configuration parameters to help standardize CARMA-Streets service, make services more readable and reduce code duplication.

## How Has This Been Tested?
Unit tested and Docker image deployed.

## Types of changes

## Checklist:


intersection_model/test/test_intersection_model.cpp
```diff
@@ -3,46 +3,28 @@
```
paulbourelly999, line 3: @dan-du-car  tried to make correct edits to test cases to get them to pass. Please let me know if I made some errors here. I think two of the test cases still fail

intersection_model/test/test_intersection_model.cpp
```diff
@@ -3,46 +3,28 @@
...
-    ASSERT_FALSE(model.get_vehicleGraph_ptr());
-    model.update_vehicle_routing_graph();
+    intersection_model::intersection_model model("West Intersection", 9001,"../../MAP/town01_vector_map_test.osm");
     ASSERT_TRUE(model.get_vehicleGraph_ptr());
 }
 
 TEST(intersection_model_test, update_intersection_info_by_all_way_stop_reg)
 {
-    intersection_model::intersection_model model("../vector_map.osm");
+    intersection_model::intersection_model model("West Intersection", 9001,"../../MAP/town01_vector_map_test.osm");
```
paulbourelly999, line 27: I am not sure I understand this test case. I assumed that this update_intersection_info_by_all_way_stop_reg() method gets called by default by the initialization method. Also for some reason link lanelets returns 12 instead of 6. The all-stop intersection I added only has 3 entry lanes and 3 departure lane. Please check out the osm file I committed.

intersection_model/test/test_intersection_model.cpp
paulbourelly999: All test cases base now

intersection_model/test/test_intersection_model.cpp
paulbourelly999: All test cases pass now

intersection_model/src/server/src/main.cpp
```diff
@@ -55,43 +57,33 @@ int main(int argc, char * argv[])
...
+    quint16 port = static_cast<quint16>(
+        streets_service::streets_configuration::get_int_config("host_port")
+    );
+
 
     QSharedPointer<OpenAPI::OAIApiRequestHandler> handler(new OpenAPI::OAIApiRequestHandler());
-    auto router = QSharedPointer<OpenAPI::OAIApiRouter>::create();
+    auto router = QSharedPointer<OpenAPI::OAIApiRouter>::create(
+        streets_service::streets_configuration::get_string_config("intersection_name"),
+        streets_service::streets_configuration::get_int_config("intersection_id"),
```
dan-du-car, line 74: It is not necessary to pass in the intersection name, id and osm file path here. Should we pass in where we call the intersection model lib so that it gives a bit context of these three config variables?

intersection_model/src/intersection_model.cpp
```diff
@@ -3,96 +3,37 @@
...
-            read_lanelet2_map(osm_file_path);
-        }
         
+        int_info.intersection_name = intersection_name;
+        int_info.intersection_id = intersection_id; 
+        read_lanelet2_map(osm_file_path);
         if (map && !map->empty())
         {
-            spdlog::info("{0} Map is loaded ...", __FILE__);
+            SPDLOG_INFO("Map is loaded ...");
```
dan-du-car, line 18: Where did you define this macros?

intersection_model/src/intersection_model.cpp
```diff
@@ -107,9 +48,9 @@ namespace intersection_model
             projector = new lanelet::projection::LocalFrameProjector(target_frame.c_str());
             map = lanelet::load(filename, *projector, &errors);
         }
-        catch(std::exception ex)
+        catch(const lanelet::LaneletError &ex)
         {         
-            spdlog::error("{0}: Cannot read osm file {1}. Error message: {2} " , __FILE__, filename,ex.what());
+            SPDLOG_ERROR("Cannot read osm file {0}. Error message: {1} ", filename, ex.what());
```
dan-du-car, line 53: Where did you define this macros?

intersection_model/src/intersection_model.cpp
```diff
@@ -171,14 +112,14 @@ namespace intersection_model
                 {
                     if (!update_intersection_info_by_all_way_stop_reg(reg_element_ptr))
                     {
-                         spdlog::critical("{0}: Failed to load intersection information based on all_way_stop regultory element. ",__FILE__);
+                         SPDLOG_CRITICAL("Failed to load intersection information based on all_way_stop regulatory element.");
```
dan-du-car, line 115: Where did you define this macros?

intersection_model/src/intersection_model.cpp
paulbourelly999: See https://spdlog.docsforge.com/v1.x/3.custom-formatting/#pattern-flags documentation. It is a spdlog define macros but their filename and file line number logging only works with this macros and not the regular method.

paulbourelly999: See the pattern flag section.

intersection_model/src/intersection_model.cpp
paulbourelly999: See comment above for spdlog documentation

intersection_model/src/intersection_model.cpp
paulbourelly999: See comment above for spdlog documentation.

intersection_model/src/server/src/main.cpp
paulbourelly999: So the intersection_model class depends on these values and the OAIDefaultApitHandler depends on the intersection_model class. 
I assume you are suggesting passing these configuration parameters in inside the OAIDefaultApiHandler constructor. I am not a huge fan of this since I do not see the purpose and it seems to just work to obscure where we get our configuration parameters from and couple our configuration logic with our REST server implementation.  Including them as parameters here means that the REST server implementation does not depend on how we get configurations but rather just that we can provide configurations.  Please let me know what you think.

paulbourelly999: Replaces intersection_model parameter with intersection model in the OAIApiRouter constructor

dan-du-car (approved):

Issue #133, opened 2022-04-13
street_configuration singleton does not update loglevel


paulbourelly999:
### Types of Issue

- [ x ] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

Streets Configuration singleton update does not check or update loglevel parameter if changed while docker container is up.
### Expected behavior
Any changes to the manifest.json configuration file will cause updates to the street_configuration singleton.
### Actual behavior

Changes to loglevel parameter will not update the loglevel for the streets_configuration singleton.
### Steps to reproduce the actual behavior

1. Deploy any carma-street container.
2. Edit loglevel parameter
3. Does loglevel update?

PR #134, opened 2022-04-14
Issue-133:Loglevel fix

paulbourelly999:
# PR Details
## Description
This fix is a change to the streets_service_base library. It is an update that allows the streets_configuration singleton to update `loglevel` configuration parameter when changes are made to the manifest.json file while the carma-streets service is still running.

## Related Issue
#133 

## Motivation and Context
Allows for real-time loglevel adjustment with service restarts

## Types of changes

- [ x ] Defect fix (non-breaking change that fixes an issue)

## Checklist:

- [ x] All new and existing tests passed.

### DIFF
--- a/streets_utils/streets_service_base/src/streets_configuration.cpp
+++ b/streets_utils/streets_service_base/src/streets_configuration.cpp
@@ -5,8 +5,13 @@ namespace streets_service {
     // Constructor
     streets_configuration::streets_configuration( const std::string &filepath ): filepath(filepath){
         SPDLOG_INFO("Printing Configuration Parameters ---------------------------------------");
+        // Parse manifest.json configuration file into rapidjson::Document
         rapidjson::Document doc = parse_configuration_file();
+        // Use service level configuration parameters from Document (i.e. loglevel and service_name)
+        // to configure spdlog default logger (terminal and file sinks)
         configure_logger(doc);
+        // Use configurations array to populate configuration map with service specific configuration
+        // parameters from Document
         update_configuration(doc);
         for(const auto& conf : configuration_map)
         {
@@ -199,12 +204,16 @@ namespace streets_service {
 
     void streets_configuration::check_update() {
         try {
+            // Check last write time of manifest.json configuration file to see if updates have been made
             std::time_t time = boost::filesystem::last_write_time(filepath);
             SPDLOG_DEBUG("Last Modified Time {0} vs Stored Last Modified Time {1}", time, last_modified);
             if ( time > last_modified) {
+                // If updates have been made parse manifest.json into Document and update loglevel and 
+                // any changed configuration parameters.
                 rapidjson::Document doc = parse_configuration_file();
                 update_log_level(doc);
                 update_configuration(doc);
+                // Set new update time to the last update read
                 last_modified = time;
             }
         }
--- a/streets_utils/streets_service_base/src/streets_configuration.cpp
+++ b/streets_utils/streets_service_base/src/streets_configuration.cpp
@@ -5,7 +5,9 @@ namespace streets_service {
     // Constructor
     streets_configuration::streets_configuration( const std::string &filepath ): filepath(filepath){
         SPDLOG_INFO("Printing Configuration Parameters ---------------------------------------");
-        check_update();
+        rapidjson::Document doc = parse_configuration_file();
+        configure_logger(doc);
+        update_configuration(doc);
         for(const auto& conf : configuration_map)
         {
             SPDLOG_INFO("{0} : {1} ", conf.first.c_str(), conf.second.value.c_str());
@@ -55,6 +57,16 @@ namespace streets_service {
         else {
             SPDLOG_WARN("No configurations found in manifest.json!");
         }   
+    };
+
+    void streets_configuration::update_log_level( const rapidjson::Document &doc ){
+        if ( doc.FindMember("loglevel")->value.IsString() ) {
+            set_loglevel( doc.FindMember("loglevel")->value.GetString() );
+        }
+        else {
+            SPDLOG_WARN("Parameter \"loglevel\" missing/incorrectly formatted in manifest.json! Setting \"loglevel\" to INFO!");
+            set_loglevel("info");
+        }
     };
 
     void streets_configuration::parse_configurations_array( const rapidjson::GenericArray<true, rapidjson::Value> &arr) {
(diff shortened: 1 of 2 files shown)


streets_utils/streets_service_base/src/streets_configuration.cpp
```diff
@@ -57,6 +59,16 @@ namespace streets_service {
         }   
     };
 
+    void streets_configuration::update_log_level( const rapidjson::Document &doc ){
+        if ( doc.FindMember("loglevel")->value.IsString() ) {
```
dan-du-car, line 63: Checking the member exist before access the value.

streets_utils/streets_service_base/src/streets_configuration.cpp
```diff
@@ -5,7 +5,9 @@ namespace streets_service {
     // Constructor
     streets_configuration::streets_configuration( const std::string &filepath ): filepath(filepath){
         SPDLOG_INFO("Printing Configuration Parameters ---------------------------------------");
-        check_update();
+        rapidjson::Document doc = parse_configuration_file();
+        configure_logger(doc);
```
dan-du-car, line 9: Add line comments about the functions here.

streets_utils/streets_service_base/src/streets_configuration.cpp
dan-du-car: Should we make this loglevel case insensitive?

paulbourelly999: This is a check to see if it exists. It also checks its type. If the member does not exist this method returns false. 

Also I don't think we should make this case insensitive. This way we force all carma-streets services to abide by common formating (parameter names and cases).

streets_utils/streets_service_base/src/streets_configuration.cpp
paulbourelly999: I can add comments about the functions if you think that is necessary. I thought the java doc comments on the methods and the method names were enough but I am always ok with adding more documentation to make things clearer.

streets_utils/streets_service_base/src/streets_configuration.cpp
```diff
@@ -5,7 +5,14 @@ namespace streets_service {
...
         SPDLOG_INFO("Printing Configuration Parameters ---------------------------------------");
-        check_update();
+        // Parse manifest.json configuration file into rapidjson::Document
+        rapidjson::Document doc = parse_configuration_file();
+        // Use service level configuration parameters from Document (i.e. loglevel and service_name)
+        // to configure spdlog default logger (terminal and file sinks)
+        configure_logger(doc);
+        // Use configurations array to populate configuration map with service specific configuration
+        // parameters from Document
+        update_configuration(doc);
```
dan-du-car, line 15: Is there any sequence calling the configure_logger and update_configuration ?

streets_utils/streets_service_base/src/streets_configuration.cpp
dan-du-car: doc.FindMember("loglevel") != EndMember()

streets_utils/streets_service_base/src/streets_configuration.cpp
```diff
@@ -5,7 +5,14 @@ namespace streets_service {
     // Constructor
     streets_configuration::streets_configuration( const std::string &filepath ): filepath(filepath){
         SPDLOG_INFO("Printing Configuration Parameters ---------------------------------------");
-        check_update();
+        // Parse manifest.json configuration file into rapidjson::Document
+        rapidjson::Document doc = parse_configuration_file();
+        // Use service level configuration parameters from Document (i.e. loglevel and service_name)
+        // to configure spdlog default logger (terminal and file sinks)
+        configure_logger(doc);
```
dan-du-car, line 12: Is this to initialize logger?

streets_utils/streets_service_base/src/streets_configuration.cpp
```diff
@@ -187,10 +204,16 @@ namespace streets_service {
...
         try {
+            // Check last write time of manifest.json configuration file to see if updates have been made
             std::time_t time = boost::filesystem::last_write_time(filepath);
             SPDLOG_DEBUG("Last Modified Time {0} vs Stored Last Modified Time {1}", time, last_modified);
             if ( time > last_modified) {
-                update_configuration(parse_configuration_file());
+                // If updates have been made parse manifest.json into Document and update loglevel and 
+                // any changed configuration parameters.
+                rapidjson::Document doc = parse_configuration_file();
+                update_log_level(doc);
```
dan-du-car, line 214: Why do we need to update_log_level every time check_update? What is the frequency of checking_update called? Or is it manually called from the client code?

streets_utils/streets_service_base/src/streets_configuration.cpp
paulbourelly999: So just for clarity when you call IsString() on EndMember() it returns false. This is why this is valid. So takes care of checking for existence and for type. I am pretty sure this is the case but I can test again to make sure.

paulbourelly999: Confirmed that this is correct by running unit tests with missing loglevel parameter. No exception thrown instead we just get the warning log message and default to info.

streets_utils/streets_service_base/src/streets_configuration.cpp
paulbourelly999: This is called only when the manifest.json file is modified since the last time we read it. It checks for updates every time we retrieve a configuration (get_<type>_configuration(std::string)). This way we could possibly update loglevel for running containers in realtime without having to restart.

streets_utils/streets_service_base/src/streets_configuration.cpp
paulbourelly999: YEs

streets_utils/streets_service_base/src/streets_configuration.cpp
paulbourelly999: Only the constructor should call configure_logger, since it is only required at initialization. Update will call update_loglevel which will just update the loglevel of the constructor. update_configuration is called only when we have modified the manifest.json file since the last time update was called. When this is the case we read the new changes from the manifest.json

dan-du-car (approved):

Issue #140, opened 2022-05-24
Move scheduling logic to library


paulbourelly999:
### Types of Issue

- [ ] Anomaly report (something appears to not work correctly)
- [ x ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

Move all scheduling algorithm logic inside the scheduling service into a library. This will allow multiple services to use this logic. Currently the logic will be required for the planned `signal_optimization_service`  and the `scheduling_service`.

### Carma streets version where this issue was discovered

3.11
### Expected behavior
NA
### Actual behavior

NA

PR #141, opened 2022-05-27
140 vehicle scheduler lib

paulbourelly999:
# PR Details
## Description

Porting UC1 scheduling logic to a library. Implement unit testing 

## Related Issue
#140 

## Motivation and Context
Allow multiple services to use scheduling logic.

## Types of changes

## Checklist:


streets_utils/streets_vehicle_scheduler/include/all_stop_vehicle_scheduler.h
```diff
@@ -0,0 +1,120 @@
...
+             * @brief Estimate clearance time any vehicle given it's link lanelet information.  
+             * 
+             * @param veh vehicle to estimate clearance time for. 
+             * @param link_lane_info link lanelet vehicle is attempting to clear. 
+             * @return uint64_t clearance time in milliseconds. 
+             */
+            uint64_t estimate_clearance_time(const streets_vehicles::vehicle &veh, const OpenAPI::OAILanelet_info &link_lane_info) const;
+
+            uint64_t estimate_earliest_time_to_stop_bar(const streets_vehicles::vehicle &veh) const;
+            double estimate_delta_x_prime( const streets_vehicles::vehicle &veh, const OpenAPI::OAILanelet_info &entry_lane ) const;
```
dan-du-car, line 94: add comments?

streets_utils/streets_vehicle_scheduler/include/all_stop_vehicle_scheduler.h
```diff
@@ -0,0 +1,120 @@
...
+
+            double calculate_v_hat(const streets_vehicles::vehicle &veh) const;
```
dan-du-car, line 96: add comments?

streets_utils/streets_vehicle_scheduler/include/all_stop_vehicle_scheduler.h
```diff
@@ -0,0 +1,120 @@
...
+
+            double calculate_acceleration_time( const streets_vehicles::vehicle &veh, const double v_hat) const;
+
+            double calculate_deceleration_time( const streets_vehicles::vehicle &veh, const double v_hat) const;
+
+            double calculate_cruising_time( const streets_vehicles::vehicle &veh, const double v_hat, const double delta_x_prime) const;
```
dan-du-car, line 102: add comments?

streets_utils/streets_vehicle_scheduler/include/all_stop_vehicle_scheduler.h
```diff
@@ -0,0 +1,120 @@
...
+            uint64_t estimate_clearance_time(const streets_vehicles::vehicle &veh, const OpenAPI::OAILanelet_info &link_lane_info) const;
+
+            uint64_t estimate_earliest_time_to_stop_bar(const streets_vehicles::vehicle &veh) const;
+            double estimate_delta_x_prime( const streets_vehicles::vehicle &veh, const OpenAPI::OAILanelet_info &entry_lane ) const;
+
+            double calculate_v_hat(const streets_vehicles::vehicle &veh) const;
+
+            double calculate_acceleration_time( const streets_vehicles::vehicle &veh, const double v_hat) const;
+
+            double calculate_deceleration_time( const streets_vehicles::vehicle &veh, const double v_hat) const;
```
dan-du-car, line 100: add comments?

streets_utils/streets_vehicle_scheduler/include/all_stop_vehicle_scheduler.h
```diff
@@ -0,0 +1,120 @@
...
+        public:
+
+            all_stop_vehicle_scheduler() = default;
+             
+            ~all_stop_vehicle_scheduler() = default;
+
+            void schedule_vehicles( std::unordered_map<std::string,streets_vehicles::vehicle> &vehicles, intersection_schedule &schedule) override;
+
+            void set_entering_time_buffer( const uint64_t buffer);
+
```
dan-du-car, line 115: add comments?

streets_utils/streets_vehicle_scheduler/include/all_stop_vehicle_scheduler.h
```diff
@@ -0,0 +1,120 @@
...
+
+
+        public:
+
+            all_stop_vehicle_scheduler() = default;
+             
+            ~all_stop_vehicle_scheduler() = default;
+
+            void schedule_vehicles( std::unordered_map<std::string,streets_vehicles::vehicle> &vehicles, intersection_schedule &schedule) override;
```
dan-du-car, line 112: add comments?

streets_utils/streets_vehicle_scheduler/include/all_stop_vehicle_scheduler.h
```diff
@@ -0,0 +1,120 @@
...
+
+            void set_entering_time_buffer( const uint64_t buffer);
```
dan-du-car, line 114: add comments?

streets_utils/streets_vehicle_scheduler/include/vehicle_scheduler.h
```diff
@@ -0,0 +1,52 @@
...
+
+
+namespace streets_vehicle_scheduler {
+    /**
+     * @brief Abstract Class for extend to implement scheduling business logic.
+     * 
+     */
+    class vehicle_scheduler {
+        protected:
+            std::shared_ptr<OpenAPI::OAIIntersection_info> intersection_info;
```
dan-du-car, line 22: add comments?

streets_utils/streets_vehicle_scheduler/include/vehicle_scheduler.h
```diff
@@ -0,0 +1,52 @@
...
+
+            /**
+             * @brief Limits how much departure position for a given vehicle can change from current reported departure position.
+             */
+            int flexibility_limit = 5;
+
+            void estimate_vehicles_at_common_time( std::unordered_map<std::string,streets_vehicles::vehicle> &vehicles,
```
dan-du-car, line 29: add comments?

streets_utils/streets_vehicle_scheduler/include/vehicle_scheduler.h
```diff
@@ -0,0 +1,52 @@
...
+
+            /**
+             * @brief Limits how much departure position for a given vehicle can change from current reported departure position.
+             */
+            int flexibility_limit = 5;
+
+            void estimate_vehicles_at_common_time( std::unordered_map<std::string,streets_vehicles::vehicle> &vehicles, 
+                                                    const u_int64_t timestamp);
+
+            OpenAPI::OAILanelet_info get_entry_lanelet_info(const streets_vehicles::vehicle &veh) const;
```
dan-du-car, line 32: add comments?

streets_utils/streets_vehicle_scheduler/include/vehicle_scheduler.h
```diff
@@ -0,0 +1,52 @@
...
+
+            OpenAPI::OAILanelet_info get_link_lanelet_info(const streets_vehicles::vehicle &veh) const;
```
dan-du-car, line 34: add comments?

streets_utils/streets_vehicle_scheduler/include/vehicle_scheduler.h
```diff
@@ -0,0 +1,52 @@
...
+
+            vehicle_scheduler() = default;
+
+            ~vehicle_scheduler() = default;
+
+            virtual void schedule_vehicles( std::unordered_map<std::string,streets_vehicles::vehicle> &vehicles, intersection_schedule &schedule) = 0;
+
+            std::shared_ptr<OpenAPI::OAIIntersection_info> get_intersection_info();
+
+            void set_intersection_info(std::shared_ptr<OpenAPI::OAIIntersection_info> &_intersection_info );
```
dan-du-car, line 47: add comments?

streets_utils/streets_vehicle_scheduler/include/vehicle_sorting.h
```diff
@@ -0,0 +1,18 @@
...
+#include "scheduling_exception.h"
+#include "intersection_schedule.h"
+
+#include <vector>
+
+namespace streets_vehicle_scheduler{
+
+	bool departure_position_comparator(const streets_vehicles::vehicle &veh1, const streets_vehicles::vehicle &veh2) ;
+
+	bool st_comparator(const streets_vehicles::vehicle &veh1, const streets_vehicles::vehicle &veh2) ;
```
dan-du-car, line 12: add comments?

streets_utils/streets_vehicle_scheduler/include/vehicle_sorting.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+#include "vehicle.h"
+#include "scheduling_exception.h"
+#include "intersection_schedule.h"
+
+#include <vector>
+
+namespace streets_vehicle_scheduler{
```
dan-du-car, line 8: add comments for each function?

streets_utils/streets_vehicle_scheduler/src/vehicle_scheduler.cpp
```diff
@@ -0,0 +1,81 @@
...
+        if (!is_found) {
+            throw scheduling_exception("No lane " + std::to_string(veh._cur_lane_id) + " found in intersection info!");
+        }
+        return link_lane;
+
+    }
+
+    void vehicle_scheduler::estimate_vehicles_at_common_time( std::unordered_map<std::string,streets_vehicles::vehicle> &vehicles, 
+                                                                const u_int64_t timestamp) {
+        for ( auto map_entry: vehicles) {
```
dan-du-car, line 54: map_entry is a vehicle? This variable name is confusing

streets_utils/streets_vehicle_scheduler/src/vehicle_scheduler.cpp
```diff
@@ -0,0 +1,81 @@
...
+            
+        }
+        if (!is_found) {
+            throw scheduling_exception("No lane " + std::to_string(veh._cur_lane_id) + " found in intersection info!");
+        }
+        return link_lane;
+
+    }
+
+    void vehicle_scheduler::estimate_vehicles_at_common_time( std::unordered_map<std::string,streets_vehicles::vehicle> &vehicles,
```
dan-du-car, line 52: Is this function to remove vehicle path future points that are outdated when scheduling service receives it? Can we make this function name more intuitive?

streets_utils/streets_vehicle_scheduler/src/vehicle_sorting.cpp
```diff
@@ -0,0 +1,38 @@
+#include "vehicle_sorting.h"
+
+namespace streets_vehicle_scheduler {
+    bool departure_position_comparator(const streets_vehicles::vehicle &veh1, const streets_vehicles::vehicle &veh2){
+		// -1 indicates unassigned departure position which should be last when sorted in ascending order
+		if ((veh1._departure_position < 1  && veh2._departure_position < 1) ||  veh1._departure_position < 1) {
```
dan-du-car, line 6: can we explain why veh1 departure position is greater than veh2 when both positions are less than one? or just veh1 position is less than 1?

streets_utils/streets_vehicle_scheduler/src/vehicle_sorting.cpp
```diff
@@ -0,0 +1,38 @@
+#include "vehicle_sorting.h"
+
+namespace streets_vehicle_scheduler {
+    bool departure_position_comparator(const streets_vehicles::vehicle &veh1, const streets_vehicles::vehicle &veh2){
```
dan-du-car, line 4: maybe name veh1 and veh2 to pre_vehicle, cur_vehicle, next_vehicle to indicate those two vehicles are in a list?

streets_utils/streets_vehicle_scheduler/src/vehicle_sorting.cpp
```diff
@@ -0,0 +1,38 @@
...
+		// -1 indicates unassigned departure position which should be last when sorted in ascending order
+		if ((veh1._departure_position < 1  && veh2._departure_position < 1) ||  veh1._departure_position < 1) {
+			return veh1._departure_position > veh2._departure_position;
+		}
+		else {
+			return veh1._departure_position < veh2._departure_position; 
+		}
+	}
+
+    bool st_comparator(const streets_vehicles::vehicle &veh1, const streets_vehicles::vehicle &veh2) {
```
dan-du-car, line 14: The same comments as the above comparator

streets_utils/streets_vehicle_scheduler/src/vehicle_sorting.cpp
```diff
@@ -0,0 +1,38 @@
...
+		// -1 indicates unassigned departure position which should be last when sorted in ascending order
+		if (veh1._actual_st == -1  || veh2._actual_st == -1 ) {
```
dan-du-car, line 16: It does not make sense to validate -1 inside a comparator function. I think we probably can split the function into two? validate -1 through exception only make sense given a particular context.

streets_utils/streets_vehicle_scheduler/src/vehicle_sorting.cpp
```diff
@@ -0,0 +1,38 @@
...
+			if ( veh1._departure_position < 1)
+				throw scheduling_exception("Vehicle " + veh1._id + " has invalid actual st of " + std::to_string(veh1._actual_st) + "!");
+			else if ( veh2._departure_position < 1 )
+				throw scheduling_exception("Vehicle " + veh2._id + " has invalid actual st of " + std::to_string(veh2._actual_st) + "!");
+			return true;
```
dan-du-car, line 21: why is this return true? If the actual_st is invalid, we should not even call this compare.

streets_utils/streets_vehicle_scheduler/src/vehicle_scheduler.cpp
paulbourelly999: it is a pair< std:string , vehicle> since it is stored in a map. I can rename this if necessary

streets_utils/streets_vehicle_scheduler/src/all_stop_vehicle_scheduler.cpp
```diff
@@ -0,0 +1,536 @@
...
+                } else {
+                    // Entering time for RDVs should be max of entering time of previous vehicle and schedule timestamp since they
+                    // can only enter the intersection once they have received a schedule allowing them to enter the intersection
+                    sched.et =  std::max(previously_scheduled_vehicle->et, option.timestamp);
+                    // Set vehicle state. Will not impact clearance time estimation since set on schedule
+                    sched.state =  streets_vehicles::vehicle_state::DV;
+                    // Departure time is estimated clearance time for vehicle and link lane plus entering time
+                    sched.dt = sched.et + estimate_clearance_time( veh, veh_lane );
+                    // Set access to true
+                    sched.access = true;
```
Saeid-Solei, line 436: If an RDV do not have conflict direction with its preceding RDV, then its ET is set to its preceding RDV's ET. But this doesn't mean the RDV should receive access. Access will be granted if ET is equal to the cur_timestamp. Also, the vehicle state will remain RDV unless ET is equal to the cur_timestamp.

streets_utils/streets_vehicle_scheduler/src/vehicle_scheduler.cpp
paulbourelly999: So this function uses kinematics to estimate vehicles at a future time instead of future path information. It then deletes any future path information that is at a time less than or equal to the estimated time. This is because the scheduling service use to approximate vehicle future state using the future point closest to the current time stamp but not exceeding it. This means we could have information for vehicles that differs in timestamp by up to the scheduling frequency. I thought this was a better way to estimate vehicles location in the future.

streets_utils/streets_vehicle_scheduler/src/vehicle_sorting.cpp
paulbourelly999: the vehicle struct initializes departure position to -1 which is an invalid value indicating the vehicle has not yet been scheduled. When we sort vehicles based on departure position we do not want un scheduled vehicles first. We want them last.

streets_utils/streets_vehicle_scheduler/src/vehicle_sorting.cpp
paulbourelly999: It does make sense. Some of these values are not always set, they are initialize as invalid values. This way we don't mistakenly consider a unset vehicles property. It allows our sort function to sort some list which may have invalid values at first (departure_position_comparator) or through an exception when we realize a value that should have been set was never set (st_comparator).

build.sh
```diff
@@ -32,42 +33,44 @@ cmake -DCMAKE_CXX_FLAGS="${COVERAGE_FLAGS}" -DCMAKE_C_FLAGS="${COVERAGE_FLAGS}"
...
 cmake -DCMAKE_CXX_FLAGS="${COVERAGE_FLAGS}" -DCMAKE_C_FLAGS="${COVERAGE_FLAGS}" -DCMAKE_BUILD_TYPE="Debug" ..
 make
-make install
 
 cd /home/carma-streets/message_services
 mkdir build
 cd /home/carma-streets/message_services/build
 cmake -DCMAKE_CXX_FLAGS="${COVERAGE_FLAGS}" -DCMAKE_C_FLAGS="${COVERAGE_FLAGS}" -DCMAKE_BUILD_TYPE="Debug" ..
 make
-make install
```
dan-du-car, line 71: Why is this removed?

streets_utils/streets_vehicle_list/include/vehicle.h
```diff
@@ -15,7 +15,7 @@ namespace streets_vehicles {
 		uint64_t timestamp;
 		/* lanelet id for future point */
 		int lane_id;
```
dan-du-car, line 17: maybe rename this lane_id to lanelet_id to be more specific?

streets_utils/streets_vehicle_list/include/vehicle.h
```diff
@@ -35,46 +35,46 @@ namespace streets_vehicles {
...
+		double _accel_max = 0.0;
+		/* maximum comfortable deceleration rate (it must have a negative value) (m/s^2) */
+		double _decel_max = 0.0;
 		/* vehicle's lane id in the last update */
-		int _cur_lane_id;
+		int _cur_lane_id = 0;
 		/* vehicle's entry lane id */
-		int _entry_lane_id;
+		int _entry_lane_id = 0;
 		/* vehicle's departure lane id */
```
dan-du-car, line 52: maybe rename to lanelet instead of lane? The same for comments?

streets_utils/streets_vehicle_list/include/vehicle.h
```diff
@@ -35,46 +35,46 @@ namespace streets_vehicles {
...
-		int _exit_lane_id;
+		int _exit_lane_id = 0;
 		/* vehicle's connection link id */
-		int _link_id;
+		int _link_id = 0;
 		/* vehicle turn direction in intersection */
 		std::string _direction;
 		/* link lane priority */
 		int _link_priority;
```
dan-du-car, line 59: Could you elaborate the priority? The lowest value is the to priority?

streets_utils/streets_vehicle_list/include/vehicle.h
```diff
@@ -35,46 +35,46 @@ namespace streets_vehicles {
...
 		u_int64_t _cur_time = 0;
-		/* the distance between the vehicle and the end of its lane in the last update */
-		double _cur_distance;
-		/* vehicle's speed in the last update */
-		double _cur_speed;
-		/* vehicle's acceleration in the last update */
-		double _cur_accel;
+		/* the distance between the vehicle and the end of its lane in the last update (m)*/
+		double _cur_distance = 0.0;
+		/* vehicle's speed in the last update (m/s) */
```
dan-du-car, line 74: Not sure the meaning "in the last update"? Is that previous status of the vehicle? or current status of the vehicle?

streets_utils/streets_vehicle_list/src/message_processors/all_stop_status_intent_processor.cpp
```diff
@@ -185,7 +185,7 @@ namespace streets_vehicles {
 			}
 			vehicle._cur_state =  vehicle_state::RDV;
 			vehicle._actual_st = vehicle._cur_time;
```
dan-du-car, line 187: For RDV vehicle, the vehicle actual stopping time is the current epoch time?

streets_utils/streets_vehicle_scheduler/include/all_stop_vehicle_scheduler.h
```diff
@@ -0,0 +1,220 @@
...
+#include "intersection_schedule.h"
+#include "scheduling_exception.h"
+#include "vehicle_sorting.h"
+
+
+namespace streets_vehicle_scheduler {
+    class all_stop_vehicle_scheduler : public vehicle_scheduler {
+        private:
+            
+            std::list<streets_vehicles::vehicle> rdvs_previously_granted_access;
```
dan-du-car, line 17: comments here?

streets_utils/streets_vehicle_scheduler/include/all_stop_vehicle_scheduler.h
```diff
@@ -0,0 +1,220 @@
...
+            /**
+             * @brief Configurable time buffer in milliseconds for estimating the time it takes a vehicle to leave its entry lane 
+             * and enter the intersection box once granted access to the intersection.
```
dan-du-car, line 20: maybe a stopped vehicle at the intersection?

streets_utils/streets_vehicle_scheduler/include/all_stop_vehicle_scheduler.h
```diff
@@ -0,0 +1,220 @@
...
+            uint64_t entering_time_buffer = 0;
+            /**
+             * @brief Schedule all currently Departing Vehicle (DVs). Estimate intersection departure times (dt's) for each vehicle
+             * based on kinematic vehicle information and intersection geometry. Method assumes empty intersection_schedule is passed in.
+             * Method will add vehicle_schedule(s) for each DV in the vector. 
+             * 
+             * @param dvs vector of unscheduled DVs
+             * @param schedule intersection_schedule to add DV scheduling information to.
+             */
+            void schedule_dvs( std::list<streets_vehicles::vehicle> &dvs, intersection_schedule &schedule) const;
```
dan-du-car, line 31: What is `kinematic` vehicle? Is this function to add the dvs which is a list of unscheduled vehicle to the second reference  schedule? Why is this schedule is empty schedule?

streets_utils/streets_vehicle_scheduler/include/all_stop_vehicle_scheduler.h
```diff
@@ -0,0 +1,220 @@
...
+            /**
+             * @brief Schedule all currently Ready to Depart Vehicles (RDVs). Consider all possible RDV departure orders. Select
+             * the one with the least calculated delay. Delay for a given vehicle is the difference between Entering Time and Stopping
```
dan-du-car, line 34: What is the `Delay`?

streets_utils/streets_vehicle_scheduler/include/all_stop_vehicle_scheduler.h
```diff
@@ -0,0 +1,220 @@
...
+             * 
+             * @param dvs vector of unscheduled DVs
+             * @param schedule intersection_schedule to add DV scheduling information to.
+             */
+            void schedule_dvs( std::list<streets_vehicles::vehicle> &dvs, intersection_schedule &schedule) const;
+            /**
+             * @brief Schedule all currently Ready to Depart Vehicles (RDVs). Consider all possible RDV departure orders. Select
+             * the one with the least calculated delay. Delay for a given vehicle is the difference between Entering Time and Stopping 
+             * Time. Method assumes intersection_schedule only populated with DV(s) vehicle_schedules. Method will add vehicle_schedule(s)
+             * for each RDV in vector.
```
dan-du-car, line 36: Not sure why the existing schedule only has schedule for DV. And it is trying to schedule the upcoming unscheduled RDV?

streets_utils/streets_vehicle_scheduler/include/all_stop_vehicle_scheduler.h
```diff
@@ -0,0 +1,220 @@
...
+             * @brief Method to determine whether vehicle was granted access to the intersection in previous schedules. To avoid granting
+             * conflicting vehicles intersection access simultaneously, the scheduling service maintains a list of vehicles it has granted
+             * access to in the past. Vehicles are removed from this list once their update reflects that they have received and processed
+             * this access.
+             * 
+             * @param veh 
+             * @return true if vehicle is in list of vehicles previously granted access.
+             * @return false if vehicle is not in list of vehicles previously granted access.
+             */
+            bool is_rdv_previously_granted_access( const streets_vehicles::vehicle &veh) const;
```
dan-du-car, line 53: Is the vehicle class keeping track of previous status like access? Why not checking whether the vehicle is currently granted access?

streets_utils/streets_vehicle_scheduler/include/all_stop_vehicle_scheduler.h
```diff
@@ -0,0 +1,220 @@
...
+            /**
+             * @brief This method will remove a vehicle (DV) once it from the list of previously granted access RDVs. This method should be
+             * called once a RDV received the schedule update that grants it access to the intersection. This is indicated by the vehicle 
+             * no longer being a RDV and instead being a DV ( @see streets_vehicle_list library).
+             * 
+             * @param veh DV to be removed from list
+             */
+            void remove_rdv_previously_granted_access( const streets_vehicles::vehicle &veh);
```
dan-du-car, line 61: Is this to remove DV from RDV list? or move RDV to DV and update status from RDV to DV?

streets_utils/streets_vehicle_scheduler/include/all_stop_vehicle_scheduler.h
```diff
@@ -0,0 +1,220 @@
...
+            /**
+             * @brief Given a list of vehicles and a vehicle id, this method returns the vehicle with a matching vehicle id if there is one.
+             * This method will throw a scheduling_exception if no vehicle with the given id is found.
+             * 
+             * @param veh_list list to search.
+             * @param veh_id vehicle id to search for.
+             * @return streets_vehicles::vehicle vehicle with matching vehicle id
+             * @throw scheduling_exception if vehicle is not in list.
+             */
+            streets_vehicles::vehicle get_vehicle_with_id( const std::list<streets_vehicles::vehicle> &veh_list, const std::string &veh_id ) const;
```
dan-du-car, line 71: maybe rename to find_vehicle_by_id?

dan-du-car: Why not return a null if not found? and ask caller to check return vehicle object?

streets_utils/streets_vehicle_scheduler/include/intersection_schedule.h
```diff
@@ -0,0 +1,104 @@
...
+         * @brief Write vehicle schedule as rapidjson::Value.
+         * 
+         * @return rapidjson::Value 
+         */
+        rapidjson::Value toJson() const;
+    };
+
+    
+
+    struct intersection_schedule {
```
dan-du-car, line 71: not sure if this name is intuitive.

streets_utils/streets_vehicle_scheduler/include/vehicle_sorting.h
```diff
@@ -0,0 +1,53 @@
...
+	/**
+	 * @brief Comparator for vehicle collections. Used for sorting vehicle objects in a collection in 
+	 * ascending order with respect to their departure position value. Values lower than one are treated
+	 * as unset and order last after the largest positive departure position
+	 * 
+	 * @param prev_veh first vehicle.
```
dan-du-car, line 14: What are these compactor actual comparing against? The pass in parameters are all two vehicles. I am assuming the departure_position is to compare the departure position field from the vehicle? Maybe it is good to update the comments to describe the detail?

streets_utils/streets_vehicle_scheduler/src/all_stop_vehicle_scheduler.cpp
```diff
@@ -0,0 +1,578 @@
...
+            else if ( veh._cur_state == streets_vehicles::vehicle_state::EV) {
+                EVs.push_back(veh);
+            }
+        }
+        
+        SPDLOG_DEBUG("Vehicles to schedule are DVs : {0}, RDVs : {1}, EVs {2} ", DVs.size(), RDVs.size(), EVs.size());
+
+        // Schedule DVs
+        if ( !DVs.empty() )
+            schedule_dvs( DVs, schedule);
```
dan-du-car, line 47: Is there any order among these three schedule calls? DV, RDV, EV?

streets_utils/streets_vehicle_scheduler/src/all_stop_vehicle_scheduler.cpp
```diff
@@ -0,0 +1,578 @@
...
+        
+        
+
+    }
+
+    double all_stop_vehicle_scheduler::estimate_delta_x_prime(const streets_vehicles::vehicle &veh, 
+                                                                const OpenAPI::OAILanelet_info &entry_lane) const{
+        if ( veh._cur_state == streets_vehicles::vehicle_state::EV ) {
+            // Get Entry Lanelet
+            double delta_x_prime = (pow(entry_lane.getSpeedLimit(),2)-pow(veh._cur_speed,2))/(2*veh._accel_max) -
```
dan-du-car, line 63: Maybe point out which equation this refer to?

streets_utils/streets_vehicle_scheduler/src/all_stop_vehicle_scheduler.cpp
```diff
@@ -0,0 +1,578 @@
...
+                pow(entry_lane.getSpeedLimit(), 2)/(2*veh._decel_max); 
+            return  delta_x_prime;
+        }
+        else {
+            SPDLOG_WARN("Cannot estimate delta x prime for vehicle {0} because it's state {1} != EV !" );
+        }
+    }
+
+    double all_stop_vehicle_scheduler::calculate_v_hat(const streets_vehicles::vehicle &veh) const {
+        double numerator = veh._decel_max *( 2 * veh._cur_distance*veh._accel_max + pow(veh._cur_speed, 2));
```
dan-du-car, line 73: If we are getting this equation from the algorithm design doc, please point out the equation number it refers to.

streets_utils/streets_vehicle_scheduler/src/all_stop_vehicle_scheduler.cpp
```diff
@@ -0,0 +1,578 @@
...
+    }
+
+    double all_stop_vehicle_scheduler::calculate_acceleration_time( const streets_vehicles::vehicle &veh, const double v_hat) const {
+        // If v_hat is less than current vehicle speed there is only time for deceleration
+        // Note: This is an error case and means stopping before the stop bar is not possible given current kinematic information
+        // and deceleration limits
+        if ( v_hat < veh._cur_speed ){
+            return 0.0;
+        } 
+        return (v_hat-veh._cur_speed)/veh._accel_max;
```
dan-du-car, line 90: If we are getting this equation from the algorithm design doc, please point out the equation number it refers to.

streets_utils/streets_vehicle_scheduler/src/all_stop_vehicle_scheduler.cpp
```diff
@@ -0,0 +1,578 @@
...
+    }
+
+    double all_stop_vehicle_scheduler::calculate_cruising_time( const streets_vehicles::vehicle &veh, const double v_hat, const double delta_x_prime) const {
+        // If v_hat is less than current speed there is only time for deceleration
+        // Note: This is an error case and means stopping before the stop bar is not possible given current kinematic information
+        // and deceleration limits
+        if ( v_hat < veh._cur_speed ) {
+            return 0.0;
+        }
+        return (veh._cur_distance - delta_x_prime)/v_hat;
```
dan-du-car, line 100: If we are getting this equation from the algorithm design doc, please point out the equation number it refers to.

streets_utils/streets_vehicle_scheduler/src/all_stop_vehicle_scheduler.cpp
```diff
@@ -0,0 +1,578 @@
...
+        // Note: This is an error case and means stopping before the stop bar is not possible given current kinematic information
+        // and deceleration limits
+        if ( v_hat < veh._cur_speed ) {
+            return 0.0;
+        }
+        return (veh._cur_distance - delta_x_prime)/v_hat; 
+    }
+
+    double all_stop_vehicle_scheduler::calculate_deceleration_time( const streets_vehicles::vehicle &veh, const double v_hat) const {
+        return -v_hat/veh._decel_max;
```
dan-du-car, line 104: If we are getting this equation from the algorithm design doc, please point out the equation number it refers to.

streets_utils/streets_vehicle_scheduler/src/all_stop_vehicle_scheduler.cpp
```diff
@@ -0,0 +1,578 @@
...
+    }
+
+    uint64_t all_stop_vehicle_scheduler::estimate_clearance_time( const streets_vehicles::vehicle &veh, 
+                                                                    const OpenAPI::OAILanelet_info &link_lane_info) const{
+        // Clearance time in seconds 
+        double clearance_time = 0;
+        // If vehicle is Departing Vehicle consider its location in the link lanelet.
+        if ( veh._cur_state == streets_vehicles::vehicle_state::DV ) {
+            // Distance covered during constant max acceleration to speed limit
+            double constant_acceleration_delta_x = (pow(link_lane_info.getSpeedLimit(), 2) - pow( veh._cur_speed, 2))/(2* veh._accel_max);
```
dan-du-car, line 480: If we are getting this equation from the algorithm design doc, please point out the equation number it refers to.

streets_utils/streets_vehicle_scheduler/src/all_stop_vehicle_scheduler.cpp
```diff
@@ -0,0 +1,578 @@
...
+        while ( !vehicle_to_be_scheduled_next.empty() );
+    }
+
+    uint64_t all_stop_vehicle_scheduler::estimate_earliest_time_to_stop_bar(const streets_vehicles::vehicle &veh) const{
+        // Distance to stop bar 
+        double delta_x = veh._cur_distance;
+        // Get Entry Lane
+        OpenAPI::OAILanelet_info lane_info =  get_entry_lanelet_info( veh );
+        // Distance necessary to get to max speed and decelerate with decel_max
+        double delta_x_prime =  estimate_delta_x_prime( veh, lane_info );
```
dan-du-car, line 360: If we are getting this equation from the algorithm design doc, please point out the equation number it refers to.

streets_utils/streets_vehicle_list/include/vehicle.h
paulbourelly999: I think this is a pretty trivial detail especially since the comment will clear up any possible confusion. I do not think this is worth changing.

streets_utils/streets_vehicle_list/include/vehicle.h
paulbourelly999: I think this is a pretty trivial detail and does not warrant changing. I don't think lanelet_id and lane_id will confuse anyone

streets_utils/streets_vehicle_list/include/vehicle.h
paulbourelly999: This is a value in the original scheduling algorithm that was never used. Essentially it provides a property that would allow us to give certain lanes priority in scheduling. Again this is not really used but it existed in the UC 1 logic and so I ported  it over to give us the opportunity to use it in the future. If we want to remove it or leave a comment about how it is not used currently, we can do this in another PR with its own issue since this PR is for the scheduling library not the vehicle list library.

streets_utils/streets_vehicle_list/src/message_processors/all_stop_status_intent_processor.cpp
paulbourelly999: No, when the vehicle state changes from EV to RDV the first time, this is considered the stopping time. This should only happen once. Again, this is the streets_vehicle_list library. I just corrected a log statement I saw. If you think there is some problem with the logic here we can write up and issue and I can look into it  but I don't think there is an issue

streets_utils/streets_vehicle_scheduler/include/all_stop_vehicle_scheduler.h
paulbourelly999: I am not sure I understand what this comment is about. Please explain in more detail.

streets_utils/streets_vehicle_scheduler/include/all_stop_vehicle_scheduler.h
paulbourelly999: The schedule is empty because DVs need to be the first vehicles scheduled (added to the schedule).

streets_utils/streets_vehicle_scheduler/include/all_stop_vehicle_scheduler.h
paulbourelly999: This list tracks vehicles we have granted access already. If we don't keep track of this it is possible we can grant two different vehicle with conflicting directions access in two sequential schedules. To explain how this would happen imagine two RDVs waiting to get access. The first time we schedule them we may grant access to one. Since the scheduling frequency is 5 hz it is possible that by the time we schedule again, we still haven't received a new vehicle update with this access reflected in it. Then if our scheduling algorithm instead grants the other one access both will think they have access.

streets_utils/streets_vehicle_scheduler/include/all_stop_vehicle_scheduler.h
paulbourelly999: This is not about changing the state of vehicles. It is about keeping track of vehicles that we have granted access, who's updated does not yet reflect that they have received this access. Once a vehicle update reflects it has received this access the all_stop_status_intent_processor in the streets_vehicle_list library will consider it as DV. At this point the scheduler will check to see if this vehicle is in the list and remove it if it is since it has now confirmed that is has received access.

streets_utils/streets_vehicle_scheduler/include/intersection_schedule.h
paulbourelly999: I think this name is very intuitive. It is the schedule for a intersection. I will leave a comment for the struct to document this as well.

streets_utils/streets_vehicle_scheduler/include/vehicle_sorting.h
paulbourelly999: These functions are necessary in order to use custom logic for std::sort or list::sort. They allow these methods to sort collections this way. I think the comment describes this very well.

streets_utils/streets_vehicle_scheduler/include/all_stop_vehicle_scheduler.h
dan-du-car: There are multiple DVs. When the schedule schedule the second DV, the schedule is not empty right?

streets_utils/streets_vehicle_scheduler/include/all_stop_vehicle_scheduler.h
dan-du-car: Could you add a field in the vehicle class to indicate whether access is confirmed, like is_access_confirmed. So we do not have to keep a separate list for RDVs to be removed because of status change? Maybe just a thought?

streets_utils/streets_vehicle_scheduler/include/intersection_schedule.h
```diff
@@ -0,0 +1,110 @@
...
+         * should ever hold the same index. Negative values are considered invalid but are 
+         * initial values set before vehicle is scheduled.
+         */
+        int dp = -1;
+        /**
+         * @brief Bool flag indicating whether vehicle has access to enter the intersection.
+         */
+        bool access = false;
+        /**
+         * @brief Vehicle State (@see @link vehicle @endlink)
```
dan-du-car, line 50: (@see @link vehicle @endlink)?

streets_utils/streets_vehicle_scheduler/include/intersection_schedule.h
```diff
@@ -0,0 +1,110 @@
...
+         */
+        streets_vehicles::vehicle_state state = streets_vehicles::vehicle_state::ND;
+        /**
+         * @brief Entry lanelet id of vehicle 
+         */
+        int entry_lane =  -1;
```
dan-du-car, line 56: Invalid lanelet id is 0

streets_utils/streets_vehicle_scheduler/include/intersection_schedule.h
```diff
@@ -0,0 +1,110 @@
...
+        /**
+         * @brief Intended lanelet id of link lanelet used to travel through intersection.
+         */
+        int link_id = -1;
```
dan-du-car, line 60: Invalid lanelet id is 0

streets_utils/streets_vehicle_scheduler/include/all_stop_vehicle_scheduler.h
paulbourelly999: A vehicle that is granted access, is by definition a DV. This buffer is for representing the time it will take a DV ( vehicle granted access) to leave the entry lane  and enter the intersection box. The purpose is to give a time estimate to EVs behind this DV when they will be able to stop at the stop bar and become RDV

paulbourelly999: The buffer is applied to estimate the next vehicles (EV) st given this DVs (ET)

streets_utils/streets_vehicle_scheduler/include/all_stop_vehicle_scheduler.h
paulbourelly999: Vehicles need to be scheduled in this order DVs, RDVs, EVs. Each schedule method requires that the previous schedule method is run first (if there are vehicle in the previous state). Otherwise it will produce a invalid schedule.

streets_utils/streets_vehicle_scheduler/include/intersection_schedule.h
dan-du-car: Maybe use a const saying `const INVALID = 0` and use it.

streets_utils/streets_vehicle_scheduler/include/all_stop_vehicle_scheduler.h
paulbourelly999: This is exactly what the list is doing.

streets_utils/streets_vehicle_scheduler/include/all_stop_vehicle_scheduler.h
paulbourelly999: It already has an access field. Once this is true, access is confirmed

streets_utils/streets_vehicle_scheduler/include/intersection_schedule.h
```diff
@@ -0,0 +1,110 @@
...
+         * @brief Vector of all scheduled vehicles
+         */
+        std::vector<vehicle_schedule> vehicle_schedules;
+        /**
+         * @brief Calculate intersection schedule delay. This is the sum of ST-ET for each vehicle schedule.
+         * This represents the cumulative time spent waiting at the stop bar for all vehicles.
+         * 
+         * @return u_int64_t 
+         */
+        u_int64_t get_delay() const;
```
dan-du-car, line 92: If two non-conflicting vehicles stopping at stopline and entering the intersection at the same time, this delay will be sum of both? Why do we need this delay?

streets_utils/streets_vehicle_scheduler/include/vehicle_scheduler.h
```diff
@@ -0,0 +1,100 @@
...
+    class vehicle_scheduler {
+        protected:
+            /**
+             * @brief shared pointer to intersection information required for scheduling.
+             * 
+             */
+            std::shared_ptr<OpenAPI::OAIIntersection_info> intersection_info;
+
+            /**
+             * @brief Limits how much departure position for a given vehicle can change from current reported departure position.
```
dan-du-car, line 29: It is a bit confusing. The given vehicle is DV status vehicle or RDV? What is this limit about?

streets_utils/streets_vehicle_scheduler/include/vehicle_scheduler.h
```diff
@@ -0,0 +1,100 @@
...
+             */
+            int flexibility_limit = 5;
+            /**
+             * @brief Method to estimate all vehicle's speed and location at a common future time. The time needs to be time since epoch in milliseconds.
```
dan-du-car, line 33: Maybe add a note to make sure applications running in all vehicles of interest and intersection should have a synchronized time?

streets_utils/streets_vehicle_scheduler/include/vehicle_scheduler.h
```diff
@@ -0,0 +1,100 @@
...
+             * 
+             * @param vehicles Vehicle for which to calculate future position and speed.
+             * @param timestamp Current or future time in milliseconds since epoch
+             * @throw scheduling_exception if vehicle update time for any vehicle is more recent than timestamp.
+             */
+            void estimate_vehicles_at_common_time( std::unordered_map<std::string,streets_vehicles::vehicle> &vehicles, 
+                                                    const u_int64_t timestamp) const;
+
+            /**
+             * @brief Helper method to get entry lane OAILanelet_info from intersection information given a vehicle.
```
dan-du-car, line 45: based on the given vehicle's current lanelet id? or entry_lane_id, link_lane_id or depart_lane_id?

streets_utils/streets_vehicle_scheduler/include/vehicle_scheduler.h
```diff
@@ -0,0 +1,100 @@
...
+            /**
+             * @brief Destroy the vehicle scheduler object
+             * 
+             */
+            virtual ~vehicle_scheduler() = default;
+            /**
+             * @brief Virtual method to schedule vehicle. Different implementations of vehicle scheduler can implement custom
+             * business logic for scheduling vehicle while taking advantage of polymorphism to dynamically assign class at runtime.
+             * 
+             * @param vehicles unordered map of vehicles to schedule.
```
dan-du-car, line 75: vehicles name is not intuitive. Not sure what is the map key

streets_utils/streets_vehicle_scheduler/include/vehicle_scheduler.h
```diff
@@ -0,0 +1,100 @@
...
+            /**
+             * @brief Set the intersection info object
+             * 
+             * @param _intersection_info 
+             */
+            void set_intersection_info(std::shared_ptr<OpenAPI::OAIIntersection_info> _intersection_info );
+            /**
+             * @brief Set the flexibility limit. This limit controls how much departure position can change for any given vehicle between
+             * successive schedules. A value of 1 represents that vehicles departure position can change by one between two schedules.
+             *
```
dan-du-car, line 94: This is a bit clearer than declaration above. Why would the departure position change between two schedules? Also why only two schedules? There might be more than two schedules for three vehicles at the intersection?

streets_utils/streets_vehicle_scheduler/include/vehicle_scheduler.h
```diff
@@ -0,0 +1,100 @@
...
+             * 
+             * @return std::shared_ptr<OpenAPI::OAIIntersection_info> 
+             */
+            std::shared_ptr<OpenAPI::OAIIntersection_info> get_intersection_info() const;
+            /**
+             * @brief Set the intersection info object
+             * 
+             * @param _intersection_info 
+             */
+            void set_intersection_info(std::shared_ptr<OpenAPI::OAIIntersection_info> _intersection_info );
```
dan-du-car, line 90: const pointer if intersection_info is not to be updated?

streets_utils/streets_vehicle_scheduler/src/vehicle_scheduler.cpp
```diff
@@ -0,0 +1,99 @@
...
+        for (const auto &lanelet : intersection_info->getEntryLanelets() ) {
+            int lane_id =lanelet.getId();
+            if ( lane_id == veh._entry_lane_id ) {
+                entry_lane =  lanelet;
+                is_found = true;
+            }
+            
+        }
+        if (!is_found) {
+            throw scheduling_exception("No lane " + std::to_string(veh._cur_lane_id) + " found in intersection info!");
```
dan-du-car, line 28: "No lane"  to "No entry/enter lane"

streets_utils/streets_vehicle_scheduler/src/vehicle_scheduler.cpp
```diff
@@ -0,0 +1,99 @@
...
+        for ( const auto &lanelet : intersection_info->getLinkLanelets() ) {
+            int lane_id = lanelet.getId();
+            if ( lane_id == veh._link_id ) {
+                link_lane =  lanelet;
+                is_found = true;
+            }
+            
+        }
+        if (!is_found) {
+            throw scheduling_exception("No lane " + std::to_string(veh._cur_lane_id) + " found in intersection info!");
```
dan-du-car, line 46: "No lane "  to "No link lane "

streets_utils/streets_vehicle_scheduler/include/all_stop_vehicle_scheduler.h
paulbourelly999: The list will essentially allow us to consider any RDVs that have been previously granted access as DVs, and therefore schedule them first.

streets_utils/streets_vehicle_scheduler/test/test_all_stop_json_schedule.cpp
```diff
@@ -0,0 +1,123 @@
...
+        }
+    };
+};
+
+
+/**
+ * @brief Test with multiple vehicle for json schedule.
+ */
+TEST_F(all_stop_json_schedule_test, json_schedule){
+    intersection_schedule schedule;
```
dan-du-car, line 56: I think we need to match the test case to the method name in the class. json_schedule to schedule_vehicles

streets_utils/streets_vehicle_scheduler/include/all_stop_vehicle_scheduler.h
```diff
@@ -0,0 +1,228 @@
...
+            double calculate_acceleration_time( const streets_vehicles::vehicle &veh, const double v_hat) const;
+            /**
+             * @brief Method to calculate time interval necessary for deceleration from v_hat to 0 to stop at stop bar. Assumes trajectory
+             * with constant maximum deceleration.
+             * 
+             * @param veh vehicle for which to calculate trajectory.
+             * @param v_hat max speed from which to decelerate from in m/s
+             * @return double time interval necessary for acceleration in seconds.
+             */
+            double calculate_deceleration_time( const streets_vehicles::vehicle &veh, const double v_hat) const;
```
dan-du-car, line 180: Do we need consider vehicle current speed?

streets_utils/streets_vehicle_scheduler/src/all_stop_vehicle_scheduler.cpp
```diff
@@ -0,0 +1,578 @@
...
+        SPDLOG_DEBUG("Delta X Prime = {0}.", delta_x_prime);
+
+        // Calculate v_hat and planned cruising time interval
+        double v_hat;
+        double t_cruising;
+        if ( delta_x >= delta_x_prime ) {
+            v_hat = lane_info.getSpeedLimit();
+            SPDLOG_DEBUG("V hat = {0}.", v_hat);
```
dan-du-car, line 368: Maybe using descriptive full words in the debug log

streets_utils/streets_vehicle_scheduler/src/all_stop_vehicle_scheduler.cpp
```diff
@@ -0,0 +1,578 @@
...
+            // set dp position for current permutation                      
+            sched.dp = starting_departure_position;
+            // Set connection link lanelet id
+            sched.link_id = veh._link_id;
+            // Set entry lanelet id 
+            sched.entry_lane = veh._entry_lane_id;
+            // Get vehicle lane info
+            OpenAPI::OAILanelet_info veh_lane = get_link_lanelet_info(veh);
+            std::shared_ptr<vehicle_schedule> latest_conflicting_vehicle = get_latest_conflicting( veh_lane, option.vehicle_schedules);
+            // If there is no previously scheduled DV
```
dan-du-car, line 424: Two vehicles turning right should also has no conflicts, but they are DVs in the schedule?

streets_utils/streets_vehicle_scheduler/src/all_stop_vehicle_scheduler.cpp
```diff
@@ -0,0 +1,578 @@
...
+            if ( latest_conflicting_vehicle == nullptr) {
+                // Check for scheduled RDVs
+                if ( option.vehicle_schedules.empty() || option.vehicle_schedules.back().access ) {
```
dan-du-car, line 427: option.vehicle_schedules.back().access? The inline comment above does not explain this?

streets_utils/streets_vehicle_scheduler/src/all_stop_vehicle_scheduler.cpp
dan-du-car: The if statement below this comment checking if there is conflicting instead of whether having RV scheduled?

streets_utils/streets_vehicle_scheduler/src/all_stop_vehicle_scheduler.cpp
```diff
@@ -0,0 +1,578 @@
...
+                    sched.access = true;
+                    // Set vehicle state. Will not impact clearance time estimation since set on schedule
+                    sched.state = streets_vehicles::vehicle_state::DV;
+                    // Entering time equals schedule time
+                    sched.et = option.timestamp;
+                    // Departure time equals entering time + clearance time
+                    sched.dt =  sched.et + estimate_clearance_time( veh, veh_lane);
+                }
+                else {
+                    // Can not grant access to vehicle if
```
dan-du-car, line 438: This comment is incomplete

streets_utils/streets_vehicle_scheduler/include/vehicle_scheduler.h
paulbourelly999: I can't see a reason why it would be changed now but in the future where there are difference MAP messages for different times of the day and stuff this could change. I don't think setting this to const offers any substantial benefit.

streets_utils/streets_vehicle_scheduler/include/vehicle_scheduler.h
paulbourelly999: I can add a description in the method comment but other than that I think it is fine

streets_utils/streets_vehicle_scheduler/include/vehicle_scheduler.h
paulbourelly999: I don't think a note like that belongs in the source code. I think that is a deployment documentation thing

streets_utils/streets_vehicle_scheduler/include/intersection_schedule.h
paulbourelly999: The way the scheduling algorithm picks the best schedule is by comparing schedules based on this delay, specifically for RDVs.

streets_utils/streets_vehicle_scheduler/include/all_stop_vehicle_scheduler.h
paulbourelly999: No this is the time it takes for the vehicle to decelerate from v_hat to 0. v_hat is equal to current speed only when there is only enough room to decelerate with max deceleration to stop at the stop bar

streets_utils/streets_vehicle_scheduler/include/intersection_schedule.h
dan-du-car: Can we give it the full  namespace::class definition for this class?

.circleci/config.yml
```diff
@@ -51,12 +51,6 @@ jobs:
             apt-get update  && apt-get install -y cmake git build-essential libboost1.65-all-dev curl unzip gcc-7 g++-7 autotools-dev automake qtbase5-dev qtbase5-dev-tools libqhttpengine-dev libssl-dev libpugixml-dev libgeographic-dev sqlite3 libsqlite3-dev python3-pip
             pip3 install gcovr
 
-            cd /home/carma-streets/streets_utils/streets_api/intersection_client_api
```
paulbourelly999, line 54: Why did you remove this.

dan-du-car:
@paulbourelly999 As we discussed, please add some final tweak on the comments. I approve this for now. Thanks!

dan-du-car (approved):

streets_utils/streets_vehicle_scheduler/include/vehicle_scheduler.h
paulbourelly999: Based on the vehicle object. Each status_intent update contains entry, departure and link lane information. This method will obtain the link lanelet id and search the intersection_info object for the corresponding link lanelet _info object

streets_utils/streets_vehicle_scheduler/include/vehicle_scheduler.h
paulbourelly999: This is between two intersection_schedules. I will clarify that.

streets_utils/streets_vehicle_scheduler/src/all_stop_vehicle_scheduler.cpp
```diff
@@ -0,0 +1,577 @@
...
+            // set est 
+            veh_sched.est = departing_veh._actual_st;
+            // set st
+            veh_sched.st = departing_veh._actual_st;
+            // set et
+            veh_sched.et =  departing_veh._actual_et;
+            // departure time is equal to current time plus clearance time for DVs
+            veh_sched.dt =  schedule.timestamp + clearance_time;
+            // set departure position
+            veh_sched.dp = departing_veh._departure_position;
```
Saeid-Solei, line 129: Setting the departure position of DVs to what it was in the previous time steps will not allow it to be updated. This way, if the DV with position index 1 leaves the intersection box, the DV with position index 2 will remain in position index 2.

streets_utils/streets_vehicle_scheduler/src/all_stop_vehicle_scheduler.cpp
```diff
@@ -0,0 +1,577 @@
...
+
+    }
+
+    void all_stop_vehicle_scheduler::schedule_rdvs( std::list<streets_vehicles::vehicle> &rdvs, intersection_schedule &schedule ) {
+        // Vector of considered intersection_schedule_options
+        std::vector<intersection_schedule> schedule_options;
+        // Sort rdvs ascending order based on departure position
+        rdvs.sort(departure_position_comparator);
+        // Earliest possible departure position is equal to first RDVs departure position
+        int starting_departure_position = rdvs.begin()->_departure_position;
```
Saeid-Solei, line 152: The earliest possible departure position must be the number of existing DVs plus 1.

Saeid-Solei (approved):

PR #142, opened 2022-05-31
Update intersection API lib signal group

dan-du-car:
# PR Details
## Description
Update Intersection API yaml to include MAP intersection information

Design new intersection geometry objects that return lanelet2 and MAP intersection information

Regenerate OpenAPI Server and Client code using swagger codegen

Replace old generated code with new generated code in intersection_client and intersection_model

## Related Issue
NA

## Motivation and Context
UC3

## How Has This Been Tested?
Local integration testing

## Types of changes

## Checklist:


streets_utils/streets_api/intersection_server_api/external/include/qhttpengine/basicauthmiddleware.h
```diff
@@ -0,0 +1,87 @@
+/*
+ * Copyright (c) 2017 Nathan Osman
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to
+ * deal in the Software without restriction, including without limitation the
+ * rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+ * sell copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
```
paulbourelly999, line 10: @dan-du-car , we should not have to include the qhttpengine library headers and binaries. These should be created during the install process.

dan-du-car: https://github.com/usdot-fhwa-stol/carma-streets/blob/4cd4454e10633f475d46f8b386e48401e915dfc7/streets_utils/streets_api/intersection_server_api/CMakeLists.txt
The generated cmakelist is including this lib

dan-du-car: When used this, sudo make install failed with githhub permission issues:
```

ExternalProject_Add(QHTTPENGINE
    GIT_REPOSITORY https://github.com/etherealjoy/qhttpengine.git
    CMAKE_ARGS  -DCMAKE_INSTALL_PREFIX=${EXTERNAL_INSTALL_LOCATION}
                -DCMAKE_PREFIX_PATH=${CMAKE_PREFIX_PATH}
)
```

paulbourelly999: We install qhttpengine as part of the build process in circleCI (https://github.com/usdot-fhwa-stol/carma-streets/blob/298344ee709ea996894136eab40cd737d7e2d055/.circleci/config.yml#L91)

streets_utils/streets_api/intersection_server_api/src/CMakeLists.txt
```diff
@@ -19,9 +19,6 @@ endif(${NODEBUG} STREQUAL "OFF")
 
 find_package(Qt5Core REQUIRED)
 find_package(Qt5Network REQUIRED)
-find_package(qhttpengine REQUIRED)
```
paulbourelly999, line 22: Why is this removed? It is still linking in the lib target

intersection_model/yaml/intersection_model_v1.2.yaml
```diff
@@ -0,0 +1,218 @@
+openapi: 3.0.1
+info:
+  version: 1.0.0
```
paulbourelly999, line 3: I don't know if we need both the old file and the new one. Maybe we can just update the old file version number here and rely on git history for the old one. The reason I say this is because the old yaml will not work with our new server code since there are additional fields. What do you think?

paulbourelly999, line 3: I don't know if we need both the old file and the new one. Maybe we can just update the old file version number here and rely on git history for the old one. The reason I say this is because the old yaml will not work with our new server code since there are additional fields. What do you think?

intersection_model/src/server/src/handlers/IntersectionModelHandler.cpp
```diff
@@ -0,0 +1,151 @@
+#include "IntersectionModelHandler.h"
+#include "intersection_model.h"
+
+namespace IntersectionModelAPI
+{
+    IntersectionModelHandler::IntersectionModelHandler(std::shared_ptr<intersection_model::intersection_model> int_worker) : int_worker_(int_worker)
```
paulbourelly999, line 6: If we want to add DEBUG logging to the handler that may be helpful for debugging. We can even use SPDLOG statements here

paulbourelly999, line 6: If we want to add DEBUG logging to the handler that may be helpful for debugging. We can even use SPDLOG statements here

paulbourelly999:
There are still two code smells related to implicit data conversions that would be nice to resolve.

paulbourelly999:
Also if we can update the README.md for the intersection_model with some description of the parts that make up the intersection_model and maybe the object definitions of the information it exposes. 

Example:
1) intersection model is made up of swagger-generated server code compiled into library, then classes to extend to house custom business logic ( DefaultHandler, DefaultRouter) 
2) Intersection Geometry consists of list of entry, link, and departure lanelets which have the following fields. ( Maybe just copy and past the object definitions from the yaml)

paulbourelly999:
Also if we can update the README.md for the intersection_model with some description of the parts that make up the intersection_model and maybe the object definitions of the information it exposes. 

Example:
1) intersection model is made up of swagger-generated server code compiled into library, then classes to extend to house custom business logic ( DefaultHandler, DefaultRouter) 
2) Intersection Geometry consists of list of entry, link, and departure lanelets which have the following fields. ( Maybe just copy and past the object definitions from the yaml)

streets_utils/streets_api/intersection_client_api/OAIServerConfiguration.h
```diff
@@ -54,7 +54,7 @@ class OAIServerConfiguration {
                 QString value = serverVariable._defaultValue;
 
                 if (!serverVariable._enumValues.empty() && !serverVariable._enumValues.contains(value)) {
-                    throw std::runtime_error(QString("The variable " + name + " in the server URL has invalid value " + value + ".").toUtf8());
+                    throw std::runtime_error(QString("The variable " + name + " in the server URL has invalid value " + value + ".").toUtf8().toStdString());
```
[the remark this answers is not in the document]
dan-du-car: I would leave the generated code as it

intersection_model/yaml/intersection_model_v1.2.yaml
dan-du-car: Updated

intersection_model/src/server/src/handlers/IntersectionModelHandler.cpp
dan-du-car: Updated

streets_utils/streets_api/intersection_server_api/src/handlers/OAIDefaultApiHandler.cpp
```diff
@@ -0,0 +1,74 @@
+/**
```
paulbourelly999, line 1: I know this is part of the generated code but I don't think we need it since we are just creating a library. I think this is generated along with the CMakeList.txt to give and example of how to implement the generated code. @dan-du-car What do you think?

streets_utils/streets_api/intersection_server_api/src/main.cpp
```diff
@@ -0,0 +1,99 @@
+/**
+ * Intersection Model API
+ * No description provided (generated by Openapi Generator https://github.com/openapitools/openapi-generator)
+ *
+ * The version of the OpenAPI document: 1.0.0
+ *
+ * NOTE: This class is auto generated by OpenAPI Generator (https://openapi-generator.tech).
```
paulbourelly999, line 7: I know this is generated code but do we need it since we are compiling this into a library. I think this code is only generated to show an example implementation of  the generated

intersection_model/readme.md
```diff
@@ -123,6 +123,9 @@ RUN make
 Navigate to /carma-streets/ directory
 docker build -t intersection_model -f intersection_model/Dockerfile .
 ```
+## API definition
+1. intersection model is made up of swagger-generated server code compiled into library, then classes to extend to house custom business logic ( DefaultHandler, DefaultRouter). The current yaml file is located [here](yaml/intersection_model_v1.2.yaml).
+2. Intersection Geometry consists of list of entry, link, and departure lanelets which have the following fields. ( Maybe just copy and past the object definitions from the yaml)
```
paulbourelly999, line 128: Remove the parenthesis and add the object definition maybe like this 
```
 lanelet_info:
    description: The information of lanelet 
    type: object
    properties:
      id:
        type: integer
      speed_limit:
        type: number
        description: Unit of measure is m/s
      conflict_lanelet_ids:
        type: array
        items:
          type: integer
        description: List of unique identifers for lanelets that have conflicts
      length:
        type: number
        description: The length of lanelet. Unit of measure is meter
      turn_direction:
        type: string
        description: Turn direction of intersection lane
      signal_group_id: 
        type: integer
        description: The matching signal group send by the SPAT message for this lanelet
      connecting_lanelet_ids:
        description: List of unique identifers for following lanelets that are connecting the depature and entry lanelets
        type: array
        items:
          type: integer
```

intersection_model/readme.md
```diff
@@ -123,6 +123,9 @@ RUN make
 Navigate to /carma-streets/ directory
```
paulbourelly999, line 123: Update README library description. This is not a library but a service used the expose intersection geometry information from MAP and lanelet2 osm map to carma-streets services that require  this information via REST server. Just a couple of sentences should be good.

intersection_model/src/server/src/handlers/IntersectionModelHandler.cpp
```diff
@@ -0,0 +1,151 @@
...
+                    lanelet_info.setId(itr->id);
+                    lanelet_info.setSpeedLimit(itr->speed_limit);
+                    lanelet_info.setLength(itr->length);
+                    res.push_back(lanelet_info);
+                }
+                reqObj->getConflictLaneletsResponse(res);
+            }
+            else
+            {
+                QString err_msg = "Invalid link lanelet id parameter.";
```
paulbourelly999, line 144: Do we want to log error responses? Just a suggestion if you think it would be good for debugging.

dan-du-car: I think it makes more sense for the client to get this error message and log it

PR #144, opened 2022-06-04
Update intersection model map msgs

dan-du-car:
# PR Details
## Description
Standup Kafka client to consume MAP msgs or Setup configurations to: 

Include MAP information in intersection geometry

Describe relationship between MAP and Lanelet2 information (Lane IDs)

Integration test intersection_model  with service that uses intersection_client

## Related Issue
NA

## Motivation and Context
UC3

## How Has This Been Tested?
local integration and unit test

## Types of changes

## Checklist:


intersection_model/include/intersection_model.h
```diff
@@ -199,6 +211,77 @@ namespace intersection_model
...
             */
             bool update_vehicle_routing_graph();
+            
+            /***
+             * @brief Update intersection info structure with incoming map msg consumed by kafka map consumer
+             * @param intersection_map populated with MAP message published by v2xhub. This map message describes intersection geometry and signal group ids
+             * @return True if the intersection info is updated, otherwise false
+             **/
+            bool update_intersecion_info_by_map_msg(const std::shared_ptr<intersection_map> int_map_msg);
+            void mapping_lanelet_id_2_lane_id(const map_referencepoint& ref_point, const map_lane& lane, const std::vector<lanelet::ConstLanelet>& subj_lanelets, std::unordered_map<long,lanelet::ConstLanelet>& lane2lanelet_m) const;
```
paulbourelly999, line 221: Needs a java-doc comment

dan-du-car: Addressed

intersection_model/include/intersection_model.h
```diff
@@ -199,6 +211,84 @@ namespace intersection_model
...
+             * @param lon 
+             * @param elev 
+             * @return lanelet::BasicPoint3d 
+             */
+            lanelet::BasicPoint3d gps_2_map_point(double lat, double lon, double elev ) const;
+
+            /**
+             * @brief Transform the map point to gps location
+             * @param x coordinate
+             * @param y coordinate
```
paulbourelly999, line 258: Units?

intersection_model/test/test_intersection_model.cpp
```diff
@@ -127,16 +241,160 @@ TEST(intersection_model_test, update_intersection_info)
...
+TEST(intersection_model_test, get_link_lanelet_by_id)
+{
+    intersection_model::intersection_model model;
+     model.read_lanelet2_map("../../sample_map/town01_vector_map_test.osm");
+    ASSERT_TRUE(model.update_intersection_info());
+    ASSERT_EQ(0, model.get_link_lanelet_by_id(168).id());
+    ASSERT_EQ(169, model.get_link_lanelet_by_id(169).id());
+}
+
+TEST(intersection_model_test, find_link_lanelet_id_by_enter_depart_lanelet_ids)
```
paulbourelly999, line 393: Brief comments to describe test case might be useful.

dan-du-car: The test case should verify the function. The description/java-doc comments are located  where the function declared. Not needed here.

intersection_model/include/intersection_model.h
dan-du-car: Addressed https://github.com/usdot-fhwa-stol/carma-streets/pull/144/commits/cf53ef6fcb506240d00e4e1cc7d81411a66d6d94

intersection_model/include/map_msg_worker.h
```diff
@@ -0,0 +1,62 @@
...
+#include <iostream>
+#include "kafka_client.h"
+#include "map.h"
+
+namespace intersection_model
+{
+    class map_msg_worker
+    {
+    private:
+        // Kafka consumer configuration
```
paulbourelly999, line 21: So implement MAP message processing for kafka is awesome but kind of makes more work for us in comparison to just implementing reading a MAP file. This is because it means we will have to write additional code in the CARMA-Streets plugin to send MAP messages to kafka on startup. I don't want to necessary abandon any work you have already done here but it may be worth while to just implement it as file reading now so we don't have to do anything on V2X-Hub. What do you think?

intersection_model/include/map_msg_worker.h
```diff
@@ -0,0 +1,62 @@
...
+        std::string map_msg_group_id;
+        std::string map_msg_topic_name;
+        std::shared_ptr<kafka_clients::kafka_consumer_worker> _map_msg_consumer;
+        std::mutex intersection_map_msg_mtx;
+        std::shared_ptr<intersection_map> _intersection_map_ptr;
+
+    public:
+        map_msg_worker() = default;
+        explicit map_msg_worker(const int intersection_id);
+        map_msg_worker(const int intersection_id, const std::string &bootstrap_server, const std::string &map_msg_group_id, const std::string &map_msg_topic_name);
```
paulbourelly999, line 32: Probably want to add comments describing these constructor parameters.

intersection_model/include/intersection_model.h
```diff
@@ -55,6 +58,7 @@ namespace intersection_model
         double length                   = 0.0;  //Unit of meter
         int32_t signal_group_id         = 0;    //0 = Not available
```
paulbourelly999, line 59: Should we have comments describing what this property is?

intersection_model/include/intersection_model.h
```diff
@@ -116,32 +128,32 @@ namespace intersection_model
...
-            const std::vector<lanelet_info_t> get_link_lanelets() const ;
+            const std::vector<lanelet_info_t>& get_link_lanelets_info() const ;
 
             /**
              * @brief Get list of conflict lanelet ids relative to the given link lanelet id
              * @param sub_link_lanelet_id is validated, and has to be link lanelet within an intersection
              * @return A vector of lanelet information struct
              * **/
-            const  std::vector<lanelet_info_t> get_conflict_lanelets(int64_t sub_link_lanelet_id);
+            std::vector<lanelet_info_t> get_conflict_lanelets_info(int64_t sub_link_lanelet_id);
```
paulbourelly999, line 150: Why do we return all other lanelet vectors as constant references except this one.

intersection_model/include/intersection_model.h
dan-du-car: It is self-explanatory.

intersection_model/include/intersection_model.h
paulbourelly999: Also should this method be constant

intersection_model/include/intersection_model.h
```diff
@@ -177,7 +189,7 @@ namespace intersection_model
              * Based on the link lanelets, it retrieves the following depature lanelets
              * @return true if the update for link lanelet and departure lanelet is successful.
              * */
-            bool update_link_departure_lanelets_by_entry_lanelet(const lanelet::Lanelet &entry_lanelet);
+            bool update_link_departure_lanelets_info_by_entry_lanelet(const lanelet::Lanelet &entry_lanelet);
```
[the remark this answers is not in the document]
dan-du-car: Addressed

intersection_model/include/intersection_model.h
```diff
@@ -199,6 +211,84 @@ namespace intersection_model
...
+
+            /**
+             * @brief Convert the centerline geometry of lane defined in J2735 MAP to a list of lanelet2 points in the OSM map.
+             * @param ref_point GPS location for a reference point inside an intersection
+             * @param lane defind in the J2735 MAP message 
+             * @return The list of lanelet2 map points
+             */
+            std::vector<lanelet::BasicPoint3d>  convert_lane_path_2_basic_points(const map_referencepoint& ref_point, const map_lane& lane) const;
+
+            /**
```
paulbourelly999, line 238: Is a line string the same as a list or points that form a line?

intersection_model/include/intersection_model.h
```diff
@@ -199,6 +211,84 @@ namespace intersection_model
...
+            
+            /***
+             * @brief Update intersection info structure with incoming map msg consumed by kafka map consumer
+             * @param intersection_map populated with MAP message published by v2xhub. This map message describes intersection geometry and signal group ids
+             * @return True if the intersection info is updated, otherwise false
+             **/
+            bool update_intersecion_info_by_map_msg(const std::shared_ptr<intersection_map> int_map_msg);
+            /**
+             * @brief Comparing the geometry between lane in the MAP message and geometry of the list of lanelets to determine which lanelet this lane path fall into.
+             * @param ref_point Reference location (latitude, longitude, elevation) of the subsequent data points
```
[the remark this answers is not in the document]
dan-du-car: Reference location (latitude, longitude, elevation) of the subsequent data points

intersection_model/include/intersection_model.h
dan-du-car: yes

intersection_model/src/intersection_model.cpp
```diff
@@ -340,20 +364,270 @@ namespace intersection_model
...
      
     const std::shared_ptr<lanelet::routing::RoutingGraph> intersection_model::get_vehicleGraph_ptr()  const
     {
         return this->vehicleGraph_ptr;
     }
 
+    bool intersection_model::update_intersecion_info_by_map_msg(const std::shared_ptr<intersection_map> int_map_msg)
+    {
+        bool is_updated = false;
+        SPDLOG_INFO("Intersection id {0} has {1} number of geometries.", int_map_msg->intersectionid, int_map_msg->geometries.size());
```
paulbourelly999, line 380: Does one geometry correspond to one intersection are is that something different.

intersection_model/src/map_msg_worker.cpp
```diff
@@ -0,0 +1,152 @@
...
+        bool is_updated = false;
+        std::unique_lock<std::mutex> lck(intersection_map_msg_mtx);
+        if (_intersection_map_ptr == nullptr)
+        {
+            SPDLOG_CRITICAL("Intersection map msg is nullptr");
+            return is_updated;
+        }
+        rapidjson::Document doc = parse_JSON_string(payload);
+        // Map data
+        if (!doc.HasParseError() && doc.FindMember("map_data")->value.IsObject())
```
paulbourelly999, line 50: I don't want to create extra work for you here but in the future we should probably split this parsing logic into different helper methods so that we don't have  so many nested if/for/try blocks. This would make the logic easier to read and maintain in my opinion.  I think for the MAP message its fine right now but in the future it would be nice to have more readable parsing logic.

dan-du-car: That might be a good idea

intersection_model/src/server/src/main.cpp
```diff
@@ -51,7 +51,30 @@ void catchUnixSignals(QList<int> quitSignals) {
 }
 #endif
 
-int main(int argc, char * argv[])
+void intersection_model_event_update(std::shared_ptr<intersection_model::intersection_model> model, std::shared_ptr<intersection_model::map_msg_worker> map_msp_worker)
```
paulbourelly999, line 54: We probably want to add intersection_model/server/src into sonar properties as a source code location for intersection_model. It wasn't previously there since it use to hold generated could but now this is custom code so we probably want sonar-scanner analysis.

intersection_model/include/intersection_model.h
```diff
@@ -155,13 +167,13 @@ namespace intersection_model
...
              **/
-            double get_speed_limit_by_lanelet(const lanelet::ConstLanelet& subj_lanelet);
+            double get_speed_limit_by_lanelet(const lanelet::ConstLanelet& subj_lanelet) const;
 
             /**
              * @brief Get the list of link lanelets that has the same entry lanelet as the subject link lanelet
              * @return A vector of link lanelet ids
              **/
-            std::vector<int> get_shared_entry_link_lanelet_ids(int64_t subject_link_lanelet_id);
+            const std::vector<int> get_shared_entry_link_lanelet_ids(int64_t subject_link_lanelet_id);
```
paulbourelly999, line 176: This should be a constant method too right?

dan-du-car: Addressed

intersection_model/src/server/src/main.cpp
dan-du-car: I think the top package level should allow the sonar scanner to scan the subsequent folders. If that is not the case, how and where exactly you want to update this?

intersection_model/include/intersection_model.h
dan-du-car: It seems that it is actually updating the object. so it cannot be a const method.

intersection_model/src/intersection_model.cpp
```diff
@@ -340,20 +364,270 @@ namespace intersection_model
...
+        double lat = static_cast<double>(ref_point.latitude) * TENTH_TO_ONE * MICRO_DEGREE_TO_DEGREE;
+        double lon = static_cast<double>(ref_point.longitude) * TENTH_TO_ONE * MICRO_DEGREE_TO_DEGREE; 
+        double elev = static_cast<double>(ref_point.elevation) * DM_TO_M; 
+        const auto&  ref_point3d = gps_2_map_point(lat, lon, elev);
+        double cur_x = ref_point3d.x();
+        double cur_y = ref_point3d.y();
+        double cur_z = ref_point3d.z();
+        for(const auto&  node : lane.nodes)
+        {
+            cur_x += static_cast<double>(node.x) * CM_TO_M;
```
paulbourelly999, line 524: So each note in the map is a relative location to the first reference node. Is that correct. Maybe some in line comments might help make that clear.

intersection_model/src/intersection_model.cpp
```diff
@@ -340,20 +364,270 @@ namespace intersection_model
...
+            return link_lanelet_id_intersect_v.front();
+        }  
+        SPDLOG_DEBUG("NO matching link lanelet for enter lanelet id = {1} and departure lanelet id = {2}", link_lanelet_id_intersect_v.front(),enter_lanelet_id, depart_lanelet_id);
+        return lanelet::InvalId;
+    }
+
+    void intersection_model::mapping_lane_id_2_lanelet_id(const map_referencepoint& ref_point, const map_lane& lane, const std::vector<lanelet::ConstLanelet>& subj_lanelets, std::unordered_map<long , lanelet::ConstLanelet>& lane2lanelet_m) const
+    {    
+        std::vector<lanelet::BasicPoint3d> basic_points = convert_lane_path_2_basic_points(ref_point, lane);
+        std::unordered_map<lanelet::Id, double> lanelet2lane_path_distance_m;
```
paulbourelly999, line 493: This is a map of lanelet ids and average distance from the basic_points line. Is that correct?

intersection_model/src/intersection_model.cpp
```diff
@@ -340,20 +364,270 @@ namespace intersection_model
...
+            SPDLOG_DEBUG("Lanelet id {0} to lane id = {1} path points average distance {2}", subj_l.id(), lane.lane_id, avg_distance);
+            lanelet2lane_path_distance_m.insert({subj_l.id(), avg_distance});   
+        }
+        //Find the nearest lanelet from the lane path by shortest average distance
+        const auto&  min_distance_pair = std::min_element(lanelet2lane_path_distance_m.begin(), lanelet2lane_path_distance_m.end(), [](const auto& l, const auto& r){return l.second < r.second; });
+        for(const auto& subj_l: subj_lanelets)
+        {
+            if(subj_l.id() == min_distance_pair->first)
+            {
+                lane2lanelet_m.insert({lane.lane_id, subj_l});
```
paulbourelly999, line 506: What does insert method for a ConstLanelet do?

dan-du-car: It is inserting into constlanelet.

dan-du-car: The std::map value is const as we are not modifying it at the caller

intersection_model/src/intersection_model.cpp
```diff
@@ -340,20 +364,270 @@ namespace intersection_model
...
+        std::vector<lanelet::BasicPoint3d> basic_points = convert_lane_path_2_basic_points(ref_point, lane);
+        std::unordered_map<lanelet::Id, double> lanelet2lane_path_distance_m;
+        for(const auto&  subj_l: subj_lanelets)
+        {
+            double avg_distance = compute_points_2_lanelet_avg_distance(basic_points, subj_l);
+            SPDLOG_DEBUG("Lanelet id {0} to lane id = {1} path points average distance {2}", subj_l.id(), lane.lane_id, avg_distance);
+            lanelet2lane_path_distance_m.insert({subj_l.id(), avg_distance});   
+        }
+        //Find the nearest lanelet from the lane path by shortest average distance
+        const auto&  min_distance_pair = std::min_element(lanelet2lane_path_distance_m.begin(), lanelet2lane_path_distance_m.end(), [](const auto& l, const auto& r){return l.second < r.second; });
```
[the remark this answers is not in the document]
dan-du-car: I think this is a valid point. I am updating it to consider to lane width as well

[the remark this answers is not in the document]
dan-du-car: https://github.com/usdot-fhwa-stol/carma-streets/pull/144/commits/863096c4a1b7f5e056295a86aa5fb0c40ef70443

intersection_model/src/intersection_model.cpp
```diff
@@ -340,20 +364,280 @@ namespace intersection_model
...
+    {
+        bool is_updated = false;
+        SPDLOG_INFO("Intersection id {0} has {1} number of geometries.", int_map_msg->intersectionid, int_map_msg->geometries.size());
+        if(int_map_msg->geometries.empty())
+        {
+            return false;
+        }  
+        // Assuming the map msg only has one geometry object
+        const auto& map_msg_geometry = int_map_msg->geometries.front();
+        std::unordered_map<long, lanelet::ConstLanelet> entry_lane2lanelet_m;
```
paulbourelly999, line 387: What are these maps for. I see one seems to be entry lanes objects to lane ids from the map.

intersection_model/include/intersection_model.h
```diff
@@ -199,6 +211,85 @@ namespace intersection_model
...
+            bool update_intersecion_info_by_map_msg(const std::shared_ptr<intersection_map> int_map_msg);
+            /**
+             * @brief Comparing the geometry between lane in the MAP message and geometry of the list of lanelets to determine which lanelet this lane path fall into.
+             * @param ref_point Reference location (latitude, longitude, elevation) of the subsequent data points
+             * @param lane MAP message lane definition includes the lane geometry, unique lane id, and connections if entry lane.
+             * @param lane_width MAP message lane definition includes the lane width.
+             * @param subj_lanelets List of lanelets
+             * @param lane2lanelet_m A mapping reference to update with lane id and its relevant lanelet id
+             */
+            void mapping_lane_id_2_lanelet_id(const map_referencepoint& ref_point, const map_lane& lane, const long lane_width, const std::vector<lanelet::ConstLanelet>& subj_lanelets, std::unordered_map<long,lanelet::ConstLanelet>& lane2lanelet_m) const;
```
paulbourelly999, line 229: Since this method is returning values by reference ( I mean its update/populating the param lane2lanelet_m and passing it back by reference. It might be helpful to state what the incoming condition of this map is expected to be and what the method does to populate it.

intersection_model/src/intersection_model.cpp
```diff
@@ -340,20 +364,280 @@ namespace intersection_model
...
+        std::unordered_map<long, std::vector<map_connection>> entry_lane2connections_m;
+        std::unordered_map<long, map_lane> link_departure_lanes_m;
```
paulbourelly999, line 389: Are all map lane objects the equivalent of departure lanes from the j2735 MAP?

intersection_model/src/intersection_model.cpp
```diff
@@ -340,20 +364,280 @@ namespace intersection_model
...
+                //Mapping MAP message lane id to lanelet id from OSM map
+                mapping_lane_id_2_lanelet_id(map_msg_geometry.refpoint, map_msg_lane, map_msg_geometry.approach.width, this->entering_lanelets, entry_lane2lanelet_m);
+                entry_lane2connections_m.insert({map_msg_lane.lane_id,  map_msg_lane.connection});
+                continue;
+            }
+            link_departure_lanes_m.insert({map_msg_lane.lane_id, map_msg_lane});
+        }     
+        
+        //Matching enter lanelet, departure lanelet to signal group id
+        std::vector<signalized_intersection_lanelets> enter_departure_lanelets2SG_id_v;
```
paulbourelly999, line 410: Are you mapping signal Group id to a pair of entry/departure lanelets rather than the link lanelet that connects them here. I am a little confused.

intersection_model/src/intersection_model.cpp
```diff
@@ -340,20 +364,280 @@ namespace intersection_model
...
+        //Update intersection link lanelet info with signal group id
+        bool is_signal_group_updated = false;   
+        std::for_each(this->int_info.link_lanelets_info.begin(), this->int_info.link_lanelets_info.end(), [&](auto& ref){
+            auto find_itr = link_lanelet2signal_group_id_m.find(ref.id);
+            if(find_itr != link_lanelet2signal_group_id_m.end())
+            {
+                ref.signal_group_id = link_lanelet2signal_group_id_m[ref.id];
+                is_signal_group_updated = true; 
+            }
+        });
```
paulbourelly999, line 456: Ok if I understand the logic correctly in this method, you are first finding the mapping between entry lane and entry lanelet using the center line provided by the J2735 MAP message and calculating an average distance from the entry-lanelet center line ( I saw you implemented lane width but it still looks like there is no minimum threshold so if the closest entry lanelet has an average distance of 10,000 m you still assign it. Then you map departure lanes to departure lanelets. Then you use these mappings to create a map of entry/departure lanelets to signal groups and from there you pass the information to the connecting lanelets. 

I assume you do this because the MAP message only describes enter or approach lanes and not connecting lanes geometry. Is any of this accurate.

paulbourelly999 (changes requested):
I think I am largely fine with this review. One thing that would make this more understandable accompanying any maps or collections you set up in methods with a description of their purpose. Also for any methods that are pass by reference and change the value of the reference, it would help if inline and method comments reflected the intention of what you are changing and maybe some information about what is assumed about the input. All together this looks very good. I left some comments about the lane to lanelet mapping method since it seems to just map each lane to its closest lanelet regardless of how large a distance we find in between them. If we did have MAP and osm files that have the same intersection name but represent different locations it seems you implementation would read them and map them despite large average distances between them. Might be reasonable to throw and exception if the lowest average distance between the two centerlines exceeds 5 meter or something (Maybe even a larger value) but we can put this off as an improvement if you think that is better.  I am ok merging this as soon as you just update the method comments that modify reference and add some inline comments for the unordered maps you create and what they are for.

intersection_model/src/intersection_model.cpp
dan-du-car: Please refer to J2735 MAP message definition. This is streets version map to mimic that.

intersection_model/include/intersection_model.h
dan-du-car: It is initialized before pass into this function. Please refer to the cpp file to see the usage.

intersection_model/src/intersection_model.cpp
dan-du-car: The map object does not state which type of lane. The information from the MAP message will tell you what type of lane it is.

intersection_model/src/intersection_model.cpp
dan-du-car: Each entry and departure combination map to one link lanelet.

paulbourelly999 (approved):

PR #147, opened 2022-06-16
issue #149: Implement Traffic Signal Controller Component

adev4a:
# PR Details
This PR adds the traffic signal controller client. Ref #149
## Description
Features developed within this implementation include using the net-snmp library to GET and SET snmp requests.

## How Has This Been Tested?
Unit tested and Integration tested

## Types of changes

## Checklist:


tsc_client_service/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
...
+target_link_libraries(${PROJECT_NAME} ${PROJECT_NAME}_lib)
+
+
+#############
+## Testing ##
+#############
+find_package(GTest REQUIRED)
+file(GLOB_RECURSE TEST_SOURCES LIST_DIRECTORIES false test/*.h test/*.cpp)
+add_executable(${PROJECT_NAME}_test
+    test/test_traffic_signal_controller_service
```
paulbourelly999, line 48: please correct this to only include {TEST_SOURCES}

tsc_client_service/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
...
+find_package(streets_service_base_lib COMPONENTS streets_service_base_lib REQUIRED)
+find_library(NETSNMPAGENT "netsnmpagent")
+find_library(NETSNMPMIBS "netsnmpmibs")
+find_library(NETSNMP "netsnmp")
+
+add_library(${PROJECT_NAME}_lib
+        src/snmp_client.cpp)
+
+add_executable( ${PROJECT_NAME} 
+src/main.cpp)
```
paulbourelly999, line 18: Small revision, consistent indentation here.

tsc_client_service/include/snmp_client.h
```diff
@@ -0,0 +1,79 @@
+#pragma once
+#define SPDLOG_ACTIVE_LEVEL SPDLOG_LEVEL_TDEBUG
```
paulbourelly999, line 2: This looks like a type. This should probably be SPDLOG_LEVEL_DEBUG

tsc_client_service/include/snmp_client.h
```diff
@@ -0,0 +1,79 @@
...
+
+#include <net-snmp/net-snmp-config.h>
+#include <net-snmp/net-snmp-includes.h>
+#include <fstream>
+
+#include <spdlog/spdlog.h>
+#include <spdlog/async.h> //support for async logging.
+#include <spdlog/sinks/daily_file_sink.h> // support for dailty file sink
+#include <spdlog/sinks/stdout_color_sinks.h> // or "../stdout_sinks.h" if no colors needed
+#include <rapidjson/document.h>
```
paulbourelly999, line 12: Please review these include statements I don't think you need them here. They are already included by the streets_service_base header files. Please only include headers, packages you are using. This way we can make sure we are only ever including the things we directly need.

tsc_client_service/include/snmp_client.h
```diff
@@ -0,0 +1,79 @@
...
+# include "streets_configuration.h"
+
+class TrafficSignalControllerService
+{
+    private:
+
+        // variable to store an snmp session
+        // struct that holds information about who we're going to be talking to
+        // We need to declare 2 of these, one to fill info with and second which is 
+        // a pointer returned by the library
```
paulbourelly999, line 27: please use java-doc style comments . For example
```
/**
* @brief <Description>
* @param <Parameter Description>
* @return < Return Description>
*/
```

paulbourelly999: Also make sure to include at least a single short sentence to describe any member variable in a java doc comment that is not completely obvious like timestamp or something.

tsc_client_service/manifest.json
```diff
@@ -0,0 +1,37 @@
...
+        {
+           "name": "port",
+           "value": 6053,
+           "description": "Target device NTCIP port",
+           "type": "INTEGER" 
+        },
+        {
+            "name": "community",
+            "value": "public",
+            "description": " ",
```
paulbourelly999, line 21: Add descriptions for all your configuration parameters please.

tsc_client_service/include/snmp_client.h
adev4a: This is just additional info from the net-snmp doc, i have those defined for the functions themselves

tsc_client_service/include/snmp_client.h
adev4a: Looks like both of these compile and give debug logs. The documentation however says its SPDLOG_LEVEL_DEBUG @abey-yoseph you'll have to fix this in yours as well

tsc_client_service/include/snmp_client.h
```diff
@@ -0,0 +1,74 @@
...
+
+        /* net-snmp version definition: SNMP_VERSION_1:0 SNMP_VERSION_2c:1 SNMP_VERSION_2u:2 SNMP_VERSION_3:3 
+        https://github.com/net-snmp/net-snmp/blob/master/include/net-snmp/library/snmp.h */
+        int snmp_version_ = 0;
+        int timeout_ = 10000;
+
+    public:
+        /** @brief Constructor for Traffic Signal Controller Service client.
+         *  Uses the arguments provided to establish an snmp connection
+         * @param ip The ip ,as a string, for the host to establish snmp communication with.
```
dan-du-car, line 45: Which host? The traffic signal controller or the machine running carma-streets?

tsc_client_service/include/snmp_client.h
```diff
@@ -0,0 +1,74 @@
...
+         * @param port Target port as integer on the host for snmp communication.
+         * @param community The community id as a string. Defaults to "public" if unassigned.
```
dan-du-car, line 47: Why do we need community?

tsc_client_service/include/snmp_client.h
```diff
@@ -0,0 +1,74 @@
...
+         * @param timeout The time in microseconds after which an snmp session request expires. Defaults to 100 if unassigned
+         * **/
+        TrafficSignalControllerService(const std::string& ip, const int& port, const std::string& community = "public", int snmp_version = 0, int timeout = 100);
+        
+        /*Disable default copy constructor*/
+        TrafficSignalControllerService() = delete;
+
+        /** @brief Returns integer value for input OID at the Traffic Signal Controller.
+         *  @param input_oid The OID to request information for.
+         *  @return Integer value at the oid, returns -1 if value cannot be requested or oid doesn't have an integer value to return.*/
```
dan-du-car, line 59: It is returning Boolean in the method signature.

tsc_client_service/src/snmp_client.cpp
```diff
@@ -0,0 +1,176 @@
...
+    
+
+    // Send the request
+    status = snmp_synch_response(ss, pdu, &response);
+
+    // Check response
+    if(status == STAT_SUCCESS && response->errstat == SNMP_ERR_NOERROR) {
+        
+        SPDLOG_INFO("STAT_SUCCESS, received a response");
+        for(auto vars = response->variables; vars; vars = vars->next_variable){
```
dan-du-car, line 78: The condition of this loop is vars? Do we need to compare with an end variable?

tsc_client_service/src/snmp_client.cpp
```diff
@@ -0,0 +1,176 @@
...
+    return true;
+}
+
+bool TrafficSignalControllerService::process_snmp_set_request(std::string input_oid, int value){
+
+    SPDLOG_DEBUG("Attemping to SET value for {0}", input_oid, " to {1}", value);
+    // Create pdu for the data
+    pdu = snmp_pdu_create(SNMP_MSG_SET);
+
+    if(!read_objid(input_oid.c_str(), OID, &OID_len)){
```
dan-du-car, line 129: Maybe add some inline comments to elaborate this function?

tsc_client_service/src/snmp_client.cpp
```diff
@@ -0,0 +1,176 @@
...
+        // If oid cannot be created
+        SPDLOG_ERROR("OID could not be created from input: {0}", input_oid);
+        snmp_perror(input_oid.c_str());
```
dan-du-car, line 132: Maybe add some inline comments to elaborate this function read_objid()?

tsc_client_service/src/snmp_client.cpp
```diff
@@ -0,0 +1,176 @@
...
+        if (status == STAT_SUCCESS)
+        {
+            SPDLOG_ERROR("Variable type: {0}",response->variables->type);
+            SPDLOG_ERROR("Error in packet.{0} ", static_cast<std::string>(snmp_errstring(static_cast<int>(response->errstat))));
+        }
+        else if (status == STAT_TIMEOUT){ 
+            SPDLOG_ERROR("Timeout, no response from server");
+        }
+        else{
+            snmp_sess_perror("snmpset", ss);
```
dan-du-car, line 163: Why do we need to call this snmp_sess_perror() after we know it is unsuccessful?

tsc_client_service/src/snmp_client.cpp
```diff
@@ -0,0 +1,176 @@
...
+            SPDLOG_ERROR("Unknown SNMP Error");
+        }
+
+        return false;
+            
+    }
+
+    if (response){
+        snmp_free_pdu(response);
```
dan-du-car, line 172: Why do we need to free pdu? also the OID_len is only initialized to MAX_OID_LEN, and did not realize it is updated at some point?

tsc_client_service/include/snmp_client.h
adev4a: I guess host here is misleading. It would be the server, which can be the traffic signal controller or any snmp enabled device

tsc_client_service/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
+cmake_minimum_required(VERSION 3.10.2)
+project(traffic_signal_controller_service)
+
+
+link_directories("/usr/lib" "/usr/local/lib")
```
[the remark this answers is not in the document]
adev4a: yes, because the net-snmp lib is getting installed in usr/local

tsc_client_service/include/snmp_client.h
adev4a: community is like the authentication required for connecting to the tsc. I guess we can think of it like a domain to access for snmp v2 and v1 where there is no password required. 
Its a necessary parameter required for every snmp call and in our case we almost always be "public".

tsc_client_service/src/main.cpp
```diff
@@ -0,0 +1,33 @@
...
+    // Load config parameters from json
+    std::string ip = streets_service::streets_configuration::get_string_config("ip");
+    int port = streets_service::streets_configuration::get_int_config("port");
+    std::string community = streets_service::streets_configuration::get_string_config("community");
+    int snmp_version = streets_service::streets_configuration::get_int_config("snmp_version");
+    int timeout = streets_service::streets_configuration::get_int_config("timeout");
+
+    TrafficSignalControllerService worker(ip, port, community, snmp_version, timeout);
+    
+    std::string input_oid = "1.3.6.1.4.1.1206.4.2.1.1.2.1.21.2";
```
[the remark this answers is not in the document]
adev4a: OID values are specific to each object we are trying to check. This one was just used for testing and will be removed once we have a pub-sub up. 
If we do decide to have an internal dictionary for the OIDs that we want to use, i can add the description there. For now these variables are a temporary hold for any OIDs we want to test.

tsc_client_service/src/snmp_client.cpp
adev4a: The condition should end if vars->next_variable doesn't exist

tsc_client_service/src/snmp_client.cpp
```diff
@@ -0,0 +1,176 @@
...
+    if(status == STAT_SUCCESS && response->errstat == SNMP_ERR_NOERROR) {
+        
+        SPDLOG_INFO("STAT_SUCCESS, received a response");
+        for(auto vars = response->variables; vars; vars = vars->next_variable){
+            // Get value of variable depending on ASN.1 type
+            // Variable could be a integer, string, bitstring, ojbid, counter : defined here https://github.com/net-snmp/net-snmp/blob/master/include/net-snmp/types.h
+
+            // get Integer value
+            if(vars->type == ASN_INTEGER){
+                long integer_response = *vars->val.integer;
```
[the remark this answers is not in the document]
adev4a: I think these errors would indicate an issue with net-snmp library itself, because there is a field for variable type which should match the value type. 
If that field is not an integer, we don't parse it within the function. I can still add the check, if you want but that's not how we'd expect it to work.
Same for the value not existing, a variable shouldn't be created for null values.

tsc_client_service/CMakeLists.txt
paulbourelly999: @adev4a

tsc_client_service/include/snmp_client.h
```diff
@@ -0,0 +1,74 @@
...
+
+class TrafficSignalControllerService
+{
+    private:
+
+        /*variables to store an snmp session*/
+        // struct that holds information about who we're going to be talking to
+        // We need to declare 2 of these, one to fill info with and second which is 
+        // a pointer returned by the library
+        struct snmp_session session;
```
paulbourelly999, line 17: Why do we have a pointer and regular member variable of the same type. What are these for.

tsc_client_service/src/main.cpp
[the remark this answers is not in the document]
paulbourelly999: If these are just for testing you can probably remove them in your commit. or at least leave inline comments about which portions are just for testing and can be removed.

tsc_client_service/src/snmp_client.cpp
```diff
@@ -0,0 +1,176 @@
...
+                SPDLOG_INFO("Integer value in object: {0}", integer_response);
```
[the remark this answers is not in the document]
adev4a: Makes sense, i just have it this way right now because we're not using the value. I can pass an integer argument by reference to fill in the value here.

tsc_client_service/include/snmp_client.h
adev4a: The comment above describes this a little, the pointer is set as a pointer to the session. Required for a number of the net-snmp functions. This needs to be a class member because its just been set when initializing the class and then used throughout

tsc_client_service/src/main.cpp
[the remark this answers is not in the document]
dan-du-car: Can we use some format like this? and concatenate to a string OID ```
Iso (1) 
.org (3) 
.dod (6) 
.internet (1) 
.private (4)
.transition (868) 
.products (2) 
.chassis (4) 
.card (1)
.slotCps (2)
.cpsSlotSummary (1) 
.cpsModuleTable (1) 
.cpsModuleEntry (1) 
.cpsModuleModel (3) .3562.3
```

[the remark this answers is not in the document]
Comment (dan-du-car) [reply tsc_client_service/src/main.cpp:21]: I think it should be managed by this particular client if this string is const for every set and get request.

File: tsc_client_service/manifest.json (hunk 6 above)
Comment (dan-du-car) [reply tsc_client_service/manifest.json]:
```
Are there any SNMP community string vulnerabilities?
Yes. The fact that the community string is left in plain text is a major security weakness. Furthermore, the frequently used default community string of “public” is the first try that any hacker will make to try to exploit the SNMP system.
```
Refer to: https://www.comparitech.com/net-admin/snmp-community-strings-windows-linux/

Comment (adev4a) [reply tsc_client_service/manifest.json]: Yes, but this is something that has to be set on the server. So we would need to change it on the traffic signal controller

File: tsc_client_service/src/snmp_client.cpp (hunk 14 above)
Comment (adev4a) [reply tsc_client_service/src/snmp_client.cpp:172]: We're freeing the response, because its a class member. Once we finish either get/set, we free it so it doesn't corrupt the other methods

File: tsc_client_service/src/snmp_client.cpp (hunk 13 above)
Comment (adev4a) [reply tsc_client_service/src/snmp_client.cpp:163]: net-snmp generates its own logs. This function is logging the error for the session.

File: tsc_client_service/src/main.cpp (hunk 16 above)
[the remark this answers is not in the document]
Comment (adev4a) [reply tsc_client_service/src/main.cpp:21]: This string isn't constant for every set and get request. Its an id for each object that we want to get or set. 
I'll remove it from here

File: tsc_client_service/src/snmp_client.cpp (hunk 19 above)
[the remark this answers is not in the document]
Comment (dan-du-car) [reply tsc_client_service/src/snmp_client.cpp:85]: Yes, please. You have it.

File: tsc_client_service/src/snmp_client.cpp [hunk 20]
```diff
@@ -0,0 +1,176 @@
...
+    // Check response
+    if(status == STAT_SUCCESS && response->errstat == SNMP_ERR_NOERROR) {
+        
+        SPDLOG_INFO("STAT_SUCCESS, received a response");
+        for(auto vars = response->variables; vars; vars = vars->next_variable){
+            // Get value of variable depending on ASN.1 type
+            // Variable could be a integer, string, bitstring, ojbid, counter : defined here https://github.com/net-snmp/net-snmp/blob/master/include/net-snmp/types.h
+
+            // get Integer value
+            if(vars->type == ASN_INTEGER){
```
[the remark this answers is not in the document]
adev4a: I looked into this and these are macros, not within a namespace. 
https://github.com/net-snmp/net-snmp/blob/b87296ae3ddd1500eb1c8a08b2769307a335a427/include/net-snmp/library/asn1.h#L68

https://github.com/net-snmp/net-snmp/blob/b87296ae3ddd1500eb1c8a08b2769307a335a427/win32/Snmp-winExtDLL.h#L100

tsc_client_service/CMakeLists.txt
adev4a: Not sure why its not showing up here, if you look at the Cmakelists, this change is on there

tsc_client_service/src/snmp_client.cpp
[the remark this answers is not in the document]
paulbourelly999: I agree with @dan-du-car. Also we need to return a value, either by reference as a parameter or from the return statement. Otherwise this is not useable for the next story

tsc_client_service/test/test_traffic_signal_controller_service.cpp
```diff
@@ -0,0 +1,38 @@
...
+    TrafficSignalControllerService worker(dummy_ip, dummy_port);
+    
+    std::string test_oid = "1.3.6.1.4.1.2021.11.9";
+
+    // Expect get call to fail since we're communicating with invalid host
+    EXPECT_FALSE(worker.process_snmp_get_request(test_oid));
+
+}
+
+TEST(traffic_signal_controller_service, test_process_snmp_set_request){
```
paulbourelly999, line 20: Could we add test cases for invalid oid strings. I think this will increase you test coverage

paulbourelly999: Like ones that would make `read_objid` fail/false.

paulbourelly999 (changes requested):
Once the get method returns values and you fix the cmake test executable target, it is good with me. You could add unit test coverage for invalid oid strings like I suggested pretty easily but it isn't necessary.

tsc_client_service/CMakeLists.txt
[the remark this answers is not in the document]
paulbourelly999: You can test this locally but I think these directories are linked by default so this line may not be necessary.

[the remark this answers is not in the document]
adev4a: Hmm i'm fairly certain /usr/local/libs needs to be added to path to link them

dan-du-car (approved):

tsc_client_service/src/main.cpp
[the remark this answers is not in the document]
paulbourelly999: I think this is far more complex than the functionality we need for this UC.  What is the benefit of assembling the OID like this. We only have to support like 6-7 OIDs out of allot define in NTCIP. This seems like a lot of extra scope for little benefit.

PR #150, opened 2022-06-21
140 vehicle scheduler lib documentation

paulbourelly999:
# PR Details
## Description

This is a documentation update for the streets_vehicle_scheduler library. In includes the addition of a README.MD which should include information about how to use this library and how to include this library in other CARMA-Streets services.
## Motivation and Context
Documentation

## How Has This Been Tested?
NA

## Types of changes

## Checklist:

- [ x] My change requires a change to the documentation.


streets_utils/streets_vehicle_scheduler/README.md
```diff
@@ -0,0 +1,35 @@
...
+      "et": 123123,
+      "dt": 123213,
+      "dp": 2,
+      "access": 0
+    }
+  ]  
+}
+```
+### Vehicle scheduler
+The `vehicle_scheduler` class is an abstract class meant to be extended to implement custom scheduling logic (see `all_stop_vehicle_scheduler`). This class has some protected members that store information about intersection geometry and other information commonly necessary for scheduling decisions. To implement custom scheduling business logic simply extend this class and implement the virtual method `schedule_vehicles` which takes in a map of `vehicles` with vehicle string IDs as keys and an empty `intersection_schedule`. Using pass by reference, the implemented method should populate this intersection schedule with schedules for all the relavent vehicles in the intersection.
```
dan-du-car, line 35: For this pass by reference which references are we passing, and are they constant?

streets_utils/streets_vehicle_scheduler/README.md
```diff
@@ -0,0 +1,35 @@
+# Streets Vehicle Scheduler Library
+
+## Introduction
+This library is intended to contain the vehicle scheduling business logic for **CARMA Streets**. This library consists of two main classes: the **intersection_schedule** struct and the **vehicle_scheduler** abstract class.
+
+### Intersection Schedule 
+The intersection schedule is the data object that holds the **vehicle_schedule(s)** for all vehicles in the intersection. Each **vehicle_schedule** contains the **vehicle ID** `v_id`, the current planned **stopping time** `st`, the current planned **entering time** `et`, the current planned **departure time** `dt`, the current planned **departure position** `dp`, and a boolean flag indicating whether the vehicle has **access** to the intersection `access`. Below is an sample JSON
+```
```
dan-du-car, line 8: Can we list the particular fields used by all_stop_intersection and signalized_intersection and categorize them with these two different types of intersections?

dan-du-car: There are some constraints on those fields like timestamp has to be within the timeout period, etc...

paulbourelly999: So no, the timestamp for the schedule is unrelated to the timestamp from the vehicle. I can add information about timeout implemented for the all_stop_intersection if you like.

streets_utils/streets_vehicle_scheduler/README.md
adev4a, line 35: Can we include some sample code for this? Just showing what is written here

streets_utils/streets_vehicle_scheduler/README.md
```diff
@@ -0,0 +1,35 @@
+# Streets Vehicle Scheduler Library
+
+## Introduction
+This library is intended to contain the vehicle scheduling business logic for **CARMA Streets**. This library consists of two main classes: the **intersection_schedule** struct and the **vehicle_scheduler** abstract class.
+
+### Intersection Schedule 
+The intersection schedule is the data object that holds the **vehicle_schedule(s)** for all vehicles in the intersection. Each **vehicle_schedule** contains the **vehicle ID** `v_id`, the current planned **stopping time** `st`, the current planned **entering time** `et`, the current planned **departure time** `dt`, the current planned **departure position** `dp`, and a boolean flag indicating whether the vehicle has **access** to the intersection `access`. Below is an sample JSON
```
adev4a, line 7: Is the sample Json parsed as the intersection schedule data object? If so can we add that to the description?
Also it looks like the intersection_schedule does more than just store the data_object, it returns delay, converts schedule to a csv, json

streets_utils/streets_vehicle_scheduler/README.md
dan-du-car: The timestamp refers to vehicle status and intent metadata field.

paulbourelly999: No so vehicle timestamp and schedule timestamp are completely independent

streets_utils/streets_vehicle_scheduler/README.md
paulbourelly999: Using pass by reference, the implemented method should populate this **intersection_schedule** with schedules for all relevant vehicles in the intersection

streets_utils/streets_vehicle_scheduler/README.md
paulbourelly999: No the sample json is what we produce. In other words on the CARMA-Streets we only handle serialization to JSON, not reading it but I provided descriptions of all the fields I think

streets_utils/streets_vehicle_scheduler/README.md
```diff
@@ -0,0 +1,71 @@
...
+## Introduction
+This library is intended to contain the vehicle scheduling business logic for **CARMA Streets**. This library consists of two main classes: the **intersection_schedule** struct and the **vehicle_scheduler** abstract class.
+
+### Intersection Schedule 
+The intersection schedule is the data object that holds the **vehicle_schedule(s)** for all vehicles in the intersection. Each **vehicle_schedule** contains the **vehicle ID** `v_id`, the current planned **stopping time** `st`, the current planned **entering time** `et`, the current planned **departure time** `dt`, the current planned **departure position** `dp`, and a boolean flag indicating whether the vehicle has **access** to the intersection `access`. Below is an sample JSON for a stopped controlled intersection. 
+**Note** : The JSON serialization will need to be overriden for **signalized** intersection_type for UC 3.
+```
+{
+  "metadata":{
+    "timestamp": 12312, // Timestamp in milliseconds
```
dan-du-car, line 12: Timestamp is the unix time generated by the carma-streets scheduling service scheduler instance to indicate when this schedule is generated. Maybe append this node above.

streets_utils/streets_vehicle_scheduler/README.md
dan-du-car: sorry, confused this with the vehicle status and intent message

streets_utils/streets_vehicle_scheduler/README.md
dan-du-car: There are two references we are passing, the map of vehicles, and the empty intersection schedule, maybe update passing the empty intersection_schedule by reference, and passing the map of vehicle by const reference, so that we know which parameters are updated, which are not.

paulbourelly999: So internally we modify the map but you are right, it does not need to be passed in by reference, we could just pass it in by value, although it may be large. Not sure how I should approach this. The two options I see :
1) Pass vehicle map by value which may be expensive but we don't need a reference for it.
2) Pass vehicle by constant reference and copy to make estimation changes. As part of scheduling we estimate the vehicles location at a common time. To do this we modify the vehicle list with our estimated trajectory information. In order to modify the vehicle list we would need to copy it first which would make this just as expensive as pass by value

@dan-du-car Is passing this vehicle map by value an adequate solution for you? It would require a copy operation but would remove any confusion about whether we use the vehicle map information after at all.

dan-du-car: Would like clarification on the usage of this map of vehicles in this schedule function for documentation purpose.

streets_utils/streets_vehicle_scheduler/README.md
dan-du-car: That does add a bit more information. It does not seem important.

streets_utils/streets_vehicle_scheduler/README.md
paulbourelly999: Ok I can add that to the README. @dan-du-car Thanks!

dan-du-car (approved):

PR #155, opened 2022-06-24
Tsc spat rx

abey-yoseph:
# PR Details
## Description

Add worker class to receive spat message UDP stream from traffic signal controller.

## How Has This Been Tested?

Tested locally with Econolite signal controller.

## Types of changes

## Checklist:
- [x ] I have added tests to cover my changes.
- [x ] All new and existing tests passed.


tsc_client_service/include/spat_receive_worker.h
```diff
@@ -0,0 +1,65 @@
+#pragma once
+#define SPDLOG_ACTIVE_LEVEL SPDLOG_LEVEL_DEBUG
```
paulbourelly999, line 2: This can be removed from all the library headers and just included in the main.cpp for the tsc service.

tsc_client_service/manifest.json
```diff
@@ -1,16 +1,16 @@
 {
     "service_name": "snmp_client",
-    "loglevel": "debug",
+    "loglevel": "info",
```
paulbourelly999, line 3: Lets leave this as default debug.

tsc_client_service/src/spat_receive_worker.cpp
```diff
@@ -0,0 +1,83 @@
...
+    int tscAddrInfo = getaddrinfo(ip_.c_str(), std::to_string(port_).c_str(), &hints, &result);
+    if (tscAddrInfo != 0) {
+        SPDLOG_DEBUG("Failed to get addr info for this streets instance");
+        return;
+    }
+
+    //creating socket
+    int sock = socket(AF_INET, SOCK_DGRAM, IPPROTO_UDP);
+    if (sock == -1) {
+        SPDLOG_DEBUG("Failed to create socket");
```
paulbourelly999, line 37: Lets make this ERROR.

paulbourelly999:
@abey-yoseph Don't worry about making changes to this PR. I will address these comments.

tsc_client_service/src/spat_receive_worker.cpp
```diff
@@ -0,0 +1,83 @@
...
+    if (bind(sock, result->ai_addr, result->ai_addrlen) == -1) {
+        SPDLOG_DEBUG("Failed to bind to socket");
+        return;
+    }
+    
+    //if we reach here, the socket has been created successfully
+    socketCreated = true;
+
+    //reading data from socket
+    while (true)
```
adev4a, line 56: Can we set this to while(socketCreated)? Is there a condition under which this loop should stop?

tsc_client_service/include/spat_receive_worker.h
dan-du-car: What is the purpose of this macro? It might be preferred to define them in a header file?

tsc_client_service/include/spat_receive_worker.h
```diff
@@ -0,0 +1,65 @@
...
+#include <sstream>
+#include <iomanip>
+
+namespace traffic_signal_controller_service
+{
+class SpatWorker 
+{
+    private:
+        /**
+         * @brief ip address of TSC
```
dan-du-car, line 20: Is this TSC a device or service endpoint ?

tsc_client_service/include/spat_receive_worker.h
```diff
@@ -0,0 +1,65 @@
...
+    private:
+        /**
+         * @brief ip address of TSC 
+         */
+        std::string ip_; 
+        
+        /**
+         * @brief ethernet port to receive spat messages on
+         */
+        int port_;
```
dan-du-car, line 27: The same as the above

tsc_client_service/include/spat_receive_worker.h
```diff
@@ -0,0 +1,65 @@
...
+         */
+        std::string ip_; 
+        
+        /**
+         * @brief ethernet port to receive spat messages on
+         */
+        int port_; 
+
+        /**
+        * @brief timeout for udp socket to TSC
```
dan-du-car, line 30: If this TSC is a service, what is timeout unit of measure? Maybe add "service" after TSC?

tsc_client_service/include/spat_receive_worker.h
```diff
@@ -0,0 +1,65 @@
...
+         * @brief boolean to verify whether socket has successfully been created
+         */
+        bool socketCreated = false; 
+
+    public:
+        /**
+         * @brief Construct a new Spat Worker object. This will initialize the member variables to the values assigned in the
+         * manifest json file
+         * 
+         * @param ip The ip address of the TSC
```
dan-du-car, line 44: The same comments as the above

tsc_client_service/manifest.json
```diff
@@ -32,6 +32,27 @@
...
+        {
+            
+            "name": "local_udp_port",
+            "value": 6053,
+            "description": "local port to receive spat messages on from TSC",
+            "type": "INTEGER"
+        },
+        {
+            
+            "name": "socket_timeout",
```
dan-du-car, line 52: unit of measure

tsc_client_service/src/main.cpp
```diff
@@ -10,14 +11,26 @@ int main()
...
     int timeout = streets_service::streets_configuration::get_int_config("timeout");
+    std::string local_ip = streets_service::streets_configuration::get_string_config("local_ip");
+    int local_port = streets_service::streets_configuration::get_int_config("local_udp_port");
+    int socketTimeout = streets_service::streets_configuration::get_int_config("socket_timeout");
 
-    traffic_signal_controller_service::snmp_client worker(ip, port, community, snmp_version, timeout);
+    traffic_signal_controller_service::snmp_client worker(target_ip, target_port, community, snmp_version, timeout);
     
+    //enable spat udp stream on tsc
+    std::string enable_spat_oid = "1.3.6.1.4.1.1206.3.5.2.9.44.1.0";
```
dan-du-car, line 26: Please use structure for each value separated by period if we need to construct different OIDs.

tsc_client_service/src/main.cpp
```diff
@@ -10,14 +11,26 @@ int main()
...
+    std::string request_type = "SET";
+    int64_t enable_spat_value = 2;
```
dan-du-car, line 28: What does this value 2 represent?

tsc_client_service/src/spat_receive_worker.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+SpatWorker::SpatWorker(const std::string& ip, const int& port, const int& socketTimeout) : ip_(ip), port_(port), socketTimeout_(socketTimeout)
+{
+    SPDLOG_DEBUG("Creating Spat Worker");
+}
+
+void SpatWorker::createSocket()
+{
+    struct addrinfo hints;
+    struct addrinfo *result;
```
dan-du-car, line 16: remove struct if not needed.

tsc_client_service/src/spat_receive_worker.cpp
```diff
@@ -0,0 +1,83 @@
...
+    memset(&hints, 0, sizeof(hints));
+    hints.ai_family = AF_UNSPEC;
+    hints.ai_socktype = SOCK_DGRAM;
+    hints.ai_protocol = IPPROTO_UDP;
+
+    //getting address info for streets, 0 indicates success
+    int tscAddrInfo = getaddrinfo(ip_.c_str(), std::to_string(port_).c_str(), &hints, &result);
+    if (tscAddrInfo != 0) {
+        SPDLOG_DEBUG("Failed to get addr info for this streets instance");
+        return;
```
dan-du-car, line 31: maybe just exit(-1);

tsc_client_service/src/spat_receive_worker.cpp
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    //set socket options with timeout from manifest json
+    tv.tv_sec = socketTimeout_;
+    tv.tv_usec = 0;
+    setsockopt(sock, SOL_SOCKET, SO_RCVTIMEO, &tv, sizeof(struct timeval));
+
+    //attempt to bind to socket
+    if (bind(sock, result->ai_addr, result->ai_addrlen) == -1) {
```
dan-du-car, line 47: What does -1 mean?

tsc_client_service/src/spat_receive_worker.cpp
```diff
@@ -0,0 +1,83 @@
...
+            ss << std::hex << std::setfill('0');
+            for (int i = 0; i < bytesReceived; i++)
+            {
+                ss << std::setw(2) << static_cast<unsigned>(spatBuf[i]);
+            }
+            std::string mystr = ss.str();
+            SPDLOG_DEBUG("Buffer contains: {0}", mystr);
+
+        }       
+    }
```
dan-du-car, line 76: Maybe sleep within the while loop to check whether receives bytes at a particular rate?

tsc_client_service/src/spat_receive_worker.cpp
```diff
@@ -0,0 +1,83 @@
...
+                ss << std::setw(2) << static_cast<unsigned>(spatBuf[i]);
+            }
+            std::string mystr = ss.str();
+            SPDLOG_DEBUG("Buffer contains: {0}", mystr);
+
+        }       
+    }    
+}
+
+bool SpatWorker::getSocketStatus() const {
```
dan-du-car, line 79: Not sure if this is really necessary

tsc_client_service/src/spat_receive_worker.cpp
```diff
@@ -0,0 +1,83 @@
...
+    //if we reach here, the socket has been created successfully
+    socketCreated = true;
+
+    //reading data from socket
+    while (true)
+    {
+        std::vector<char> spatBuf(maxDataSize);
+        ssize_t bytesReceived = recv(sock, spatBuf.data(), spatBuf.size(), 0);
+
+        if (bytesReceived != -1)
```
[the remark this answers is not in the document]
abey-yoseph: If it does not receive anything over the socket for some reason. I don't think we would exit if it doesn't receive anything.

tsc_client_service/src/spat_receive_worker.cpp
abey-yoseph: No I think we would want the loop to run forever.

tsc_client_service/include/spat_receive_worker.h
abey-yoseph: Sorry, I think that is actually a typo. It should be the local IP address of carma streets. This IP and port will be used to receive spat data on the UDP socket.

tsc_client_service/include/spat_receive_worker.h
abey-yoseph: Timeout is measured in seconds

tsc_client_service/manifest.json
abey-yoseph: seconds

tsc_client_service/src/main.cpp
abey-yoseph: 2 enables the udp spat stream on the traffic signal controller.

tsc_client_service/src/main.cpp
```diff
@@ -10,14 +11,26 @@ int main()
...
+    worker.process_snmp_request(enable_spat_oid, request_type, enable_spat_value);
+
+    //instantiate spat receive worker
+    traffic_signal_controller_service::SpatWorker spatWorker(local_ip, local_port, socketTimeout);
+    spatWorker.createSocket();
```
dan-du-car, line 33: This needs to be in a separate thread if this service does data receiving and message publishing operations.

tsc_client_service/include/spat_receive_worker.h
```diff
@@ -0,0 +1,65 @@
...
+         * @param socketTimeout Timeout, in seconds, for udp socket to TSC
+         */
+        SpatWorker(const std::string& ip, const int& port, const int& socketTimeout);
+
+        /**
+         * @brief Create a socket to the ip and port member variables. If it is successfully created, the received content will
+         * be printed to the spd log.
+         * 
+         */
+        void createSocket();
```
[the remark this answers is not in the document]
dan-du-car: I see. The main is calling this createSocket function. So the main() will be hanging and wait for incoming message. But this service also does the send NTCIP command to the traffic signal controller. How do we manage that?

tsc_client_service/manifest.json
dan-du-car: Please update the comments or description to note them. Thanks.

tsc_client_service/src/main.cpp
dan-du-car: Is there documentation about different values? Can we add the reference here in inline comments?

tsc_client_service/include/spat_receive_worker.h
paulbourelly999: In order to use the logging macros it has to be added. Otherwise we can switch below debug.  Please refer to the logging macros section of the rapidjson documentation (https://spdlog.docsforge.com/v1.x/1.quickstart/#create-a-logger-with-multiple-sinks-each-sink-with-its-own-formatting-and-log-level)

tsc_client_service/include/spat_receive_worker.h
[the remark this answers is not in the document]
paulbourelly999: Create socket does not call any NTCIP commands. main calls NTCIP set on the oid to enable spat broadcast.  This worker simply reads UDP packets sent here. Eventually it will also translate and forward those to kafka as SPAT JSON using the streets_signal_phase_and_timing_library. @dan-du-car Please let me know if there is something here you want addressed.

tsc_client_service/src/main.cpp
paulbourelly999: I do not thing this is necessary. Our entire UC 3 work is limited to 5 or 6 OIDs. We could store these as constants. @dan-du-car Please explain how you would structure this differently and the intended benefit more.

tsc_client_service/src/spat_receive_worker.cpp
paulbourelly999: I don't think sleep is necessary but running test cases with the timeout set to 1 and the host set to 127.0.0.1 confirms that timeout does not function the way we expect. I thought the timeout would be triggered if we don't receive any data but currently it does not look like that.

dan-du-car: If there is no data received, the while loop is infinitely running. Do we have a better way to handle this?

tsc_client_service/src/main.cpp
dan-du-car: https://github.com/usdot-fhwa-stol/carma-streets/pull/147#discussion_r901843427

tsc_client_service/include/spat_receive_worker.h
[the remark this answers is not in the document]
dan-du-car: NTCIP commands referring to GET and SET with SNMP lib? What is this socket about? Isn't it receiving data from traffic signal controller, the data is mapped into the NTCIP object? Then converted to SPAT? So forwarding NTCIP SET commands operation should be in one thread. I am not sure what the purpose of GET if we have socket to receive NTCIP object. Another separate thread to be dedicated to receiving NTCIP object.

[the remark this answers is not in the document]
paulbourelly999: There are two functionalities here Dan. The first is listening to SPaT NTCIP packets over UDP and the other is making NTCIP calls. We have stories for READing the SPaT NTCIP packets in the next sprint and stories for making the GET calls that we need to maintain state of the traffic signal controller in the next sprint as well. I believe we can create these threads in those stories. I don't think threading needs to be address here since there is still no functionality the service is executing yet. Currently we just have the snmp client which allows us to make SET and GET calls and the spatworker which standups up a UDP listener and listens for SPAT.

tsc_client_service/src/main.cpp
paulbourelly999: I think this is far more complex than the functionality we need for this UC. What is the benefit of assembling the OID like this. We only have to support like 6-7 OIDs out of allot define in NTCIP. This seems like a lot of extra scope for little benefit.

tsc_client_service/src/main.cpp
paulbourelly999: I don't think we need OIDs referenced inline. I think we can just include any documentation in the README.

tsc_client_service/src/spat_receive_worker.cpp
paulbourelly999: I am looking into how to handle this for a udp socket

paulbourelly999: Fixed this, and added a unit test for it. receive  method is blocking and waits for timeout. Added else that gets executes on timeout.

tsc_client_service/src/spat_receive_worker.cpp
paulbourelly999: Do we want the tsc_service to restart if it can't receive SPaT data or just retry a connection. Also calling exit inside the worker makes it a little more difficult to unit test

tsc_client_service/src/spat_receive_worker.cpp
```diff
@@ -0,0 +1,83 @@
...
+        if (bind(sock, result->ai_addr, result->ai_addrlen) == -1) {
+            SPDLOG_ERROR("Failed to bind to socket");
+            return;
+        }
+        
+        //if we reach here, the socket has been created successfully
+        socketCreated_ = true;
+
+        //reading data from socket
+        while (true)
```
dan-du-car, line 56: Please refer to this code https://github.com/usdot-fhwa-OPS/V2X-Hub/blob/a93ad7563e8f75ace2729efbe40de2767c90e36d/src/v2i-hub/SpatPlugin/src/signalController.cpp#L106 
I think it makes more sense to always try to check the socket connection and sleep for some seconds at the outer while loop. Then at another inner while loop (with connection succeed condition),  it receives the data from traffic signal controller. 
Also socketCreated_ variable is not used, and maybe should be removed.

paulbourelly999: https://github.com/usdot-fhwa-OPS/V2X-Hub/blob/a93ad7563e8f75ace2729efbe40de2767c90e36d/src/v2i-hub/SpatPlugin/src/signalController.cpp#L146 . If you look at the code the ethernet connection bool is just managed by if bytes received is 0 or -1.  We are doing the same thing here just you don't need the bool to break out of the while loop. You can just break out of the while loop when the if condition is false. I can add a condition for 0 which Abey didn't seem to have. Not sure what that condition represents but otherwise this code is the same.

paulbourelly999: Will remove socket created.

tsc_client_service/src/main.cpp
paulbourelly999: Replaced this with a const string , see ntcip_oid.h

tsc_client_service/src/spat_receive_worker.cpp
```diff
@@ -0,0 +1,83 @@
...
+    SPDLOG_DEBUG("Creating Spat Worker");
+}
+
+void SpatWorker::createSocket()
+{
+    struct addrinfo hints;
+    struct addrinfo *result;
+    struct timeval tv;
+
+    int maxDataSize = 1000;
```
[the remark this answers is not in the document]
adev4a: Can we add this as a comment @paulbourelly999

tsc_client_service/src/spat_worker.cpp
```diff
@@ -0,0 +1,78 @@
...
+        SPDLOG_DEBUG("Creating Spat Worker");
+    }
+
+    void spat_worker::listen_udp_spat() const
+    {
+        addrinfo hints;
+        addrinfo *result;
+        timeval tv;
+
+        int max_data_size = 1000;
```
dan-du-car, line 19: Is this max 1000 number of bytes?

tsc_client_service/src/spat_worker.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+        //reading data from socket
+        while (true)
+        {
+            std::vector<char> spat_buf(max_data_size);
+            ssize_t bytes_received = recv(sock, spat_buf.data(), spat_buf.size(), 0);
+
+            if (bytes_received > 0)
+            {
+                spat_buf.resize(bytes_received);
```
dan-du-car, line 58: If the received bytes size is greater than the max_data_size, it will resize the container and add the extra bytes? Maybe make a note here with inline comments?
Do we need to specify the max size of the spat buf vector?

tsc_client_service/src/spat_worker.cpp
```diff
@@ -0,0 +1,78 @@
...
+                SPDLOG_DEBUG("Num bytes received: {0}", bytes_received);
+
+                std::stringstream ss;
```
dan-du-car, line 61: Would recommend to have  a function std::string convertBytesToString(std::vector<char> bytes_received) to wrap line 62 - 68, and have  a unit test for this.

dan-du-car (approved):

tsc_client_service/include/spat_receive_worker.h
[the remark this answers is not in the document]
dan-du-car: Sounds good.

adev4a (approved):

PR #163, opened 2022-07-08
Extract phase id from signal group

adev4a:
# PR Details
This PR adds a mapping between phase id and signal group id(as defined in in SPAT messages)
## Types of changes

## Checklist:

### DIFF
--- a/tsc_client_service/include/snmp_client.h
+++ b/tsc_client_service/include/snmp_client.h
@@ -39,6 +39,12 @@ class snmp_client
         /*Time after which the the snmp request times out*/
         int timeout_ = 10000;
 
+        /* Map between phase numbers(key) and signal group ids(value) for all vehicle phases in the Traffic Signal Controller*/
+        std::unordered_map<int,int> phase_num_map_;
+
+        /* Map between signal group ids(key) and phase numbers(value) for all vehicle phases in the Traffic Signal Controller*/
+        std::unordered_map<int,int> signal_group_map_;
+
     public:
         /** @brief Constructor for Traffic Signal Controller Service client.
          *  Uses the arguments provided to establish an snmp connection
@@ -67,6 +73,11 @@ class snmp_client
          *  @param request_type The request type for which the error is being logged (GET/SET).
          *  @param response The snmp_pdu struct */
         void log_error(const int& status, const std::string& request_type, snmp_pdu *response);
+
+        /** @brief Constructs a map between phase number and signal group ids
+        * According to the NTCIP 1202 v03 documentation signal group ids in the SPAT message are the channel numbers in the TSC
+        * **/
+        void get_phasenum_signalgroup_map();
 
         /** @brief Destructor for client. Closes the snmp session**/
         ~snmp_client(){
--- a/tsc_client_service/include/snmp_client.h
+++ b/tsc_client_service/include/snmp_client.h
@@ -75,7 +75,7 @@ class snmp_client
          *  @param status The integer value corresponding to net-snmp defined errors. macros considered are STAT_SUCCESS(0) and STAT_TIMEOUT(2)
          *  @param request_type The request type for which the error is being logged (GET/SET).
          *  @param response The snmp_pdu struct */
-        void log_error(const int& status, const int& request_type, snmp_pdu *response);
+        void log_error(const int& status, const int& request_type, snmp_pdu *response) const;
 
         /** @brief Destructor for client. Closes the snmp session**/
         ~snmp_client(){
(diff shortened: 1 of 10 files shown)


tsc_client_service/src/snmp_client.cpp
```diff
@@ -163,4 +171,65 @@ void snmp_client::log_error(const int& status, const std::string& request_type,
     
 }
 
+int64_t& snmp_client::get_max_channels(){
```
paulbourelly999: Does this need to be &int64_t? Could this not be a regular int? Is the max channels such a large number that we need to start it in a int64_t?

tsc_client_service/src/snmp_client.cpp
```diff
@@ -163,4 +171,65 @@ void snmp_client::log_error(const int& status, const std::string& request_type,
...
+        SPDLOG_ERROR("Failed to get max channels : {0} ", e.what());
+    }
+
+    return max_channels_in_tsc;
+}
+
+std::vector<int>& snmp_client::get_vehicle_phase_channels(int64_t& max_channels){
+    std::vector<int> vehicle_phase_channels;
+    // Loop through channel control types and add channels with vehicle phase to list
+    int64_t vehicle_control_type  = 0;
```
paulbourelly999, line 189: Just for clarity maybe we should mention the phase types we are ignoring ( i.e pedestrian ... etc)

tsc_client_service/src/snmp_client.cpp
```diff
@@ -163,4 +171,61 @@ void snmp_client::log_error(const int& status, const std::string& request_type,
...
+        {
+            vehicle_phase_channels.push_back(channel_num);
+        }
+
+    }
+
+    return vehicle_phase_channels;
+}
+
+void snmp_client::map_phase_and_signalgroup(std::vector<int>& vehicle_phase_channels)
```
paulbourelly999, line 210: vehicle_phase_channels param can be constant since we are not changing it  right?

tsc_client_service/src/snmp_client.cpp
```diff
@@ -163,4 +171,61 @@ void snmp_client::log_error(const int& status, const std::string& request_type,
...
+    // Get phases associated with vehicle phase channels
+    std::string request_type = "GET";
+    for(int channel : vehicle_phase_channels)
+    {
+        int64_t phase_num = 0;
+        std::string control_source_parameter_oid = ntcip_oids::CHANNEL_CONTROL_SOURCE_PARAMETER + "." + std::to_string(channel);
+        process_snmp_request(control_source_parameter_oid, request_type, phase_num);
+
+        // According to NTCIP 1202 v03 returned value of 0 here would mean a phase is not associated with the channel
+        if(phase_num != 0)
```
paulbourelly999, line 223: Should we log some warning incase this happens. This would mean that we got a channel we think is associated with a vehicle phase but when we obtain the channel source it resolves to zero so we don't end up mapping it.

tsc_client_service/include/snmp_client.h
```diff
@@ -39,6 +39,12 @@ class snmp_client
         /*Time after which the the snmp request times out*/
         int timeout_ = 10000;
 
+        /* Map between phase numbers(key) and signal group ids(value) for all vehicle phases in the Traffic Signal Controller*/
+        std::unordered_map<int,int> phase_num_map_;
```
paulbourelly999, line 43: Maybe we should move these methods and information into a separate class. There are two potential benefits to doing so. First, we can mock out all the SNMP client methods to allow us to test our logic given some expected returns. Secondly, this way we separate the communication logic (snmp) from our NTCIP processing logic. 

In my opinion, if we had a separate class like `traffic_signal_state` that had an `initialize` method that would call these methods to populate these maps, this implementation would be a little cleaner.

paulbourelly999 (changes requested):

tsc_client_service/src/snmp_client.cpp
paulbourelly999: OK maybe even a DEBUG level log statement so we can ensure we understand what is happening for vehicle phases not represented in the MAP

tsc_client_service/include/snmp_client.h
```diff
@@ -7,6 +7,14 @@
 
 namespace traffic_signal_controller_service
 {
+
+enum request_type
```
paulbourelly999, line 11: Sonar scanner recommends you make this `enum class` instead of `enum`. (https://stackoverflow.com/questions/18335861/why-is-enum-class-preferred-over-plain-enum). The main idea is that `enum` will implicitly convert to other types like int and other enums where as `enum class` will not.  This is preferred for readability and safety so we can more easily detect when we are making incorrect comparisons. `enum class` can still be explicitly cast to int of course.

paulbourelly999 (approved):

Issue #162, opened 2022-07-08
Implement Scheduling Service to use configurable parameter to select intersection type


paulbourelly999:
### Types of Issue

- [ ] Anomaly report (something appears to not work correctly)
- [ x] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

Currently in the scheduling service we have a configuration parameter for intersection type but we are not reading it. Additionally the all_stop_scheduling_worker and  all_stop_scheduling_service do not allow for assigning any status_and_intent_processor or vehicle_scheduler but instead expect an all_stop specific status_and_intent_processor and scheduler. The scheduling service needs to be able to configurably use, different scheduling and vehicle processing business logic
### Carma streets version where this issue was discovered

Latest

### Expected behavior
Configuration Parameter controls which type of scheduling/vehicle update processing is used by the scheduling service
### Actual behavior

Configuration Parameter is not read.
### Steps to reproduce the actual behavior

1. Change configuration parameter
2. Nothing happens

Issue #161, opened 2022-07-07
Fix intersection_client issue in scheduling_service


Saeid-Solei:
### Types of Issue

- [x] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

After initializing kafka, intersection_model, and scheduling_service, and sending a sample status and intent message, scheduling_service throws an exception: "Map of vehicles to be scheduled is empty but list of EVs to be scheduled is not!"

Based on the scheduling service log file, the intersection_client is not receiving the correct intersection information:

"[2022-07-07 16:39:46.731] [info] [intersection_client.cpp:37] request_intersection_info succeed!
**[2022-07-07 16:39:46.731] [debug] [intersection_client.cpp:38] intersection name: 
[2022-07-07 16:39:46.731] [debug] [intersection_client.cpp:39] intersection id: 1806426938**
[2022-07-07 16:39:46.731] [info] [intersection_client.cpp:56] Done with request_intersection_info
[2022-07-07 16:39:46.731] [info] [intersection_client.cpp:15] Intersection information is updated successfuly! "

[scheduling_service_2022-07-07.txt](https://github.com/usdot-fhwa-stol/carma-streets/files/9067298/scheduling_service_2022-07-07.txt)

### Carma streets version where this issue was discovered

3.11.0

### Expected behavior
scheduling_service is able to schedule the vehicles.
intersection_client in the scheduling_service is able to get the intersection information.

### Actual behavior

scheduling service throws and exception => "Map of vehicles to be scheduled is empty but list of EVs to be scheduled is not!"
intersection_client in the scheduling_service is not able to get the intersection information properly.

### Steps to reproduce the actual behavior

1. Initialize kafka and intersection_model services
2. Start scheduling_service
3. Send a sample status and intent message (run simulate_vehicle_status_and_intent_pub.py file located in the script folder to send a set of sample status and intent messages)

PR #164, opened 2022-07-08
Issue-162: Bug fixes

paulbourelly999:
+ Add trace SPDLOG definition to each compilation unit
+ Use polymorphism for pointers in scheduling_service
+ Fix vehicle_list library log statements
+ Fix scheduling Service dockerfile

# PR Details
## Description
This PR fixes several bugs/implementation detailed for the scheduling_service and its dependencies. The largest change is that that scheduling_service and the scheduling_worker now utilize polymorphism by using abstract classes as shared_ptr types for the vehicle_scheduler. Also updates intersection_client `connect` lamda expressions to fix core dump on scheduling.

## Related Issue
#162 
#161 

## Motivation and Context
Fix bugs 

## How Has This Been Tested?
Run scheduling_service with intersection client and simulate_status_and_intent.py

1) docker compose up intersection_model
2) build scheduling_service
3) run scheduling_service
4) run scripts/simulate_vehicle_status_and_intent.py
## Types of changes

- [ x ] Defect fix (non-breaking change that fixes an issue)

## Checklist:


scheduling_service/include/scheduling_service.h
```diff
@@ -70,34 +71,37 @@ namespace scheduling_service{
...
-        void schedule_veh(std::shared_ptr<kafka_clients::kafka_producer_worker> _producer_worker, std::shared_ptr<all_stop_scheduling_worker> _scheduling_worker, std::shared_ptr<streets_vehicles::vehicle_list> _vehicle_list_ptr, std::shared_ptr<streets_vehicle_scheduler::all_stop_vehicle_scheduler> _scheduler_ptr) const;
+        void schedule_veh() const;
 
 		/**
 		 * @brief Method to configure spdlog::logger for logging scheduling metrics into daily rotating csv file.
 		 */
 		void configure_csv_logger() const;
 
+
+        void set_scheluding_worker(std::shared_ptr<scheduling_worker> sched_worker);
```
Saeid-Solei, line 97: Are these used anywhere?

Saeid-Solei (approved):

scheduling_service/src/intersection_client.cpp
```diff
@@ -41,12 +41,17 @@ namespace scheduling_service {
...
-
+            loop.quit(); 
+		});
+		// Lambda expression slot for get intersection information error signal ( See Qt Signal/Slot documentation)
         connect(&apiInstance, &OpenAPI::OAIDefaultApi::getIntersectionInfoSignalE, 
-			[&error_str, &loop]{ 
-			SPDLOG_ERROR("Error happened while issuing intersection model GET information request : {0}",  error_str.toStdString());
-			loop.quit(); });
+			[&intersection_info_valid, &loop, this]
+				( [[maybe_unused]] const OpenAPI::OAIIntersection_info &int_info, QNetworkReply::NetworkError error, QString error_str ){
```
dan-du-car, line 49: What does this  unused do [[maybe_unused]]?

scheduling_service/src/scheduling_service.cpp
```diff
@@ -162,15 +171,15 @@ namespace scheduling_service{
...
 
 				std::string msg_to_send = int_schedule.toJson();
 
 				SPDLOG_DEBUG("schedule plan: {0}", msg_to_send);
 
 				/* produce the scheduling plan to kafka */
-				_producer_worker->send(msg_to_send);
+				producer_worker->send(msg_to_send);
 
 				last_schedule_timestamp = int_schedule.timestamp;
```
dan-du-car, line 184: Is this if (_scheduling_worker->start_next_schedule(last_schedule_timestamp, scheduling_delta)) statement needed for checking the last_schedule_timestamp?

scheduling_service/include/scheduling_service.h
```diff
@@ -70,34 +71,70 @@ namespace scheduling_service{
...
-        void schedule_veh(std::shared_ptr<kafka_clients::kafka_producer_worker> _producer_worker, std::shared_ptr<all_stop_scheduling_worker> _scheduling_worker, std::shared_ptr<streets_vehicles::vehicle_list> _vehicle_list_ptr, std::shared_ptr<streets_vehicle_scheduler::all_stop_vehicle_scheduler> _scheduler_ptr) const;
+        void schedule_veh() const;
 
 		/**
 		 * @brief Method to configure spdlog::logger for logging scheduling metrics into daily rotating csv file.
 		 */
 		void configure_csv_logger() const;
 
+        /**
+         * @brief Set the scheduling worker object for unit testing
```
dan-du-car, line 97: Can we use the initialize function for the unit test to make sure it defines those pointers? These 
 set functions are not delivering any business values, and only for unit testing purposes. Do we have better way to unit test this without adding additional functions?

scheduling_service/src/intersection_client.cpp
```diff
@@ -15,8 +15,8 @@ namespace scheduling_service {
 				SPDLOG_INFO("Intersection information is updated successfuly! ");
 				return true;
 			}
-			// usleep takes micro seconds
-			sleep(sleep_secs);
+			// sleep takes seconds
+			std::this_thread::sleep_for(std::chrono::seconds(sleep_secs));
```
dan-du-car, line 19: It seems both sleep do the same thing.

streets_utils/streets_vehicle_list/src/message_processors/all_stop_status_intent_processor.cpp
```diff
@@ -265,7 +265,7 @@ namespace streets_vehicles {
 
 	bool all_stop_status_intent_processor::is_vehicle_stopped(const vehicle &vehicle) const {
 		SPDLOG_DEBUG("Stopping condition spd: {0}, dst: {1}; Vehicle condition spd: {2}, dst: {3} .", 
-			stopping_speed, stopping_distance, vehicle._cur_speed, vehicle._cur_distance);
+			_stopping_speed, _stopping_distance, vehicle._cur_speed, vehicle._cur_distance);
```
dan-du-car, line 268: Can we pass in these two _stopping_speed, _stopping_distance variables instead of instance member? It seems they are used only by this is_vehicle_stopped() function in the processor.

scheduling_service/src/intersection_client.cpp
paulbourelly999: It is to prevent a code-smell. Sonar scanner will create a code-smell for any unused parameters in a method signature. but for the Qt REST client we have to implement the error signal connect with this parameter. The unused tells the compiler, anyone else using this code, and sonar scanner that this parameter maybe unused

scheduling_service/src/scheduling_service.cpp
paulbourelly999: I think technically probably not since we already sleep for the interval time but I think it is a say way of ensuring the correct frequency.

scheduling_service/include/scheduling_service.h
paulbourelly999: The init functionality initializes these pointers based on configuration values in the manifest json. This allows us to initialize them however we want. Initially I was a little hesitant to write set methods that are only used in unit testing but I think the value that offer for unit testing I think makes them work while. What do you think?

Ideally I think we should write the classes so that initialization is configurable using public methods/fields, but in this case I think this is an appropriate work around.

scheduling_service/src/intersection_client.cpp
paulbourelly999: but sleep() is deprecated. It results in a code smell

scheduling_service/src/scheduling_service.cpp
dan-du-car: If this last_schedule_timestamp  is always the latest timestamp now, the start_next_schedule() function seems to sleep for additional 1 second.  For each iterator, it might take less than a second. Let us say 900 milliseconds. The sleep will wait for additional 100 milliseconds. The last_schedule_timestamp   is 100 milliseconds behind?  start_next_schedule() will compare the now() with the last_schedule_timestamp  to make sure the difference is 1000 milliseconds. So until the next iteration, it waits 900 + 100 + 900 milliseconds?

streets_utils/streets_vehicle_list/src/message_processors/all_stop_status_intent_processor.cpp
dan-du-car: I meant using method parameter.

scheduling_service/test/test_all_stop_scheduling_worker.cpp
```diff
@@ -80,13 +81,29 @@ TEST_F(all_stop_scheduling_worker_test, start_next_schedule_true)
...
+    u_int64_t last_schedule_timestamp = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch()).count();
+
+    last_schedule_timestamp -= 2000;
+
+    ASSERT_TRUE( sched_worker -> start_next_schedule(last_schedule_timestamp, scheduling_delta) );
 
 }
 
 
 TEST_F(all_stop_scheduling_worker_test, schedule_vehicles)
```
dan-du-car, line 101: Maybe update test case target to scheduling worker as we renamed the worker class.

scheduling_service/include/scheduling_service.h
paulbourelly999: I can do that as well.

scheduling_service/test/test_all_stop_scheduling_worker.cpp
paulbourelly999: OK the reason I kept this name is because the test case setup creates an all stop scheduler. For signalized we could create another Test case

dan-du-car (approved):

Issue #159, opened 2022-07-06
Implement UC3 Scheduling Logic in CS Scheduling Library


Saeid-Solei:
### Types of Issue

- [ ] Anomaly report (something appears to not work correctly)
- [x] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

Add UC 3 scheduling logic to the existing streets_vehicle_scheduler library

### Carma streets version where this issue was discovered

3.11.0

### Expected behavior
NA

### Actual behavior

NA

PR #167, opened 2022-07-22
159 Implement UC3 Scheduling Logic in CS Scheduling Library

Saeid-Solei:
# PR Details
## Description

Add UC 3 scheduling logic to the existing streets_vehicle_scheduler library

## Related Issue

#159 

## Types of changes

## Checklist:


streets_utils/streets_vehicle_scheduler/src/signalized_vehicle_scheduler.cpp
```diff
@@ -0,0 +1,421 @@
...
+    //         phase_end_time += tbd_duration;
+    //     }
+    // }
+
+
+    void signalized_vehicle_scheduler::estimate_et(const streets_vehicles::vehicle &veh, const signalized_vehicle_schedule &preceding_veh, signalized_vehicle_schedule &sched, const uint64_t phase_start_time, const uint64_t phase_end_time, const uint64_t schedule_timestamp) const {
+
+        // Get link lanelet information for ev
+        OpenAPI::OAILanelet_info link_lane = get_link_lanelet_info( veh );
+        SPDLOG_INFO( "Link lanelet for vehicle {0} is {1}.", veh._id, link_lane.getId());
```
paulbourelly999, line 221: We do not want any INFO level log statements that will print for every scheduling iteration. To have them INFO level locally for debugging is fine but make sure you remove log statements like this that would unnecessarily pollute the log files.

scheduling_service/manifest.json
```diff
@@ -20,6 +20,12 @@
             "description":"Kafka topic to which the scheduling service will consume status and intent updates.",
             "type": "STRING"
         },
+        {
+            "name": "spat_topic",
+            "value": "modified spat",
```
paulbourelly999, line 25: not spaces in topic name. Please use and underscore instead

scheduling_service/manifest.json
```diff
@@ -103,6 +109,18 @@
...
+        {
+            "name": "initial_green_buffer",
+            "value": 2000,
+            "description": "The configurable time interval at the beginning of a green phase that is considered for estimating vehicles' entering times.",
+            "type": "INTEGER"
+        },
+        {
+            "name": "final_green_buffer",
+            "value": 2000,
+            "description": "The configurable time interval at the end of a green phase that is considered for estimating vehicles' entering times.",
```
paulbourelly999, line 122: Please include units

paulbourelly999: Also yellow and red clearance intervals seem to address this type of buffer already. Could you please explain why we need this additional buffer.

streets_utils/streets_vehicle_scheduler/include/all_stop_intersection_schedule.h
```diff
@@ -0,0 +1,117 @@
...
+
+#include <spdlog/spdlog.h>
+#include <rapidjson/rapidjson.h>
+#include <rapidjson/writer.h>
+#include <rapidjson/stringbuffer.h>
+#include "intersection_schedule.h"
+
+namespace streets_vehicle_scheduler {
+    /**
+     * @brief Object to represent schedule for single vehicle.
```
paulbourelly999, line 11: add all_stop schedule for single vehicle

streets_utils/streets_vehicle_scheduler/include/signalized_vehicle_scheduler.h
```diff
@@ -0,0 +1,235 @@
...
+
+            /**
+             * @brief Estimate the intersection departure times (dt's) for all currently Departing Vehicle (DVs) based on kinematic vehicle 
+             * information and intersection geometry. Method assumes empty intersection_schedule is passed in. Method will add vehicle_schedule(s) 
+             * for each DV in the vector. 
+             * 
+             * @param dvs vector of unscheduled DVs.
+             * @param schedule signalized_intersection_schedule to add DV scheduling information to.
+             */
+            void schedule_dvs( std::list<streets_vehicles::vehicle> &dvs, std::shared_ptr<signalized_intersection_schedule> &schedule ) const;
```
paulbourelly999, line 47: Do you modify the list of dvs at all in this method or can this be constant

scheduling_service/manifest.json
Saeid-Solei: We need the initial green buffer since the vehicle aims to enter the intersection box with max speed. Without the initial green buffer, a small inaccuracy in the entering time estimation might cause the vehicle to enter the intersection box during a red phase. 

The final green buffer is not required technically. But in Use Case 2 testing, CARMA Platform has a check for the estimated entering time. If the estimated entering time is very close to the end of the green phase, CARMA Platform considers the beginning of the next green phase as the vehicle's entering time. This check is added to account for the inaccuracy in the entering time estimation. Since the implemented algorithms in UC 2 worked well, I though it is a good idea to have this configurable parameter here as well.

streets_utils/streets_vehicle_scheduler/include/signalized_vehicle_scheduler.h
```diff
@@ -0,0 +1,235 @@
...
+            /**
+             * @brief Schedule all Entering Vehicles (EVs). Method first organizes all EVs into lists based on their entry
+             * lane. These list are ordered in ascending order with the vehicle closest to the intersection in a given lane 
+             * being the first vehicle in that entry lanes vehicle list. Then, for each entry lane, an entering time (ET)
+             * for each EV in the entry lane is calculated sequentially starting from the first vehicle until in the list.
+             * ET is estimated based on the EV's earliest entering time (EET), its preceding vehicle's estimated ET, and 
+             * the modifed spat.
```
paulbourelly999, line 54: Are there any assumptions on the intersection_schedule passed in here. The assumption is that all dvs have been scheduled or is that not an assumption. Can these scheduling methods happen completely independently.

streets_utils/streets_vehicle_scheduler/include/signalized_vehicle_scheduler.h
Saeid-Solei: No I do not modify it. I will make is constant.

streets_utils/streets_vehicle_scheduler/include/signalized_vehicle_scheduler.h
```diff
@@ -0,0 +1,235 @@
...
+             * 
+             * @param veh vehicle for which to calculate distance.
+             * @param max_speed speed limit of the vehicle's entry lane.
+             * @param departure_speed departure speed = speed limit of the vehicle's link lane.
+             * @return double distance in meters
+             */
+            double calculate_distance_accel_and_decel( const streets_vehicles::vehicle &veh, const double max_speed, const double departure_speed ) const;
+            /**
+             * @brief Calculate the distance required to speed up (slow down) to departure speed with maximum acceleration (maximum deceleration). 
+             * If the departure speed is grater than the vehicle's current speed, the vehicle has to speed up with maximum acceleration. Otherwise,
```
paulbourelly999, line 92: Correct `grater` to `greater`

streets_utils/streets_vehicle_scheduler/include/signalized_vehicle_scheduler.h
Saeid-Solei: Similar to all_stop_vehicle_scheduler, the assumption here is that all DVs are scheduled first. Now in this use case, DVs do not need any schedule. but the scheduler needs to estimated the DT of DVs first in order to estimated the first EV from an entry lane.

streets_utils/streets_vehicle_scheduler/src/signalized_vehicle_scheduler.cpp
```diff
@@ -0,0 +1,386 @@
+#include "signalized_vehicle_scheduler.h"
+
+namespace streets_vehicle_scheduler {
+
+    void signalized_vehicle_scheduler::schedule_vehicles( std::unordered_map<std::string, streets_vehicles::vehicle> &vehicles, std::shared_ptr<intersection_schedule> &i_sched) {
+        
+        auto schedule = std::dynamic_pointer_cast<signalized_intersection_schedule> (i_sched);
```
paulbourelly999, line 7: We make want to catch exceptions thrown by this and throw our own scheduling_exception

streets_utils/streets_vehicle_scheduler/src/signalized_vehicle_scheduler.cpp
```diff
@@ -0,0 +1,386 @@
...
+        
+        if ( vehicles.empty() ) {
+            SPDLOG_INFO("No vehicles to schedule.");
```
paulbourelly999, line 10: This log statement should be debug

streets_utils/streets_vehicle_scheduler/src/signalized_vehicle_scheduler.cpp
```diff
@@ -0,0 +1,386 @@
...
+
+
+    signal_phase_and_timing::movement_state signalized_vehicle_scheduler::find_movement_state_for_lane(const OpenAPI::OAILanelet_info &entry_lane_info) const {
+
+        signal_phase_and_timing::movement_state move_state; 
+        if (spat_ptr) {
+            for (const auto& ms : spat_ptr->intersections.front().states){
+                if (ms.signal_group == entry_lane_info.getSignalGroupId()) {
+                    move_state = ms;
+                    break;
```
paulbourelly999, line 148: replace this with `return ms`. Return will break the loop.

streets_utils/streets_vehicle_scheduler/src/signalized_vehicle_scheduler.cpp
```diff
@@ -0,0 +1,386 @@
...
+                }
+            }
+            return move_state;
```
paulbourelly999, line 151: If no movement state is found, you will return an empty one. It might be better to just throw and exception in this case.

streets_utils/streets_vehicle_scheduler/test/test_all_stop_json_csv_schedule.cpp
```diff
@@ -20,15 +20,15 @@ namespace {
...
+        std::unique_ptr<vehicle_scheduler> scheduler;
 
         /**
          * @brief Test Setup method run before each test.
          * 
          */
         void SetUp() override {
             scheduler = std::unique_ptr<all_stop_vehicle_scheduler>(new all_stop_vehicle_scheduler());
-            scheduler->set_flexibility_limit(2);
+            // std::dynamic_pointer_cast<streets_vehicle_scheduler::all_stop_vehicle_scheduler>(scheduler)->set_flexibility_limit(2);
```
paulbourelly999, line 31: Remove commented out code.

scheduling_service/manifest.json
paulbourelly999: Oh ok initial green buffer is the time buffer to apply to the time when the phase will become green to avoid vehicles that are ahead of schedule entering before the phase is green. and final green is an additional buffer for the end of a green phase that allow additional safety on top of the yellow clearance and red clearance

scheduling_service/include/scheduling_service.h
```diff
@@ -17,35 +17,36 @@
...
+    private:
 
-		std::string bootstrap_server;
+        std::string bootstrap_server;
         std::string group_id;
         std::string consumer_topic;
         std::string producer_topic;
+        std::string spat_topic;
 
         std::shared_ptr<OpenAPI::OAIIntersection_info> intersection_info_ptr;
```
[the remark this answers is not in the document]
Saeid-Solei: This pointer is removed.

scheduling_service/src/intersection_client.cpp
```diff
@@ -3,55 +3,55 @@
...
-		});
+            [&intersection_info_valid, &loop, this]
+                ( [[maybe_unused]] const OpenAPI::OAIIntersection_info &int_info, QNetworkReply::NetworkError error, QString error_str ){ 
+            SPDLOG_ERROR("Error happened while issuing intersection model GET information request : {0} code : {1}",  error_str.toStdString(), error);
+            intersection_info_valid = false;
+            intersection_info_ptr = nullptr;
+            loop.quit(); 
+        });
 
         apiInstance.getIntersectionInfo();
```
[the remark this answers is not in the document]
Saeid-Solei: Which method are you talking about?

scheduling_service/CMakeLists.txt
```diff
@@ -37,7 +38,8 @@ target_link_libraries(${PROJECT_NAME} PUBLIC
     intersection_client_api_lib 
     streets_service_base_lib::streets_service_base_lib 
     streets_vehicle_list_lib::streets_vehicle_list_lib 
-    streets_vehicle_scheduler_lib::streets_vehicle_scheduler_lib)
+    streets_vehicle_scheduler_lib::streets_vehicle_scheduler_lib
+    streets_signal_phase_and_timing_lib::streets_signal_phase_and_timing_lib)
```
paulbourelly999, line 42: Technically you do not need to add dependencies of included targets like the scheduling library. The install step in the scheduling library should take care of this. Specifically the file inside /streets_utils/streets_vehicle_scheduler/cmake you need to add a 
```
find_dependency(streets_signal_phase_and_timing REQUIRED)

```

scheduling_service/src/scheduling_service.cpp
```diff
@@ -2,259 +2,313 @@
...
+            processor->set_timeout(streets_service::streets_configuration::get_int_config("exp_delta"));
+            
+            SPDLOG_INFO("Vehicle list is configured successfully! ");
+            return true;
+        }
+        else if (intersection_type.compare("signalized_intersection") == 0) {
+            vehicle_list_ptr->set_processor(std::make_shared<streets_vehicles::signalized_status_intent_processor>());
+            std::dynamic_pointer_cast<streets_vehicles::signalized_status_intent_processor>(vehicle_list_ptr->get_processor())->set_timeout(streets_service::streets_configuration::get_int_config("exp_delta"));
+
+            SPDLOG_INFO("Vehicle list is configured successfully! ");
```
paulbourelly999, line 127: Better log statement or remove this. It is the same one as line 120. Either change it to something intersection_type specific or remove it.

scheduling_service/src/scheduling_service.cpp
```diff
@@ -2,259 +2,313 @@
...
+        else {
+            SPDLOG_ERROR("Failed configuring Vehicle List. Scheduling Service does not support intersection_type : {0}!", intersection_type);
+            return false;
+        }
+        
     }
 
 
-	bool scheduling_service::config_scheduler() 
+    bool scheduling_service::config_scheduler(std::shared_ptr<OpenAPI::OAIIntersection_info> intersection_info_ptr)
```
paulbourelly999, line 138: Why do this method need a parameter. Isn't intersection_info_ptr already a member variable of scheduling_service?

scheduling_service/src/scheduling_service.cpp
```diff
@@ -2,259 +2,313 @@
...
+        }
+        else if ( intersection_type.compare("signalized_intersection") == 0 ) {
+            scheduler_ptr = std::make_shared<streets_vehicle_scheduler::signalized_vehicle_scheduler>();
+            scheduler_ptr->set_intersection_info(intersection_info_ptr);
+            
+            auto processor = std::dynamic_pointer_cast<streets_vehicle_scheduler::signalized_vehicle_scheduler>(scheduler_ptr);
+            processor->set_initial_green_buffer(streets_service::streets_configuration::get_int_config("initial_green_buffer"));
+            processor->set_final_green_buffer(streets_service::streets_configuration::get_int_config("final_green_buffer"));
+            
+            SPDLOG_INFO("Scheduler is configured successfully! ");
```
paulbourelly999, line 158: Better log statement. Same comment as above

scheduling_service/src/scheduling_service.cpp
```diff
@@ -2,259 +2,313 @@
...
+        SPDLOG_INFO("Starting spat consumer thread.");
+        while (spat_consumer_worker->is_running()) 
+        {
+            
+            const std::string spat_msg = spat_consumer_worker->consume(1000);
+
+            if(spat_msg.length() != 0 && spat_ptr)
+            {                
+
+                spat_ptr->fromJson(spat_msg);
```
paulbourelly999, line 200: this method throws exceptions. You may want to implement try_catch to shutdown gracefully. Just a thought

scheduling_service/src/scheduling_worker.cpp
```diff
@@ -4,23 +4,33 @@ namespace scheduling_service{
...
-			auto logger = spdlog::get("csv_logger");
-			if ( logger != nullptr ){
-				logger->info( int_schedule.toCSV());
-			}
-		}
-
-		return int_schedule;
-	}
+    std::shared_ptr<streets_vehicle_scheduler::intersection_schedule> scheduling_worker::schedule_vehicles(std::unordered_map<std::string, streets_vehicles::vehicle> veh_map, 
+                                                                                                std::shared_ptr<streets_vehicle_scheduler::vehicle_scheduler> scheduler) const
```
paulbourelly999, line 8: Why do you need this as a method parameter when its already a member variable of the class.

streets_utils/streets_vehicle_scheduler/include/signalized_vehicle_scheduler.h
```diff
@@ -0,0 +1,236 @@
...
+            /**
+             * @brief A shared pointer to the modified_spat.
+             * 
+             */
+            std::shared_ptr<signal_phase_and_timing::spat> spat_ptr;
+
+            /**
+             * @brief The configurable time interval at the beginning of a green phase that is considered for estimating vehicles' entering times.
+             * The earliest possible entering time during a phase is set to the beginning of the green phase plus the initial_green_buffer.
+             * to-do add unit
```
paulbourelly999, line 29: Make sure you add unit. It helps to write "TODO" like this because sonar scanner will see it as a code smell so we don't lose track of it.

streets_utils/streets_vehicle_scheduler/include/signalized_vehicle_scheduler.h
```diff
@@ -0,0 +1,236 @@
...
+             */
+            uint64_t initial_green_buffer;
+
+            /**
+             * @brief The configurable time interval at the end of a green phase that is considered for estimating vehicles' entering times.
+             * The latest possible entering time during a phase is set to the end of the green phase minus the final_green_buffer.
+             * to-do add unit
```
paulbourelly999, line 36: Add unit

streets_utils/streets_vehicle_scheduler/src/signalized_vehicle_scheduler.cpp
```diff
@@ -0,0 +1,397 @@
...
+            signalized_vehicle_schedule preceding_veh;
+            for (const auto &veh_sched : schedule->vehicle_schedules ) {
+                if ( veh_sched.entry_lane == entry_lane && veh_sched.state == streets_vehicles::vehicle_state::DV && (preceding_veh.v_id == "" || veh_sched.et < preceding_veh.et)) {
+                    preceding_veh = veh_sched;
+                }
+            }
+            
+            // Find the entry lane object
+            OpenAPI::OAILanelet_info entry_lane_info;
+            for ( const auto &lane : intersection_info->getEntryLanelets() ){
```
paulbourelly999, line 119: I think the base scheduler class has a helper method that does exactly this.

streets_utils/streets_vehicle_scheduler/src/signalized_vehicle_scheduler.cpp
```diff
@@ -0,0 +1,397 @@
...
+            signal_phase_and_timing::movement_state move_state = find_movement_state_for_lane(entry_lane_info);
+
+            for (const auto &ev : evs_in_lane){
+                signalized_vehicle_schedule sched;
+                SPDLOG_DEBUG( "Estimating schedule for {0}.", ev._id);
+                estimate_et(ev, preceding_veh, sched, move_state, schedule->timestamp);
+                // Add vehicle schedule to the schedule list.
+                schedule->vehicle_schedules.push_back(sched);
+                // Update the preceding vehicle schedule.
+                preceding_veh = sched;
```
paulbourelly999, line 137: It seems to me that the preceding vehicle gets reset on each loop from line 105. I don't think this is the intended behavior you wanted.

streets_utils/streets_vehicle_scheduler/src/signalized_vehicle_scheduler.cpp
```diff
@@ -0,0 +1,397 @@
...
+                    is_successful = true;
+                    break;
+                }
+                SPDLOG_DEBUG( "The estimated ET for vehicle {0} is later than the end of the phase. Estimated ET = {1}, end of the phase", veh._id, et, move_event.timing.get_epoch_min_end_time() - final_green_buffer );
+            }
+        }
+
+        // TBD area
+        if (!is_successful) {
+            et = std::max(first_available_et, std::max(eet, move_state.state_time_speed.back().timing.get_epoch_min_end_time() + initial_green_buffer));
```
paulbourelly999, line 209: We need to make note of the fact that we are expecting the movement event list to be ordered <first to last>

streets_utils/streets_vehicle_scheduler/src/signalized_vehicle_scheduler.cpp
```diff
@@ -0,0 +1,397 @@
...
+            double eet = t_accel + t_cruising + t_decel;
+            if (eet <= 0) {
+                SPDLOG_DEBUG("The estimated eet for vehicle {0} has a negative value = {1}!", veh._id, eet);
+            }
+            return static_cast<uint64_t>(ceil(eet * 1000.0)) + veh._cur_time;
+        }
+        else
+        {
+            SPDLOG_DEBUG("Cannot estimate the earliest entering time for vehicle {0} which is not an EV!", veh._id);
+            return -1;
```
paulbourelly999, line 265: Might be cleaner to through an exception here since it is an invalid input

streets_utils/streets_vehicle_scheduler/src/signalized_vehicle_scheduler.cpp
```diff
@@ -0,0 +1,397 @@
...
+        
+        return static_cast<uint64_t>(ceil( 1000*(veh._reaction_time + ((veh._length + veh._min_gap)/speed)) ));
+    } 
+
+
+    uint64_t signalized_vehicle_scheduler::estimate_clearance_time( const streets_vehicles::vehicle &veh ) const {
+        // Get Link Lane
+        OpenAPI::OAILanelet_info link_lane =  get_link_lanelet_info( veh );
+        if ( veh._cur_state == streets_vehicles::vehicle_state::DV ) {
+            return static_cast<uint64_t>( ceil(1000 * veh._cur_distance / link_lane.getSpeedLimit()) );
```
paulbourelly999, line 365: Have you thought about using vehicle speed for DVs. I know we are targeting link_lane speed limit but might be more accurate with vehicle speed. Just a thought

scheduling_service/CMakeLists.txt
Saeid-Solei: Done!

scheduling_service/src/scheduling_service.cpp
Saeid-Solei: Done!

scheduling_service/src/scheduling_service.cpp
Saeid-Solei: Per Anish's comment, since the intersection_info pointer in the scheduling service is only used for configuring the scheduler, I have removed the variable intersection_info_ptr from scheduling_service class and instead, pass the intersection_info pointer to the config_scheduler method.

scheduling_service/src/scheduling_service.cpp
Saeid-Solei: Done!

streets_utils/streets_vehicle_scheduler/include/signalized_vehicle_scheduler.h
Saeid-Solei: Done!

streets_utils/streets_vehicle_scheduler/include/signalized_vehicle_scheduler.h
Saeid-Solei: Done!

streets_utils/streets_vehicle_scheduler/src/signalized_vehicle_scheduler.cpp
```diff
@@ -0,0 +1,397 @@
...
+        
+        for ( const auto &[entry_lane, evs_in_lane] : vehicle_lane_map ) {           
+            
+            SPDLOG_DEBUG("Scheduling EVs from entry lane {0} ", entry_lane);
+
+            // Find the last DV from the entry lane
+            signalized_vehicle_schedule preceding_veh;
+            for (const auto &veh_sched : schedule->vehicle_schedules ) {
+                if ( veh_sched.entry_lane == entry_lane && veh_sched.state == streets_vehicles::vehicle_state::DV && (preceding_veh.v_id == "" || veh_sched.et < preceding_veh.et)) {
+                    preceding_veh = veh_sched;
```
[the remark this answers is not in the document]
Saeid-Solei: In the estimate_et method (lines 184-189), the method checks if the preceding_veh object has an empty id or not. If yes, the first available entering time (ET) before considering the earliest entering time (EET) will be the schedule timestamp. Otherwise, it will be the max of the schedule timestamp and the preceding vehicle's ET plus the minimum time headway.

streets_utils/streets_vehicle_scheduler/src/signalized_vehicle_scheduler.cpp
Saeid-Solei: The method you are referring to gets a vehicle object as an input. In this set of line, the algorithm did not initiate scheduling the vehicles yet. Basically, if we want to use vehicle_scheduler::get_entry_lanelet_info() method, we should just get the first EV's vehicle object and use it to get the entry lane object. If that is fine, I can make the changes.

streets_utils/streets_vehicle_scheduler/src/signalized_vehicle_scheduler.cpp
Saeid-Solei: That is the intended behavior actually. The for loop at line 105 iterates over all entry lanes. for each entry lane, we first find the preceding DV, if exist. Then we update the preceding vehicle info as we schedule the EVs one by one. When moving to the next entry lane, the preceding vehicle shall be updated to the DV that entered the intersection box from this new entry lane.

streets_utils/streets_vehicle_scheduler/src/signalized_vehicle_scheduler.cpp
Saeid-Solei: Correct. using vehicle speed and acceleration\deceleration pieces definitely increases the accuracy of the estimated DT of DVs. But since this estimated DT is not used in the scheduling algorithm, I thought maybe we can go with the simple inaccurate calculation. In an ideal case, the DT of DVs shall be estimated the same way as we estimate ET. 

Now basically, there is no need for scheduling DVs and estimating their DTs at all. but the reason I kept it in this library is that we need to know the ET of the last DV that entered the intersection box from each entry lane. The schedule_dvs method will include all DVs in the signalized_intersection_schedule object and can be easily accessible by the schedule_evs method.

streets_utils/streets_vehicle_scheduler/src/signalized_vehicle_scheduler.cpp
```diff
@@ -0,0 +1,397 @@
...
+
+
+    uint64_t signalized_vehicle_scheduler::estimate_clearance_time( const streets_vehicles::vehicle &veh ) const {
+        // Get Link Lane
+        OpenAPI::OAILanelet_info link_lane =  get_link_lanelet_info( veh );
+        if ( veh._cur_state == streets_vehicles::vehicle_state::DV ) {
+            return static_cast<uint64_t>( ceil(1000 * veh._cur_distance / link_lane.getSpeedLimit()) );
+        }
+        else {
+            return static_cast<uint64_t>( ceil(1000 * link_lane.getLength() / link_lane.getSpeedLimit()) );
```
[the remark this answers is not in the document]
Saeid-Solei: The value inside the ceil method will have decimals. However, the output of this method is milliseconds, not second with decimals. We can round it down as well, but the conservative solution will be rounding it up.

scheduling_service/src/scheduling_service.cpp
Saeid-Solei: Done!

scheduling_service/src/scheduling_worker.cpp
Saeid-Solei: It is not a member variable of the class currently.

scheduling_service/test/test_scheduling_service.cpp
```diff
@@ -64,12 +56,12 @@ TEST(scheduling_service_test, start)
...
     auto info =  std::make_shared<OpenAPI::OAIIntersection_info>();
     info->fromJson(QString::fromStdString(json));
-    ss.set_intersection_info( info );
+    // ss.set_intersection_info( info );
```
paulbourelly999, line 59: remove these commented calls

scheduling_service/test/test_scheduling_service.cpp
```diff
@@ -28,23 +28,15 @@ TEST(scheduling_service_test, config_vehicle_list)
...
```
paulbourelly999, line 37: Remove these commented calls

Saeid-Solei: Done!

scheduling_service/test/test_scheduling_service.cpp
Saeid-Solei: Done!

paulbourelly999 (approved):
Looks good

PR #168, opened 2022-07-25
#169: monitor tsc state

adev4a:
# PR Details
This PR adds functionality to the  tsc_client_service to monitor the traffic signal controller(TSC) state. The state is defined as a collection of configurations which are requested from the TSC through SNMP get calls.
The tsc_client_service requests and processes configuration information from the TSC to create a state definition that can be used to modify phases according to a known plan.

## Related Issue
https://github.com/usdot-fhwa-stol/carma-streets/issues/169

## How Has This Been Tested?
Unit Tested and Integration tested.

## Types of changes

## Checklist:


tsc_client_service/CMakeLists.txt
```diff
@@ -51,6 +51,8 @@ set(SOURCES ${TEST_SOURCES} WORKING_DIRECTORY  ${PROJECT_SOURCE_DIR}/test)
 add_executable(${PROJECT_NAME}_test 
     ${TEST_SOURCES}
 )
-target_link_libraries(${PROJECT_NAME}_test ${PROJECT_NAME}_lib ${GTEST_LIBRARIES})
+target_link_libraries(${PROJECT_NAME}_test
```
paulbourelly999, line 54: is there a cleaner way to do this? From the googletest README.md it looks like you should be able to just call
```
find_package(GTest::gtest)
find_package(GTest::gmock)

target_link_libraries(${PROJECT_NAME}_test PUBLIC GTest::gtest GTest::gmock)
or 
target_link_libraries(${PROJECT_NAME}_test PUBLIC gtest gmock)

```

tsc_client_service/include/monitor_tsc_state.h
```diff
@@ -4,31 +4,49 @@
 
 namespace traffic_signal_controller_service
 {
+    struct signal_group_state
+    {
+        int signal_group_id;
```
paulbourelly999, line 9: Leave comment about how `signal_group_id` is J2735 identifier and `phase_num` is NTCIP identifier

tsc_client_service/include/monitor_tsc_state.h
```diff
@@ -4,31 +4,49 @@
 
 namespace traffic_signal_controller_service
 {
+    struct signal_group_state
```
paulbourelly999, line 7: For classes/structs like this we should start leaving javadoc comments about what these are meant to be used for and how they are used.

tsc_client_service/include/monitor_tsc_state.h
```diff
@@ -4,31 +4,49 @@
...
+        /*Maximum green durations for each active vehicle phase*/
+        int max_green;
+        /*Default Green duration for phase */
+        int green_duration;
+        /*Yellow signal duration for phase */
+        int yellow_duration;
+        /* Red clearace time for phase */
+        int red_clearance;
+        /*Red signal duration for phase */
+        int red_duration;
```
paulbourelly999, line 23: Since this is not a NTCIP defined property, we should probably make sure we explicitly define what this measurement is. In our case it is the total time a phase is predicted to be red before its next green transition right?

tsc_client_service/include/monitor_tsc_state.h
```diff
@@ -4,31 +4,49 @@
 
 namespace traffic_signal_controller_service
 {
+    struct signal_group_state
+    {
+        int signal_group_id;
+        /*Phase associated with signal group*/
+        int phase_num;
+        /*Minimum green durations for each active vehicle phase*/
+        int min_green;
```
paulbourelly999, line 13: None of these have units. I assume they are milliseconds right.

tsc_client_service/include/monitor_tsc_state.h
```diff
@@ -4,31 +4,49 @@
...
+        /* Mapping between signal group ids(key) and their states(value) defined as a signal group state struct*/
+        std::unordered_map<int, signal_group_state> signal_group_state_map_;
+
         /* Map between phase numbers(key) and signal group ids(value) for all active vehicle phases in the Traffic Signal Controller*/
         std::unordered_map<int,int> phase_num_map_;
 
-        /* Map between signal group ids(key) and phase numbers(value) for all active vehicle phases in the Traffic Signal Controller*/
-        std::unordered_map<int,int> signal_group_map_;
+        /* The sequence of phases in ring 1 of TSC*/
+        std::vector<int> phase_seq_ring1_;
```
paulbourelly999, line 46: once we add pedestrian phases we may want to ensure that it is clear we are assuming the first two rings are vehicle phases only.

tsc_client_service/include/snmp_client.h
```diff
@@ -15,6 +15,28 @@ enum class request_type
...
 };
 
+/** @brief A struct to hold the value being sent to the TSC, can be integer or string. Type needs to be defined*/
+struct snmp_response_obj
+{
+    /** @brief The type of value being requested or set, on the TSC */
+    enum class response_type
+    {
+        INTEGER,
+        STRING
```
paulbourelly999, line 25: How is phase sequence returned. Is it `STRING` response type?

tsc_client_service/src/monitor_tsc_state.cpp
```diff
@@ -60,18 +153,172 @@ namespace traffic_signal_controller_service
...
         
+        return red_duration; 
+    }
+
+    std::vector<int> tsc_state::phase_seq(int ring_num)
+    {
+        std::vector<int> phase_seq;
+        // Read sequence 1 data for first 2 rings
+        request_type request_type = request_type::GET;
+        std::string phase_seq_oid= ntcip_oids::SEQUENCE_DATA + "." + "1" + "." + std::to_string(ring_num); //Sequence 1 for ring
```
paulbourelly999, line 274: why is the additional `.1.` not part of the constant?

tsc_client_service/src/snmp_client.cpp
```diff
@@ -111,23 +117,47 @@ bool snmp_client::process_snmp_request(const std::string& input_oid, const reque
                 // get Integer value
                 if(vars->type == ASN_INTEGER){
                     if(vars->val.integer){
-                        value = *vars->val.integer;
-                        SPDLOG_INFO("Integer value in object: {0}", value);
+                        val.val_int = *vars->val.integer;
+                        SPDLOG_INFO("Integer value in object: {0}", val.val_int);
```
paulbourelly999, line 121: Does this log statement need to be info level.

paulbourelly999 (changes requested):
Left some comments about documentation mostly. Everything looks good to me. The only change I am passionate about is the CMakeList.txt file changes. Seems like a weird way to include a pretty standard library. Please check if we can use find_package instead

tsc_client_service/include/snmp_client.h
adev4a: Yes, Sequence and concurrent phases are returned as string responses

tsc_client_service/src/monitor_tsc_state.cpp
adev4a: The .1. defines the sequence number and the next constant defines the ring. Since we are only querying sequence 1 by default, that's hard coded here

paulbourelly999 (approved):
Looks good to me

tsc_client_service/include/monitor_tsc_state.h
```diff
@@ -3,32 +3,53 @@
...
+    {
+        /* signal group id identifier for J2735 which is mapped to phase number in NTCIP */
+        int signal_group_id;
+        /*Phase associated with signal group*/
+        int phase_num;
+        /*Minimum green durations for each active vehicle phase in miiliseconds*/
+        int min_green;
+        /*Maximum green durations for each active vehicle phase in miiliseconds*/
+        int max_green;
+        /*Default Green duration for phase in miiliseconds*/
```
dan-du-car, line 19: milliseconds

tsc_client_service/src/monitor_tsc_state.cpp
```diff
@@ -10,35 +10,127 @@ namespace traffic_signal_controller_service
...
+
+        // Loop through states once other state parameters are defined to get the red duration
+        for(auto& state : signal_group_state_map_)
+        {
+            state.second.red_duration = get_red_duration(state.second.phase_num);
+        }
+
+
+        // Print signal_group map
+        for (const auto& phase : signal_group_phase_map_)
```
dan-du-car, line 43: Can we remove this for loops just for print purpose?

tsc_client_service/src/snmp_client.cpp
```diff
@@ -111,23 +117,47 @@ bool snmp_client::process_snmp_request(const std::string& input_oid, const reque
...
-                    SPDLOG_ERROR("Received a message type which isn't an Integer");
+                    SPDLOG_ERROR("Received a message type which isn't an integer or string");
                     return false;
                 }
             }
         }
         else if(request_type == request_type::SET){
-            SPDLOG_DEBUG("Success in SET for OID: {0}", input_oid," ; Value: {1}", value);
+            
+            if(val.type == snmp_response_obj::response_type::INTEGER){
```
dan-du-car, line 151: Why do we have two different response types, and it seems to only print the data after the conditional checking.

tsc_client_service/src/monitor_tsc_state.cpp
adev4a: I've added this intentionally so we can see the mapped values. I think it would be useful to keep in the log, but if you think its not, i can remove it

tsc_client_service/src/snmp_client.cpp
adev4a: We have 2 different types of values that we are requesting, integer and string. This represents that. 
Since the request method is common for both of them, specifying the type lets us know which response type to process

Issue #169, opened 2022-07-25
Implement Traffic Signal Controller state monitoring 


adev4a:
### Types of Issue

- [ ] Anomaly report (something appears to not work correctly)
- [x] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

In order to support Hold/Force-off and Omit calls, the traffic signal controller service needs to be aware of the configuration on the TSC.This is done by sending multiple snmp requests to the TSC and then using this information in addition to the broadcasted SPaT message to understand how the phases should be modified.
The identified fields to be requested from the TSC are:
Maximum Green Duration, Minimum Green Duration, Mapping between Channel Number and vehicle phases, Yellow Duration, Red Clearance Time, Phase sequence in a ring and phases concurrent to a given phase (will be active at the same time)

After obtaining the phase configuration information, the spat message needs to be updated with the calculated future movement events for the signal groups associated with a vehicle phase.

### Expected behavior
The tsc_client_service can generate a state data structure with the required information for all vehicle phases.
### Actual behavior
N/A

### Steps to reproduce the actual behavior

N/A

PR #181, opened 2022-08-11
add following events: Ref Issue # 169

adev4a:
# PR Details
This PR adds methods to update incoming spat with future movement events for each signal group in the spat msg.
## Related Issue
 Implement Traffic Signal Controller state monitoring #169 

## Types of changes

## Checklist:

### DIFF
--- a/tsc_client_service/include/monitor_tsc_state.h
+++ b/tsc_client_service/include/monitor_tsc_state.h
@@ -150,8 +150,8 @@ namespace traffic_signal_controller_service
             ** @param phase_state signal_group_state for the phase of which the movement events are a part
             ** @return predicted next movement event
             * **/
-            signal_phase_and_timing::movement_event get_following_event(signal_phase_and_timing::movement_event& current_event,
-                                                                 uint64_t current_event_end_time, const signal_group_state& phase_state);
+            signal_phase_and_timing::movement_event get_following_event(const signal_phase_and_timing::movement_event& current_event,
+                                                                 uint64_t current_event_end_time, const signal_group_state& phase_state) const;
 
             //Add Friend Test to share private members
             FRIEND_TEST(traffic_signal_controller_service, test_get_following_movement_events);                                                              
--- a/tsc_client_service/include/monitor_tsc_state.h
+++ b/tsc_client_service/include/monitor_tsc_state.h
@@ -3,6 +3,8 @@
 #include "snmp_client.h"
 #include "snmp_client_exception.h"
 #include "ntcip_oids.h"
+#include "spat.h"
+#include <gtest/gtest_prod.h>
 
 namespace traffic_signal_controller_service
 {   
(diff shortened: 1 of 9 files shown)


tsc_client_service/include/monitor_tsc_state.h
```diff
@@ -123,6 +132,30 @@ namespace traffic_signal_controller_service
...
+            ** @param current_event movement_event from the next movement needs to be predicted
+            ** @param current_event_end_time End time of the current event in epoch time (milliseconds)
+            ** @param phase_state signal_group_state for the phase of which the movement events are a part
+            ** @return predicted next movement event
+            * **/
+            signal_phase_and_timing::movement_event get_following_event(const signal_phase_and_timing::movement_event& current_event,
+                                                                 uint64_t current_event_end_time, const signal_group_state& phase_state) const;
+
+            //Add Friend Test to share private members
+            FRIEND_TEST(traffic_signal_controller_service, test_get_following_movement_events);
```
paulbourelly999, line 157: Could you link some documentation to how you came up with this implementation to help me understand it.

tsc_client_service/src/monitor_tsc_state.cpp
```diff
@@ -80,6 +80,106 @@ namespace traffic_signal_controller_service
...
+                case signal_phase_and_timing::movement_phase_state::protected_movement_allowed : //Green
+                    current_event_end_time_epoch = start_time + phase_state.green_duration;
+                    break;                                                                
+
+                case signal_phase_and_timing::movement_phase_state::protected_clearance : //Yellow
+                    current_event_end_time_epoch = start_time + phase_state.yellow_duration;
+                    break;
+
+                default:
+                    SPDLOG_ERROR("This movement phase is not supported. Movement phase type: {0}", int(current_movement.state_time_speed.front().event_state));
```
paulbourelly999, line 164: Please confirm this but I believe unused phases will be movement_phase_state::dark and pedestrian phases will be walk clear or don't walk so this is probably not an exit condition. Maybe a debug log will help us see when this condition is meant. Similar stuff can be said for the switch statement above.

tsc_client_service/src/monitor_tsc_state.cpp
```diff
@@ -80,6 +80,106 @@ namespace traffic_signal_controller_service
...
+        for (auto movement : spat_ptr->intersections.front().states)
+        {
+            int signal_group_id = movement.signal_group;
+
+            // Assumption here is that the movement states have only one event, since this method adds future events to the list.
+            //Throw exception is list size is great than 1
+            if (movement.state_time_speed.size() > 1)
+            {
+                SPDLOG_ERROR("Event list has more than one events, not usable when adding future movement events. Associated with Signal Group: {0}", signal_group_id);
+                throw snmp_client_exception("Event list has more than one events, not usable when adding future movement events. Associated with Signal Group:" + std::to_string(signal_group_id));
```
paulbourelly999, line 130: Lets use a  different exception for this, since it really does not have anything to do with SNMP client.

tsc_client_service/src/monitor_tsc_state.cpp
```diff
@@ -80,6 +80,106 @@ namespace traffic_signal_controller_service
...
+                    current_event_end_time_epoch = start_time + phase_state.green_duration;
+                    break;                                                                
+
+                case signal_phase_and_timing::movement_phase_state::protected_clearance : //Yellow
+                    current_event_end_time_epoch = start_time + phase_state.yellow_duration;
+                    break;
+
+                default:
+                    SPDLOG_ERROR("This movement phase is not supported. Movement phase type: {0}", int(current_movement.state_time_speed.front().event_state));
+                    return false; //TODO: Should this be an exit condition?
```
paulbourelly999, line 165: Since this is the only condition that will cause this method to return false maybe we should switch this to a void method and catch any exception it throughs.

tsc_client_service/include/monitor_tsc_state.h
adev4a: Yes, sure. Here's some documentation describing it https://google.github.io/googletest/reference/testing.html#FRIEND_TEST

tsc_client_service/src/monitor_tsc_state.cpp
adev4a: So, phases that are not vehicle phases would never be evaluated at this point, because on Line 141 we ignore phases not in the state_map.
These would however include vehicle phases like dark, permissive_movement_allowed etc.I can change this to a debug log, but we also need to add an exception here, since it would fail Line 167 and 168

tsc_client_service/include/tsc_service.h
```diff
@@ -150,5 +151,9 @@ namespace traffic_signal_controller_service {
              */
             void produce_spat_json() const;
 
+            // Configurable boolean to enable tsc_state to update incoming spat with future movement events calculated using 
+            // traffic signal controller configuration information
+            bool use_tsc_state_spat_update_ = true;
```
paulbourelly999, line 156: We probably want to read this parameter from the manifest json using the streets_configuration singleton like we do for the other configuration parameter instead of hard coding it.

paulbourelly999 (changes requested):
Use streets_configuration to read configuration from manifest json

tsc_client_service/include/monitor_tsc_state.h
```diff
@@ -55,6 +58,13 @@ namespace traffic_signal_controller_service
             /* The sequence of vehicle phases in ring 2 of TSC*/
             std::vector<int> phase_seq_ring2_;
 
+            // Number of required following movements on receiving a spat_ptr
+            int required_following_movements_ = 3;
```
Saeid-Solei, line 62: Could you explain what is this parameter? is this the configurable number of future phases in the modified spat?

paulbourelly999: Yes we could add it but we probably don't need to add it as a parameter read from the manifest json

tsc_client_service/src/tsc_service.cpp
```diff
@@ -23,6 +23,7 @@ namespace traffic_signal_controller_service {
                 return false;
             }
             //Initialize TSC State
+            bool use_tsc_state_spat_update_ = streets_service::streets_configuration::get_boolean_config("use_tsc_state_spat_update");
```
paulbourelly999, line 26: This is incorrect. Why do you have a class member and a local variable. Remove the local variable and just set the class member

paulbourelly999 (approved):
Looks good to me!

dan-du-car (approved):

PR #189, opened 2022-08-24
Update spat with desired phase plan

dan-du-car:
# PR Details
## Description
 For TMSO UC3, CARMA-Streets is responsible for optimizing traffic signal phases and timing at a signalized intersection. The TSC (Traffic Signal Controller) Service is responsible for making the relevant NTCIP calls to the traffic signal controller to produce the requested signal phases and timings. The request signal phases and timings will be provided by the Signal Optimization (SO) Service as a Desired Phase Plan. The scope of this story is to design and document Desired Phase Plan message and the communication sequence between SO Service and the TSC Service. The scope also includes implementing a kafka consumer inside the TSC Service to consume Desired Phase Plan and update the Modified SPaT MovementEventList data with Desired Phase Plan.

## Related Issue
#197 

## Motivation and Context
UC3

## How Has This Been Tested?
Unit test

## Types of changes

## Checklist:


tsc_client_service/src/tsc_service.cpp
```diff
@@ -151,6 +170,9 @@ namespace traffic_signal_controller_service {
                         catch(const traffic_signal_controller_service::monitor_states_exception &e){
                             SPDLOG_ERROR("Could not update movement events, spat not published. Encounted exception : \n {0}", e.what());
                         }
+                    }else if(use_desired_phase_plan_update_){
```
Saeid-Solei, line 173: If both use_tsc_state_spat_update_ and use_desired_phase_plan_update_ are true, then the modified spat will not be updated based on the desired_phase_plan_update. Is this the expected behavior?

tsc_client_service/src/tsc_service.cpp
```diff
@@ -167,11 +189,25 @@ namespace traffic_signal_controller_service {
...
     }
 
+    void tsc_service::consume_desired_phase_plan() const {
+       while (desired_phase_plan_consumer->is_running())
+        {
+            const std::string payload = desired_phase_plan_consumer->consume(1000);
+            if (payload.length() != 0)
+            {
+                SPDLOG_DEBUG("Consumed: {0}", payload);
+                std::scoped_lock<std::mutex> lck{dpp_mtx};
```
Saeid-Solei, line 199: Would it be possible to use lock inside the method rather than here?

tsc_client_service/src/monitor_desired_phase_plan.cpp
```diff
@@ -0,0 +1,277 @@
...
+            else if (cur_movement_state_ref.state_time_speed.front().event_state == signal_phase_and_timing::movement_phase_state::protected_clearance) // YELLOW
+            {
+                // Updating the current YELLOW movement event
+                int spat_current_yellow_duration = tsc_state_ptr->get_signal_group_state_map()[current_signal_group_id].yellow_duration;
+                uint64_t calculated_yellow_end_time_epoch = cur_movement_state_ref.state_time_speed.front().timing.get_epoch_start_time() + spat_current_yellow_duration;
+                cur_movement_state_ref.state_time_speed.front().timing.set_min_end_time(calculated_yellow_end_time_epoch);
+
+                // Add RED movement event
+                signal_phase_and_timing::movement_event red_movement_event;
+                uint64_t calculated_red_end_time_epoch = desired_green_end_time_epoch + desired_yellow_duration + desired_red_clearance;
```
Saeid-Solei, line 169: To be consistent with the yellow start and end time, use the same timing used above.
Basically, the calculated_red_end_time_epoch shall be equal to the cur_movement_state_ret start time + spat_current_yellow_duration + desired_red_clearance.

tsc_client_service/src/monitor_desired_phase_plan.cpp
```diff
@@ -0,0 +1,277 @@
...
+            {
+                // Updating RED movement event
+                cur_movement_state_ref.state_time_speed.front().timing.set_min_end_time(desired_green_start_time_epoch);
+
+                // Append GREEN, YELLOW, and RED movement events
+                append_full_green_yellow_red_phases_by_desired_green(cur_movement_state_ref, desired_sg_green_timing, desired_yellow_duration, desired_red_clearance);
+            }
+            else
+            {
+                SPDLOG_ERROR("SPAT current movement event has to be green or red event state!");
```
Saeid-Solei, line 138: Why there is no case for yellow? Is this because we do not allow a phase to be repeated? If yes, I would say let have a case here to be general. We can define a configurable indicator in the SO service to allow or don't allow a phase to be repeated.

tsc_client_service/src/tsc_service.cpp
dan-du-car: It makes sense to lock in the service level, and the other methods are called by this one update function in this service.

tsc_client_service/src/tsc_service.cpp
dan-du-car: I think if both true, then use_tsc_state_spat_update_ is used. If the use_tsc_state_spat_update_ is false, then desired phase plan is used.

dan-du-car: We can update the manifest.json to mark use_tsc_state_spat_update_ default to false, use_desired_phase_plan_update_ to default true if not the desired behavior.  We do not have hard requirement saying which one shall be the expected behavior.

streets_utils/streets_signal_phase_and_timing/include/time_change_details.h
```diff
@@ -113,6 +113,20 @@ namespace signal_phase_and_timing{
...
          */
         uint64_t get_epoch_next_time() const;
+        /**
+         * @brief Convert epoch timestamp in milliseconds  into milliseconds of the current hours
+         *
+         * @param hour_tenth_secs  The tenth of seconds of the current hour
+         * @return uint64_t epoch timestamp in milliseconds
+         */
+        uint16_t convert_msepoch_to_hour_tenth_secs(uint64_t epoch_time_ms) const;
+        void set_start_time(uint64_t epoch_time_ms);
```
paulbourelly999, line 123: I will add header comments for each of these methods

tsc_client_service/include/monitor_desired_phase_plan.h
```diff
@@ -0,0 +1,28 @@
...
+
+namespace traffic_signal_controller_service
+{
+    class monitor_desired_phase_plan
+    {
+    private:
+        std::shared_ptr<streets_desired_phase_plan::streets_desired_phase_plan> desired_phase_plan_ptr;
+
+        // Conversion constants
+        void process_first_desired_green(signal_phase_and_timing::movement_state &cur_movement_state_ref, const streets_desired_phase_plan::signal_group2green_phase_timing &desired_sg_green_timing, const std::shared_ptr<tsc_state> tsc_state_ptr) const;
```
paulbourelly999, line 15: I will include comments for these methods.

tsc_client_service/src/monitor_desired_phase_plan.cpp
paulbourelly999: Please refer to documentation on confluence design review page.

tsc_client_service/src/tsc_service.cpp
paulbourelly999: I think we should have a single configurable parameter since they are mutually exclusive.

tsc_client_service/src/monitor_desired_phase_plan.cpp
Saeid-Solei: In Dan's design, she assumes that if a signal group is included in the first movement group from the desired spat plan, it cannot have a yellow signal state. The reason I guess, is that if the current signal state is yellow, it means it was green before that. And since the signal group is included in the first movement group from the desired spat plan, it will be green again after the current yellow and upcoming red clearance. Now this make sense if we assume no green phase can be repeated immediately. However, I feel like it is better to leave the responsibility of making such a decision to the SO service. From the TSC service side, it has to regard the yellow signal state in this case similar to the red clearance signal state.

paulbourelly999: @Saeid-Solei  we spoke about this. We will address any possibility of having back to back green phases for a single single group in a separate issue if we even need it.

tsc_client_service/src/monitor_desired_phase_plan.cpp
paulbourelly999: So desired yellow duration and desired red clearance could actually take different values than the red clearance and yellow change for this signal group. To explain, since two signal groups in a movement group are not required to have the same yellow change and red clearance it is possible for them to be different. In a case where they are different the max combination of the two will influence the end time of any red phase predicted from this desired phase plan entry since the red phase should extend passed its out red clearance to the max yellow change plus red clearance of the signal group currently (or most recently ) grant green. Please let me know if that makes sense.

streets_utils/streets_signal_phase_and_timing/src/models/time_change_details.cpp
```diff
@@ -89,6 +89,14 @@ namespace signal_phase_and_timing {
         return epoch_start_time;
     }
 
+    uint16_t time_change_details::convert_msepoch_to_hour_tenth_secs(uint64_t epoch_time_ms) const{
+        auto system_time = std::chrono::system_clock::now();
+        auto duration = system_time.time_since_epoch();
+        auto hours_since_epoch = std::chrono::duration_cast<std::chrono::hours>(duration).count();
+        auto hour_tenth_secs = (epoch_time_ms - (hours_since_epoch * HOUR_TO_SECONDS * SECOND_TO_MILLISECONDS))/100;
```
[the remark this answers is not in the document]
paulbourelly999: Not a complicate change but the name is meant to indicate that this is tenths of seconds from the current hour. I can change it though if necessary.

adev4a (approved):

tsc_client_service/src/monitor_desired_phase_plan.cpp
Saeid-Solei: I agree. But still, if the spat_current_yellow_duration is used for calculating calculated_yellow_end_time_epoch above, the same spat_current_yellow_duration shall be used for calculating calculated_red_end_time_epoch regardless of whether we want to use the max value or not.

Other than that, I do not see any max value calculation. Basically, calculated_red_end_time_epoch shall be max of:

1 - the start of the current yellow phase + spat_current_yellow_duration + desired_red_clearance
2 - the start of the current yellow phase + the current yellow duration of the other signal group from the same movement group + the desired red clearance of the other signal group from the same movement group

Am I correct?

Saeid-Solei: looks correct to me!

Saeid-Solei: The only comment I have left is that signal_group_with_largest_clearance does not necessary have the larger yellow duration. However, we are adding the desired_yellow_duration of signal_group_with_largest_clearance when we need to update the end of the yellow duration. 
Not a big deal since the scheduling algorithm regards the yellow phase as a red phase (it does not allow an entering time falls into a yellow phase). Just wanted to bring this up in case it caused issues later on!

tsc_client_service/src/tsc_service.cpp
Saeid-Solei: Are we planning to address this later?

tsc_client_service/src/monitor_desired_phase_plan.cpp
paulbourelly999: Created the following issue to address this. (https://github.com/usdot-fhwa-stol/carma-streets/issues/200)

Saeid-Solei (approved):

PR #192, opened 2022-08-29
 Issue-202: broadcast_tsc_configuration

adev4a:
# PR Details
This PR adds a kafka publisher to the TSC service to broadcast TSC configuration state information required by the SO service in order to create movement events using the desired phase plan.

A streets utils package is added to encode-decode the tsc config state to/from json in order to broadcast the information with kafka.
## Related Issue
#202 

## How Has This Been Tested?
Unit Tested and Locally Integration Tested.

## Types of changes

## Checklist:


streets_utils/streets_tsc_configuration/CMakeLists.txt
```diff
@@ -0,0 +1,89 @@
...
+set(SOURCES ${TEST_SOURCES} WORKING_DIRECTORY  ${PROJECT_SOURCE_DIR}/test)
+
+
+add_executable(${BINARY} ${TEST_SOURCES}
+                )
+
+add_test(NAME ${BINARY} COMMAND ${BINARY})
+target_include_directories(${BINARY} PUBLIC ${PROJECT_SOURCE_DIR}/include)
+target_link_libraries(${BINARY} PUBLIC 
+                Boost::system
```
paulbourelly999, line 82: If you include the library you do not need to include its dependencies. Those will automatically be included I think.

streets_utils/streets_tsc_configuration/README.md
```diff
@@ -0,0 +1,44 @@
...
+            "signal_group_id": 2,
+            "yellow_change_duration":2000,
+            "red_clearance":300,
+            "concurrent_signal_groups":[5,6]
+        }
+}
+```
+
+# TSC Configuration State Information
+Traffic signal controller configuration represents the setup on the physical device. In order to access that information SNMP requests need to be sent to the system. In order to control the current phases/signal_groups on the controller, this information is required to understand ring-barrier setups
```
paulbourelly999, line 24: When writing a paragraph in MD you do not need to add new line characters. The viewer will line wrap appropriately.

tsc_client_service/include/tsc_service.h
```diff
@@ -58,8 +72,18 @@ namespace traffic_signal_controller_service {
             std::shared_ptr<intersection_client> intersection_client_ptr;
             
             // Configurable boolean to enable tsc_state to update incoming spat with future movement events calculated using 
+            // traffic signal controller configuration information
+            bool use_tsc_state_spat_update_ = true;
```
paulbourelly999, line 76: Why did you add this parameter back. We only need a single parameter for using spat or desired phase plan

tsc_client_service/src/tsc_service.cpp
```diff
@@ -183,16 +195,33 @@ namespace traffic_signal_controller_service {
...
+    void tsc_service::produce_tsc_config_json() {
+        // Publish tsc_config information 10 times
+        try {
+            while(tsc_config_state_ptr && tsc_config_producer_counter_ < 10)
+            { 
+                
+                tsc_config_producer->send(tsc_config_state_ptr->toJson());
+                tsc_config_producer_counter_ ++;
+
+                std::this_thread::sleep_for(std::chrono::milliseconds(100)); // Sleep for 5 seconds between publish
```
paulbourelly999, line 217: This will only sleep for .1 seconds right? (100 ms = 0.1 seconds)

paulbourelly999 (changes requested):
Left some comments

tsc_client_service/src/tsc_service.cpp
adev4a: Yes, I had to change this because some of the other unit tests were timing out with a 5 second sleep. I can update the comment

tsc_client_service/include/tsc_service.h
adev4a: Must have come through with a bad merge with develop, I'll remove it

streets_utils/streets_tsc_configuration/CMakeLists.txt
adev4a: Not sure i understand which dependencies you're talking about here

streets_utils/streets_tsc_configuration/README.md
```diff
@@ -0,0 +1,43 @@
...
+            "yellow_change_duration":2000,
+            "red_clearance":300,
+            "concurrent_signal_groups":[5,6]
+        }
+}
+```
+
+# TSC Configuration State Information
+Traffic signal controller configuration represents the setup on the physical device. In order to access that information SNMP requests need to be sent to the system. In order to control the current phases/signal_groups on the controller, this information is required to understand ring-barrier setups 
+as well as the duration of yellow and red clearance, which cannot be controlled directly.The tsc_config_state message consists of this information for all **active vehicle phases/signal_groups**.
```
paulbourelly999, line 25: This does not need an end line character

streets_utils/streets_tsc_configuration/src/tsc_configuration_state.cpp
```diff
@@ -0,0 +1,114 @@
...
+        
+        state.AddMember("red_clearance", red_clearance, allocator);
+
+        if( !concurrent_signal_groups.empty()) {
+            rapidjson::Value concurrent_groups(rapidjson::kArrayType);
+            for(const auto &signal_group : concurrent_signal_groups){
+                concurrent_groups.PushBack(signal_group, allocator);
+            }
+            state.AddMember("concurrent_signal_groups", concurrent_groups, allocator);
+        }
```
Saeid-Solei, line 26: Would it be better if we always have the member, but leave it empty if the signal group does not have concurrent signal groups?

streets_utils/streets_tsc_configuration/test/test_tsc_configuration_state.cpp
```diff
@@ -0,0 +1,57 @@
...
+    state.concurrent_signal_groups.push_back(2);
+    state.concurrent_signal_groups.push_back(3);
+
+    auto json = state.toJson(allocator);
+    ASSERT_EQ(10, json["yellow_change_duration"].GetUint());
+
+    streets_tsc_configuration::tsc_configuration_state config;
+    config.tsc_config_list.push_back(state);
+    auto config_json = config.toJson();
+    ASSERT_FALSE(config_json.empty());
```
Saeid-Solei, line 55: I do not see any test related to concurrent_signal_groups. Since the data type of the concurrent_signal_groups is different than others, could you add an Assert statement for the concurrent_signal_groups?

streets_utils/streets_tsc_configuration/src/tsc_configuration_state.cpp
adev4a: I would be ok with that, but i wasn't sure how the Get Int call would work in cases where the list is empty. I couldn't find a method to check size of the array before reading. 
Either way this is just for conversion to and from Json. For any service using the message it would effectively be the same, since the state struct still has a concurrent_signal_groups member, which would be empty.

paulbourelly999: @adev4a  Either way works but please make sure you document in the readme how this will be treated if there are no concurrent groups, preferably as a comment to the JSON payload and also in a paragraph somewhere. Currently it looks like we are treating `concurrent_signal_groups` as a optional field which is fine. Please leave a comment that when it is empty the field will be missing. Good catch @Saeid-Solei

paulbourelly999 (changes requested):
Please update documentation to describe how you expect phases with no concurrent signal groups to be sent.

Saeid-Solei (approved):

paulbourelly999 (approved):

Issue #195, opened 2022-08-29
Fix signalized_vehicle_scheduler's understanding of signal group\lanelet mapping


Saeid-Solei:
### Types of Issue

- [x] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

Currently, the signalized_vehicle_scheduler assumes that each entry lane from the intersection_model is connected to a given signal group id and has a signal group id field. But actually, each connection link inside the intersection box have a signal group id field rather than each entry lane. 

In signalized_vehicle_scheduler, there is a method that provides the movement_state list for each entry lane based on the entry lane id and its associated signal group ids. This method needs to be updated to provide the movement_state list for each vehicle\entry lane  based on the connection lane ids and their associated signal group ids.

### Carma streets version where this issue was discovered

4.2.0

### Expected behavior
signalized_vehicle_scheduler gets the connection link ids mapping to signal group ids from the intersection model.

### Actual behavior

signalized_vehicle_scheduler assumes that the intersection model provides the entry lane ids mapping to signal group ids (rather than connection link ids mapping to signal group ids).

PR #203, opened 2022-09-01
195 scheduling mapping issue

Saeid-Solei:
# PR Details
## Description

Currently, the signalized_vehicle_scheduler assumes that each entry lane from the intersection_model is connected to a given signal group id and has a signal group id field. But actually, each connection link inside the intersection box have a signal group id field rather than each entry lane.

In signalized_vehicle_scheduler, there is a method that provides the movement_state list for each entry lane based on the entry lane id and its associated signal group ids. This method needs to be updated to provide the movement_state list for each vehicle\entry lane based on the connection lane ids and their associated signal group ids.

## Related Issue

https://github.com/usdot-fhwa-stol/carma-streets/issues/195

## How Has This Been Tested?
Unit Tests

## Types of changes

## Checklist:


streets_utils/streets_vehicle_scheduler/src/signalized_vehicle_scheduler.cpp
```diff
@@ -145,22 +146,37 @@ namespace streets_vehicle_scheduler {
...
     signal_phase_and_timing::movement_state signalized_vehicle_scheduler::find_movement_state_for_lane(const OpenAPI::OAILanelet_info &entry_lane_info) const {
 
+        // check if all links connected to the entry lane have the same signal ids or not!
+        uint8_t signal_group_id = 0;
+        bool first_link_visited = false;
+        for ( const auto &lane : intersection_info->getLinkLanelets() ) {
+            auto connection_lanelet_ids = entry_lane_info.getConnectingLaneletIds();
+            if ( std::count(connection_lanelet_ids.begin(), connection_lanelet_ids.end(), lane.getId()) ) {
+                if ( !lane.getSignalGroupId() ) {
+                    throw scheduling_exception("The link lanelet does not have group_id!");
```
paulbourelly999, line 156: Please make these exceptions clearer since they do not indicate and an illegal MAP. They just violate one of our simplifying assumptions. Please state the simplifying assumption. So something more like:

This entry lanes has not connection lanes 
or 
This entry lane has multiple connection lanes with different signal group ids. The signalized vehicle scheduler is only capable of understanding intersection where all connection lanes from a single entry lane share a signal group id.

streets_utils/streets_vehicle_scheduler/src/signalized_vehicle_scheduler.cpp
```diff
@@ -145,22 +146,37 @@ namespace streets_vehicle_scheduler {
 
     signal_phase_and_timing::movement_state signalized_vehicle_scheduler::find_movement_state_for_lane(const OpenAPI::OAILanelet_info &entry_lane_info) const {
```
paulbourelly999, line 147: Update the documentation for this method

streets_utils/streets_vehicle_scheduler/src/signalized_vehicle_scheduler.cpp
```diff
@@ -145,22 +146,37 @@ namespace streets_vehicle_scheduler {
...
 
+        // check if all links connected to the entry lane have the same signal ids or not!
+        uint8_t signal_group_id = 0;
+        bool first_link_visited = false;
+        for ( const auto &lane : intersection_info->getLinkLanelets() ) {
+            auto connection_lanelet_ids = entry_lane_info.getConnectingLaneletIds();
+            if ( std::count(connection_lanelet_ids.begin(), connection_lanelet_ids.end(), lane.getId()) ) {
```
paulbourelly999, line 154: What is this if statement meant to do. Has this been locally tested against a MAP?

streets_utils/streets_vehicle_scheduler/src/signalized_vehicle_scheduler.cpp
```diff
@@ -145,22 +146,37 @@ namespace streets_vehicle_scheduler {
 
     signal_phase_and_timing::movement_state signalized_vehicle_scheduler::find_movement_state_for_lane(const OpenAPI::OAILanelet_info &entry_lane_info) const {
 
+        // check if all links connected to the entry lane have the same signal ids or not!
+        uint8_t signal_group_id = 0;
+        bool first_link_visited = false;
+        for ( const auto &lane : intersection_info->getLinkLanelets() ) {
```
paulbourelly999, line 152: Why are we looping through all connection lanes?

streets_utils/streets_vehicle_scheduler/src/signalized_vehicle_scheduler.cpp
Saeid-Solei: The first throw happens when a link lanelet does not have a group id. This is a MAP error I believe.
The second throw happens when the UC3 assumption (the assumption of all link lanelets from a signle entry lane share a signal_group_id) is not satisfied. I have added another sentence to explain this as you suggested.

streets_utils/streets_vehicle_scheduler/src/signalized_vehicle_scheduler.cpp
Saeid-Solei: Done!

streets_utils/streets_vehicle_scheduler/src/signalized_vehicle_scheduler.cpp
Saeid-Solei: If the connection link "lane" is included in the list of connection links that are connected to the entry lane.
And yes, no unit test fails.

paulbourelly999: Leave a comment to describe that you are confirming the lane id exists in the vector of connection_lanelet_ids

streets_utils/streets_vehicle_scheduler/include/signalized_vehicle_scheduler.h
```diff
@@ -165,7 +165,10 @@ namespace streets_vehicle_scheduler {
              */
             uint64_t estimate_clearance_time( const streets_vehicles::vehicle &veh ) const;
             /**
-             * @brief find the movement state info from the modified spat for a given entry lane.  
+             * @brief find the movement state info from the modified spat for vehicles from a given entry lane.
+             * Note: The signalized_vehicle_scheduler is only capable of understanding intersection where all connection lanes 
+             * from a single entry lane share a signal_group_id. Therefore, vehicles from an entry lane with different directions
+             * at the intersection box shall be able to receive protected green at the same time.
```
paulbourelly999, line 171: We haven't been great at documenting this but add a @throws descriptions of when this method will throw and exception.

paulbourelly999 (changes requested):
Make the suggested changes

streets_utils/streets_vehicle_scheduler/src/signalized_vehicle_scheduler.cpp
adev4a, line 154: If we just want to check if that specific Id exists in the list, why not just use std::find? count is trying to get the number of occurrences, so it will go through the complete list before stopping, std::find will stop as soon as the element is found.

streets_utils/streets_vehicle_scheduler/src/signalized_vehicle_scheduler.cpp
```diff
@@ -145,22 +146,37 @@ namespace streets_vehicle_scheduler {
...
+            auto connection_lanelet_ids = entry_lane_info.getConnectingLaneletIds();
```
adev4a, line 153: connection_lanelet_ids remains unchanged in the for loop, it should be created outside of it

streets_utils/streets_vehicle_scheduler/src/signalized_vehicle_scheduler.cpp
```diff
@@ -145,22 +146,37 @@ namespace streets_vehicle_scheduler {
...
         else {
-            if ( spat_ptr ) {
-                throw scheduling_exception("SPaT is not found!");
-            }
-            else {
-                throw scheduling_exception("the lanelet does not have group_id!");
-            }
+            throw scheduling_exception("SPaT is not found!");
         }
         return move_state;
```
adev4a, line 181: Looks like if we reach here, move_state is unassigned, should we be returning an uninitialized variable? Or throw an exception that movement_state with the required signal_group_id wasn't found

streets_utils/streets_vehicle_scheduler/src/signalized_vehicle_scheduler.cpp
Saeid-Solei: Good point. Done!

streets_utils/streets_vehicle_scheduler/src/signalized_vehicle_scheduler.cpp
Saeid-Solei: Corrected.

streets_utils/streets_vehicle_scheduler/src/signalized_vehicle_scheduler.cpp
Saeid-Solei: You are right. One throw is missing and I have added it. 
I think it will not reach to this last return. But I kept it here to get rid of the code smell. Essentially, this function with either return a valid move_state or throw an exception.

streets_utils/streets_vehicle_scheduler/include/signalized_vehicle_scheduler.h
Saeid-Solei: Done!

streets_utils/streets_vehicle_scheduler/src/signalized_vehicle_scheduler.cpp
Saeid-Solei: Done!

streets_utils/streets_vehicle_scheduler/src/signalized_vehicle_scheduler.cpp
adev4a: In that case a better way to handle the code smell is removing the return from the loop and replacing it with a break. That way your final return is valid

adev4a (approved):

PR #205, opened 2022-09-08
CircleCI: pass gh-pages branch

codygarver:
# PR Details
## Description

Because gh-pages is force pushes the content of the html dir, it's required to add a minimal CircleCI config in order to get the branch to pass as opposed to merely skipping it.

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -33,4 +33,3 @@ jobs:
       with:
         github_token: ${{ secrets.GITHUB_TOKEN }}
         publish_dir: ./html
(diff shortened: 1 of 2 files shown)


codygarver:
I think skipping the build is the right thing to do but that still shows Workflow Failure and missing config.yml complaints on the Circle CI homepage for carma-streets.

If we want green every time, this is the way.

I'm open to suggestions for an even more minimal config.yml but this is the fastest one I could think of, it seems if you don't use docker it takes even longer to run.

SaikrishnaBairamoni (approved):

Issue #215, opened 2022-09-19
TSC Service Exception Removal


Saeid-Solei:
### Types of Issue

- [x] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

TSC Service throws an exception if the Traffic Signal Controller's ring-barrier design does not include all defined phases. However, it is possible to have ring-barrier design that does not include all phases. An example would be a 1-ring design with four through phases for a single-lane four-way signalized intersection.

### Carma streets version where this issue was discovered

4.2.0

### Expected behavior
TSC Service continues its process if the ring-barrier design does not include all defined phases.

### Actual behavior

TSC Service throws an exception if the ring-barrier design does not include all defined phases.

### Steps to reproduce the actual behavior

1. Deploy CARMA-Streets.
2. Set configuration for UC3 scheduling logic.
3. Configure the Traffic Signal Controller's ring-barrier design to include a subset of defined phases.
4. Restart the TSC Service.

PR #210, opened 2022-09-16
228-control tsc state

adev4a:
# PR Details
This PR adds a control_tsc_state class to construct snmp commands to set HOLD and OMIT calls on the traffic signal controller.
The commands are added to a queue and run according to the time specified in the desired phase plan.

## Related Issue
https://github.com/usdot-fhwa-stol/carma-streets/issues/228

## How Has This Been Tested?
Unit Tested and Integration Tested.

## Types of changes

- [ ] Defect fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that cause existing functionality to change)

## Checklist:

- [ ] I have added any new packages to the sonar-scanner.properties file
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
[CARMA Contributing Guide](https://github.com/usdot-fhwa-stol/carma-platform/blob/develop/Contributing.md) 
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #229, opened 2022-10-04
fix dual ring assumption

adev4a:
# PR Details
## Description
Fixes issue in TSC service where using with a single ring controller led to segfaults.

## Related Issue
https://github.com/usdot-fhwa-stol/carma-streets/issues/215

## Types of changes

## Checklist:

### DIFF
--- a/tsc_client_service/include/monitor_tsc_state.h
+++ b/tsc_client_service/include/monitor_tsc_state.h
@@ -233,13 +233,13 @@ namespace traffic_signal_controller_service
              */
             int get_phase_number(const int signal_group_id);
             /**
-             * @brief Get the signal group id using phase number
+             * @brief Get the signal group id using phase number.Only for vehicle phases
              * 
              * @param phase_number 
              * @return int
              * @throws monitor_states_exception if phase number is less than 1. 
              */
-            int get_signal_group_id(const int phase_number);
+            int get_vehicle_signal_group_id(const int phase_number);
 
             /**
              * @brief Initialize tsc_state by making SNMP calls to TSC for phase sequence and timing information.
--- a/tsc_client_service/include/monitor_tsc_state.h
+++ b/tsc_client_service/include/monitor_tsc_state.h
@@ -103,7 +103,7 @@ namespace traffic_signal_controller_service
             std::vector<std::vector<int>> get_active_ring_sequences(int max_rings, std::unordered_map<int,int>& vehicle_phase_2signalgroup_map, int sequence = 1) const;
 
             void map_phase_and_signalgroup(const std::vector<std::vector<int>>& active_ring_sequences, std::unordered_map<int,int>& vehicle_phase_2signalgroup_map, 
-                                                                                            std::unordered_map<int, int>& signal_group_2vehiclephase_map);
+                                                                                            std::unordered_map<int, int>& signal_group_2vehiclephase_map) const;
 
             /** 
              * @brief Method for getting maximum channels for the traffic signal controller
--- a/tsc_client_service/include/monitor_tsc_state.h
+++ b/tsc_client_service/include/monitor_tsc_state.h
@@ -43,16 +43,16 @@ namespace traffic_signal_controller_service
             std::shared_ptr<snmp_client> snmp_client_worker_;
 
             /* Mapping between signal group ids(key) and their states(value) defined as a signal group state struct for all active vehicle phases*/
-            std::unordered_map<int, signal_group_state> signal_group_state_map_;
+            std::unordered_map<int, signal_group_state> signal_group_2tsc_state_map_;
             
             /* Map between signal group ids(key) and phase numbers(value) for all active vehicle phases in the Traffic Signal Controller*/
-            std::unordered_map<int, int> signal_group_phase_map_;
+            std::unordered_map<int, int> signal_group_2vehiclephase_map_;
 
             /* Map between phase numbers(key) and signal group ids(value) for all active vehicle phases in the Traffic Signal Controller*/
-            std::unordered_map<int,int> vehicle_phase_num_map_;
+            std::unordered_map<int,int> vehicle_phase_2signalgroup_map_;
 
             /* Map between phase numbers(key) and signal group ids(value) for all active pedestrian phases in the Traffic Signal Controller*/
-            std::unordered_map<int,int> ped_phase_num_map_;
+            std::unordered_map<int,int> ped_phase_2signalgroup_map_;
 
             /* The sequence of vehicle phases in ring 1 of TSC*/
             std::vector<int> phase_seq_ring1_;
@@ -82,23 +82,40 @@ namespace traffic_signal_controller_service
 
             /** @brief Creates vectors of channels associated with a vehicle phase and a pedestrian. Ignores overlap, ped Overlap, queueJump and other (types defined in NTCIP1202 v03)
             **  @param max_channels The maximum number of channels in the traffic signal controller.
-            **  @param vehicle_phase_channels a vector of active vehicle phases associated with a channel
-            **  @param ped_phase_channels a vector of active pedestrian phases associated with a channel
+            **  @param vehicle_channels a vector of active vehicle phases associated with a channel
+            **  @param ped_channels a vector of active pedestrian phases associated with a channel
             **/
-            void get_phase_channels(int max_channels, std::vector<int>& vehicle_phase_channels, std::vector<int>& ped_phase_channels) const;
+            void get_channels(int max_channels, std::vector<int>& vehicle_channels, std::vector<int>& ped_channels) const;
 
-            /** @brief Constructs a map between phase number and signal group ids
-            ** @param phase_channels a vector of channel numbers in the traffic signal controller associated with a phase
-            ** @param is_source_vehicle_channel a boolean that indicates whether the phase_channels arg is for channels associated with vehicle phases. 
-            ** In case they are not its assumed that they are pedestrian phases
-            * According to the NTCIP 1202 v03 documentation signal group ids in the SPAT message are the channel numbers in the TSC
-            * **/
-            void map_phase_and_signalgroup(const std::vector<int>& phase_channels, bool is_source_vehicle_channel);
+            /** @brief Creates vectors of phases associated with a vehicle phase and a pedestrian.
+            **  @param channels a map of known channels for vehicle or pedestrian to the corresponding phases
+            **  @return a vector of phases associated with the channels
+            **/
+            std::unordered_map<int,int> get_phases_associated_with_channel(std::vector<int>& channels);
+
+            /** 
+             * @brief Method for getting active rings in sequence (default sequence = 1)
+             * @param max_rings Maximum number of rings in the traffic signal controller
+             * @param vehicle_phase_2signalgroup_map a map for all vehicle phases and signal groups associated with them
+             * @param sequence The sequence number to check for active rings, defaults to 1
+             * @return List of active rings in the traffic signal controller
+            **/
+            std::vector<std::vector<int>> get_active_ring_sequences(int max_rings, std::unordered_map<int,int>& vehicle_phase_2signalgroup_map, int sequence = 1);
+
+            void map_phase_and_signalgroup(const std::vector<std::vector<int>>& active_ring_sequences, std::unordered_map<int,int>& vehicle_phase_2signalgroup_map, 
+                                                                                            std::unordered_map<int, int> signal_group_2vehiclephase_map);
+
             /** 
              * @brief Method for getting maximum channels for the traffic signal controller
              * @return number of maximum channels in the traffic signal controller
             **/
             int get_max_channels() const;
+
+            /** 
+            ** @brief Method for getting maximum rings for the traffic signal controller
+            ** @return number of maximum rings in the traffic signal controller
+            **/
+            int get_max_rings() const;
 
             /** @brief Get minimum green time for a phase
             ** @param phase_num The phase for which the min green needs to be requested
@@ -132,15 +149,10 @@ namespace traffic_signal_controller_service
 
             /** @brief Get a sequence of phases following the given phase
             ** @param phase_num The phase for which the sequence needs to be obtained
+            ** @param active_ring_sequences The sequence of phases in active rings in the traffic signal controller
             ** @return a vector as a sequence of phases starting from the given phase
             * **/
-            std::vector<int> get_following_phases(int phase_num);
-
-            /** @brief Get a sequence of phases in the given ring
-            ** @param ring_num The phase for which the sequence needs to be obtained
-            ** @return a vector as a sequence of phases in the ring
-            * **/
-            std::vector<int> phase_seq(int ring_num);
+            std::vector<int> get_following_phases(int phase_num, const std::vector<std::vector<int>>& active_ring_sequences);
 
             /** @brief The concurrent phases that the given phase can be green along with
             ** @param ring_num The phase for which the concurrent phases needs to be obtained
@@ -168,9 +180,9 @@ namespace traffic_signal_controller_service
             * **/
             signal_phase_and_timing::movement_event get_following_event(const signal_phase_and_timing::movement_event& current_event,
                                                                  uint64_t current_event_end_time, const signal_group_state& phase_state) const;
-
+            
             //Add Friend Test to share private members
-            FRIEND_TEST(traffic_signal_controller_service, test_get_following_movement_events);                                                              
+            FRIEND_TEST(monitor_state_test, test_get_following_movement_events);                                                              
 
         public:
             /** 
@@ -184,6 +196,12 @@ namespace traffic_signal_controller_service
              * @return std::unordered_map<int, signal_group_state>& 
              */
             std::unordered_map<int, signal_group_state>& get_signal_group_state_map();
+            /**
+             * @brief Returns a map of signal group ids to pedestrian phases
+             * 
+             * @return a map of signal group ids to pedestrian phases
+             */
+            std::unordered_map<int, int>& get_signal_group_to_ped_phase_map();
             /** 
              * @brief Returns a map of pedestrian phases to signal group ids
              * @return a map of pedestrian phases to signal group ids 
--- a/tsc_client_service/include/monitor_tsc_state.h
+++ b/tsc_client_service/include/monitor_tsc_state.h
@@ -77,19 +77,22 @@ namespace traffic_signal_controller_service
              * @brief Creates tsc_configuration_state object pointer which is used to forward configuration information required
              * to understand ring and barrier information on the traffic signal controller.
              * This includes signal_group_id, yellow_change duration, red_clearance and concurrent signal groups
+             * @return shared ptr to tsc_config state
              */
-            void define_tsc_config_state();
+            std::shared_ptr<streets_tsc_configuration::tsc_configuration_state> define_tsc_config_state();
 
             /** @brief Creates vectors of channels (or signal group id) associated with a vehicle phase and a pedestrian. Ignores overlap, ped Overlap, queueJump and other (types defined in NTCIP1202 v03)
+             *  Expects empty input arguments which are filled in by reference.
              *  @param max_channels The maximum number of channels in the traffic signal controller.
              *  @param vehicle_channels a vector of active vehicle phases associated with a channel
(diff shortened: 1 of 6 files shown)


adev4a:
Merge after https://github.com/usdot-fhwa-stol/carma-streets/pull/210/files
Unit tests update in monitor_tsc_state needed here

tsc_client_service/src/monitor_tsc_state.cpp
```diff
@@ -71,13 +78,14 @@ namespace traffic_signal_controller_service
...
-                SPDLOG_DEBUG("Concurrent Phases:");
+                SPDLOG_DEBUG("Concurrent Signal Groups:");
                 for(auto signal_group : state.concurrent_signal_groups){
                     SPDLOG_DEBUG("{0}", signal_group);
                 }  
             }
             return true;
-        }catch ( const snmp_client_exception &e ) {
+
+            }catch ( const snmp_client_exception &e ) {
```
Saeid-Solei, line 88: Wasn't the indentation correct?

tsc_client_service/include/monitor_tsc_state.h
```diff
@@ -82,24 +82,41 @@ namespace traffic_signal_controller_service
...
-            **  @param vehicle_phase_channels a vector of active vehicle phases associated with a channel
-            **  @param ped_phase_channels a vector of active pedestrian phases associated with a channel
+            **  @param vehicle_channels a vector of active vehicle phases associated with a channel
+            **  @param ped_channels a vector of active pedestrian phases associated with a channel
             **/
-            void get_phase_channels(int max_channels, std::vector<int>& vehicle_phase_channels, std::vector<int>& ped_phase_channels) const;
+            void get_channels(int max_channels, std::vector<int>& vehicle_channels, std::vector<int>& ped_channels) const;
+
+            /** @brief Creates vectors of phases associated with a vehicle phase and a pedestrian.
+            **  @param channels a map of known channels for vehicle or pedestrian to the corresponding phases
```
Saeid-Solei, line 91: Switch param and return!

tsc_client_service/src/monitor_tsc_state.cpp
```diff
@@ -454,9 +504,10 @@ namespace traffic_signal_controller_service
 
         //extract phase numbers from strings
         for(auto con_phase :  concurrent_phase_data.val_string)
-        {   
-            concurrent_signal_groups.push_back(signal_group_phase_map_[int(con_phase)]);
-            
+        {   auto concurrent_phase = int(con_phase);
```
Saeid-Solei, line 507: Is this here by mistake?

tsc_client_service/src/monitor_tsc_state.cpp
```diff
@@ -469,19 +520,31 @@ namespace traffic_signal_controller_service
...
     {
-        return signal_group_state_map_;
+        return signal_group_2tsc_state_map_;
     }        
 
+    std::unordered_map<int, int> tsc_state::get_signal_group_to_ped_phase_map()
+    {
+        std::unordered_map<int,int> signal_group_to_ped_phase;
+
+        for(auto it = ped_phase_2signalgroup_map_.begin(); it != ped_phase_2signalgroup_map_.end();){
```
Saeid-Solei, line 540: Should we add if signal_group_to_ped_phase is not empty and throw exception if it is?

tsc_client_service/include/monitor_tsc_state.h
```diff
@@ -82,24 +82,41 @@ namespace traffic_signal_controller_service
...
             **  @param max_channels The maximum number of channels in the traffic signal controller.
-            **  @param vehicle_phase_channels a vector of active vehicle phases associated with a channel
-            **  @param ped_phase_channels a vector of active pedestrian phases associated with a channel
+            **  @param vehicle_channels a vector of active vehicle phases associated with a channel
+            **  @param ped_channels a vector of active pedestrian phases associated with a channel
             **/
-            void get_phase_channels(int max_channels, std::vector<int>& vehicle_phase_channels, std::vector<int>& ped_phase_channels) const;
+            void get_channels(int max_channels, std::vector<int>& vehicle_channels, std::vector<int>& ped_channels) const;
+
+            /** @brief Creates vectors of phases associated with a vehicle phase and a pedestrian.
```
paulbourelly999, line 90: This comment seams to be for the wrong method. This method does not return a vector

tsc_client_service/include/monitor_tsc_state.h
```diff
@@ -82,24 +82,41 @@ namespace traffic_signal_controller_service
 
             /** @brief Creates vectors of channels associated with a vehicle phase and a pedestrian. Ignores overlap, ped Overlap, queueJump and other (types defined in NTCIP1202 v03)
```
paulbourelly999, line 83: Please mention that channel number is synonymous with signal group id right?

tsc_client_service/include/monitor_tsc_state.h
```diff
@@ -82,24 +82,41 @@ namespace traffic_signal_controller_service
...
+            void get_channels(int max_channels, std::vector<int>& vehicle_channels, std::vector<int>& ped_channels) const;
+
+            /** @brief Creates vectors of phases associated with a vehicle phase and a pedestrian.
+            **  @param channels a map of known channels for vehicle or pedestrian to the corresponding phases
+            **  @return a vector of phases associated with the channels
+            **/
+            std::unordered_map<int,int> get_phases_associated_with_channel(const std::vector<int>& channels) const;
+
+            /** 
+             * @brief Method for getting active rings in sequence (default sequence = 1)
```
paulbourelly999, line 97: Define activate. Maybe "populated" is more descriptive. Also give information about sequence and why we default to 1. "Sequence table contains multiple sequence plans, consisting of the phase sequence in each ring available in the traffic signal controller. We assume the active sequence plan is always sequence number 1." Something like that

tsc_client_service/include/monitor_tsc_state.h
```diff
@@ -82,24 +82,41 @@ namespace traffic_signal_controller_service
...
+             * @param max_rings Maximum number of rings in the traffic signal controller
+             * @param vehicle_phase_2signalgroup_map a map for all vehicle phases and signal groups associated with them
+             * @param sequence The sequence number to check for active rings, defaults to 1
+             * @return List of active rings in the traffic signal controller
+            **/
+            std::vector<std::vector<int>> get_active_ring_sequences(int max_rings, std::unordered_map<int,int>& vehicle_phase_2signalgroup_map, int sequence = 1) const;
+
+            void map_phase_and_signalgroup(const std::vector<std::vector<int>>& active_ring_sequences, std::unordered_map<int,int>& vehicle_phase_2signalgroup_map,
```
paulbourelly999, line 105: There is no comment for this method.

tsc_client_service/include/monitor_tsc_state.h
```diff
@@ -82,24 +82,41 @@ namespace traffic_signal_controller_service
...
-            * **/
-            void map_phase_and_signalgroup(const std::vector<int>& phase_channels, bool is_source_vehicle_channel);
             /** 
              * @brief Method for getting maximum channels for the traffic signal controller
              * @return number of maximum channels in the traffic signal controller
             **/
             int get_max_channels() const;
 
+            /** 
+            ** @brief Method for getting maximum rings for the traffic signal controller
```
paulbourelly999, line 115: Why do your javadoc comments have two **. Please just use the regular one like above.

tsc_client_service/src/monitor_tsc_state.cpp
```diff
@@ -206,50 +214,87 @@ namespace traffic_signal_controller_service
...
+            {
+                if(std::find(ring.begin(), ring.end(), phase) !=ring.end()){
+                    found_phase = true;
+                    break;
+                }    
+            }
+            
+            if(!found_phase)
+            {
+                // if phase not found in active sequence, remove entry
```
paulbourelly999, line 257: So this will remove any phases not present in active ring sequences?

tsc_client_service/test/test_monitor_state.cpp
```diff
@@ -54,6 +54,13 @@ namespace traffic_signal_controller_service
             ON_CALL(*unique_client, process_snmp_request(channel_control_oid, _ , _) )
                 .WillByDefault(testing::DoAll(testing::SetArgReferee<2>(channel_control_resp), testing::Return(true)));
 
+            // Define Max Rings
+            snmp_response_obj max_rings_in_tsc;
+            max_rings_in_tsc.val_int = 4;
+            max_rings_in_tsc.type = snmp_response_obj::response_type::INTEGER;
+            ON_CALL(*unique_client, process_snmp_request(ntcip_oids::MAX_RINGS, _ , _) )
```
paulbourelly999, line 61: I think we spoke about this but if we could eventually move these calls from "ON_CALL" to "EXPECT_CALL" I think that would be good. Not sure high priority

tsc_client_service/test/test_monitor_desired_phase_plan.cpp
```diff
@@ -152,7 +152,7 @@ namespace traffic_signal_controller_service
             phase_1_state.yellow_duration = 1000;
             phase_1_state.red_clearance = 1000;
             phase_1_state.red_duration = 10000;
-            // tsc_state_ptr->signal_group_state_map_.insert({1, phase_1_state});
+            // tsc_state_ptr->signal_group_2tsc_state_map_.insert({1, phase_1_state});
```
paulbourelly999, line 155: Remove any commented out code please.

tsc_client_service/src/monitor_tsc_state.cpp
paulbourelly999: No it is a local variable that he creaes in the line before, It will always be empty.

tsc_client_service/include/snmp_client.h
```diff
@@ -34,7 +34,7 @@ struct snmp_response_obj
 
     //snmp response values can be any asn.1 supported types. 
     //Integer and string values can be processed here
-    int64_t val_int = 0;
+    int64_t val_int;
```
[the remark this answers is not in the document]
adev4a: I removed this so that an empty return value doesn't default to 0. I tested with this change on the TSC and it doesn't seem to break anything. 
This is following an issue we were seeing with concurrent signal_groups being assigned a value of 0 if there was response was empty.

tsc_client_service/src/monitor_tsc_state.cpp
adev4a: Yes, that's correct. We are only tracking active vehicle phases

tsc_client_service/include/monitor_tsc_state.h
adev4a: I double checked this, the param and return seem to be indicating the correct definition. Can you explain what you meant?

tsc_client_service/src/monitor_tsc_state.cpp
adev4a: If the signal_group_to_ped_phase map is empty, it wont enter this loop. Throwing an exception here would mean the TSC service needs pedestrian phases to exist on the traffic signal controller, which isn't true right?

The need for this map is only while processing spat messages and if there are no pedestrian phases in the controller, we wont get any spat associated with them.

paulbourelly999: I think the way you have it currently is correct anish unless I am missing something. signal_grou_to_ped_phase will alway be empty which is the intent since you are creating a new MAP, and like you said, it wont enter the loop if ped_phase_2signal_group_map is empty.

tsc_client_service/include/snmp_client.h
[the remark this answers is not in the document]
paulbourelly999: I think that is the correct behavior though. It should be zero if the response is empty since 0 is an invalid phase_number/signal group. Not setting a default value I believe will cause this to have a random value based on its memory address at initialization. (https://www.learncpp.com/cpp-tutorial/uninitialized-variables-and-undefined-behavior/)

tsc_client_service/include/monitor_tsc_state.h
```diff
@@ -43,16 +43,16 @@ namespace traffic_signal_controller_service
...
-            std::unordered_map<int, signal_group_state> signal_group_state_map_;
+            std::unordered_map<int, signal_group_state> signal_group_2tsc_state_map_;
             
             /* Map between signal group ids(key) and phase numbers(value) for all active vehicle phases in the Traffic Signal Controller*/
-            std::unordered_map<int, int> signal_group_phase_map_;
+            std::unordered_map<int, int> signal_group_2vehiclephase_map_;
 
             /* Map between phase numbers(key) and signal group ids(value) for all active vehicle phases in the Traffic Signal Controller*/
-            std::unordered_map<int,int> vehicle_phase_num_map_;
+            std::unordered_map<int,int> vehicle_phase_2signalgroup_map_;
```
dan-du-car, line 52: What is vehicle phase?

dan-du-car: can we use one of these two maps: std::unordered_map<int,int> vehicle_phase_2signalgroup_map_; std::unordered_map<int, int> signal_group_2vehiclephase_map_ if we want to find by key or by value?

dan-du-car:
How does this code update fix the segmentation fault?

tsc_client_service/include/monitor_tsc_state.h
adev4a: The traffic signal controller has a total of 16 possible phases, out of 8 are vehicle phases and the other 8 are pedestrian.
Here we are looking at the active vehicle phases

adev4a:
> How does this code update fix the segmentation fault?

The seg fault was due to an assumption made in the implementation that the controller would always be dual ring. This fix removes that assumption and has been tested with a single ring setup

tsc_client_service/include/snmp_client.h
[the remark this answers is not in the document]
adev4a: Yeah that makes sense. I've added this back

tsc_client_service/src/monitor_tsc_state.cpp
```diff
@@ -11,51 +11,58 @@ namespace traffic_signal_controller_service
...
-            // Get phase sequences for ring 1 and ring 2
-            phase_seq_ring1_ = phase_seq(1);
-            phase_seq_ring2_ = phase_seq(2);
-
-            // Define state of each signal group
-            for (const auto& signal_group : signal_group_phase_map_)
+            // Loop through all channels and check which are associated with a vehicle phase and pedestrian phase
+            std::vector<int> vehicle_channels; 
+            std::vector<int> ped_channels;
+            get_channels(max_channels_in_tsc, vehicle_channels, ped_channels);
```
dan-du-car, line 19: Is this line updating the vehicle and ped channels? Can we have a more descriptive name for update? or add an line comment about this get_channels function?

tsc_client_service/src/monitor_tsc_state.cpp
```diff
@@ -11,51 +11,58 @@ namespace traffic_signal_controller_service
...
+            
+            // Get Phases associated with vehicle channels and pedestrian channels
+            vehicle_phase_2signalgroup_map_ = get_phases_associated_with_channel(vehicle_channels);
+            ped_phase_2signalgroup_map_ = get_phases_associated_with_channel(ped_channels);
+            
+            int max_rings_in_tsc = get_max_rings();
+            // Get vehicle sequence for active phases. By default in sequence 1
+            std::vector<std::vector<int>> active_ring_sequences = get_active_ring_sequences(max_rings_in_tsc, vehicle_phase_2signalgroup_map_);
+            // Remove vehicle phases not in map
```
dan-du-car, line 28: There are two maps? Remove vehicle phases if not in which map? and maybe the comments should also explain why do we need active ring?

tsc_client_service/src/monitor_tsc_state.cpp
```diff
@@ -11,51 +11,58 @@ namespace traffic_signal_controller_service
...
             // Loop through states once other state parameters are defined to get the red duration
-            for(auto& [signalgroup_id, state] : signal_group_state_map_)
+            for(auto& [signalgroup_id, state] : signal_group_2tsc_state_map_)
             {
                 
                 state.red_duration = get_red_duration(state.phase_num);
             }
 
             // Define tsc config state
             define_tsc_config_state();
```
dan-du-car, line 57: why do we need this?

tsc_client_service/src/monitor_tsc_state.cpp
```diff
@@ -208,50 +215,87 @@ namespace traffic_signal_controller_service
...
-    int tsc_state::get_max_channels() const {
+    std::unordered_map<int,int> tsc_state::get_phases_associated_with_channel(const std::vector<int>& channels) const {
+        
+        std::unordered_map<int,int> phases_to_signal_group;
         request_type request_type = request_type::GET;
-        snmp_response_obj max_channels_in_tsc;
-        max_channels_in_tsc.type = snmp_response_obj::response_type::INTEGER;
 
-        if(!snmp_client_worker_->process_snmp_request(ntcip_oids::MAX_CHANNELS, request_type, max_channels_in_tsc))
+        for(int channel : channels)
```
dan-du-car, line 275: So this channel number is signal group number?

tsc_client_service/include/monitor_tsc_state.h
paulbourelly999: This comment is still wrong

tsc_client_service/src/monitor_tsc_state.cpp
adev4a: We're just updating the same map to only include phases in the ring sequence (active phases). And sure, i can add a more descriptive comment

tsc_client_service/src/monitor_tsc_state.cpp
adev4a: That is needed to create a message which is sent to the SO service. I can add a comment

tsc_client_service/src/monitor_tsc_state.cpp
adev4a: Makes sense, I can change this to only refer to signal group id

tsc_client_service/include/monitor_tsc_state.h
```diff
@@ -80,102 +80,116 @@ namespace traffic_signal_controller_service
              */
             void define_tsc_config_state();
 
-            /** @brief Creates vectors of channels associated with a vehicle phase and a pedestrian. Ignores overlap, ped Overlap, queueJump and other (types defined in NTCIP1202 v03)
-            **  @param max_channels The maximum number of channels in the traffic signal controller.
-            **  @param vehicle_phase_channels a vector of active vehicle phases associated with a channel
-            **  @param ped_phase_channels a vector of active pedestrian phases associated with a channel
+            /** @brief Creates vectors of channels (or signal group id) associated with a vehicle phase and a pedestrian. Ignores overlap, ped Overlap, queueJump and other (types defined in NTCIP1202 v03)
```
paulbourelly999, line 83: Please indicate that you are returning these by parameter reference. So you expected empty vectors to be passed in and populated inside the method an returned by reference.

tsc_client_service/include/monitor_tsc_state.h
```diff
@@ -80,102 +80,116 @@ namespace traffic_signal_controller_service
...
+             * @param vehicle_phase_2signalgroup_map a map for all vehicle phases and signal groups associated with them
+             * @param sequence The sequence number to check for active rings, defaults to 1
+             * @return List of active rings in the traffic signal controller
+            **/
+            std::vector<std::vector<int>> get_active_ring_sequences(int max_rings, std::unordered_map<int,int>& vehicle_phase_2signalgroup_map, int sequence = 1) const;
+
+            /** 
+             * @brief Method for mapping vehicle phases and signal groups
+             * @param active_ring_sequences A list of phases from the traffic signal controller in rows corresponding to the ring number
+             * @param vehicle_phase_2signalgroup_map a map for all vehicle phases and signal groups associated with them
```
paulbourelly999, line 110: For any methods where you expect empty data types to be passed in and populated within the method please indicate this in the comment. Also indicate any initial conditions you require (i.e. empty map) and any possible outcomes ( populated map if works , empty map if it fails.)

dan-du-car (dismissed):

paulbourelly999 (approved):
Looks good to me.

Issue #223, opened 2022-09-20
TSC Service segfault 


paulbourelly999:
### Types of Issue

- [ x] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

When attempting to run the **TSC Service** with the **East Intersection TSC**, the **TSC Service** crashes with a segmentation fault. When looking at the logs it seems to indicate that some of the retrieved information about phases seems incorrect. For example the logs indicate that for phase number 8, the concurrent phases are 0, which is an invalid phase number. The East Intersection configuration actually has 4 as a concurrent phase.
### Carma streets version where this issue was discovered

Latest
### Expected behavior
**TSC Service** is able to read phase configuration and phase sequence data without crashing.
### Actual behavior

**TSC Service** crashes after reading and sending phase configuration and sequence data from East Intersection. It also seems to read phase and sequence data incorrectly from the TSC.
### Steps to reproduce the actual behavior

1. Configure **TSC Service** to connect to East Intersection TSC.
2. Configure **Intersection Model** with East Intersection lanelet2 map.
3. Run CARMA-Streets
4. TSC Service crashes

PR #230, opened 2022-10-05
223 fix concurrent phase

paulbourelly999:
# PR Details
## Description

TSC Service was throwing segmentation faults when initializing with TSC configuration where signal group id's were not always equal to phase numbers. This was due to a bug in the implementation using the wrong map (i.e using phase number to signal group map instead of signal group to phase number map) to make SNMP requests. The SNMP request would return 0 which would eventually cause a segmentation fault. 

As a fix, this PR implements helper methods that use the signal group id to phase number and phase number to signal group id maps to avoid having to use these maps directly. 

Also moved Mock SNMP Client from Test Case to TSC Service library so any test case can use it. 

Updated Mock macros calls from `ON_CALL` which does not enforce any assertion on method execution and generates warnings to `EXPECT_CALL` which does enforce method execution assertions. `ON_CALL` should be used only for "uninteresting" method calls where we don't need to know how many times or if at all a method of a mock class is called where as `EXPECT_CALL` can be used to specifically assert how many times a mock class method is called.  
## Related Issue
#223 

## Motivation and Context
Bug fix

## How Has This Been Tested?
Unit Testing

## Types of changes

## Checklist:

- [ x] I have added tests to cover my changes.
- [x ] All new and existing tests passed.


tsc_client_service/test/test_udp_socket_listener.cpp
```diff
@@ -54,9 +54,9 @@ namespace traffic_signal_controller_service
      */
     TEST(udp_socket_listener_test, test_invalid_ip)
     {
-        std::string tsc_ip_bad = "asdhas.asd";
```
paulbourelly999, line 57: Updated because this test was throwing segmentation fault on execution.

Saeid-Solei (approved):

tsc_client_service/include/monitor_tsc_state.h
```diff
@@ -196,6 +197,23 @@ namespace traffic_signal_controller_service
             **/
             const std::unordered_map<int,int>& get_vehicle_phase_map();
 
+            /**
+             * @brief Get the phase number using signal group id.
+             * 
+             * @param signal_group_id 
+             * @return int
+             * @throws monitor_states_exception if signal group id is less than 1.
```
dan-du-car, line 205: Why sg cannot be less than 1

tsc_client_service/src/monitor_tsc_state.cpp
```diff
@@ -320,6 +320,34 @@ namespace traffic_signal_controller_service
...
+        if (signal_group_id >= 1) {
+            if ( signal_group_phase_map_.find(signal_group_id) != signal_group_phase_map_.end() ) {
+                return signal_group_phase_map_[signal_group_id];
+            } 
+            else {
+                throw monitor_states_exception("No phase number found for signal id " + std::to_string(signal_group_id) + "!");
+            }
+        }
+        else {
+            throw monitor_states_exception("Signal group ids less than 1 are invalid!");
```
dan-du-car, line 333: maybe print out the sg id ?

tsc_client_service/src/monitor_tsc_state.cpp
```diff
@@ -320,6 +320,34 @@ namespace traffic_signal_controller_service
...
+        if (phase_number >= 1) {
+            if ( vehicle_phase_num_map_.find(phase_number) != vehicle_phase_num_map_.end() ) {
+                return vehicle_phase_num_map_[phase_number];
+            } 
+            else {
+                throw monitor_states_exception("No signal group id found for phase number " + std::to_string(phase_number) + "!");
+            }
+        }
+        else {
+            throw monitor_states_exception("Phase numbers less than 1 are invalid!");
```
dan-du-car, line 347: maybe print out the phase number ?

tsc_client_service/src/monitor_tsc_state.cpp
```diff
@@ -416,10 +432,9 @@ namespace traffic_signal_controller_service
     std::vector<int> tsc_state::phase_seq(int ring_num)
     {
         std::vector<int> phase_seq;
-        // Read sequence 1 data for first 2 rings
+        // Read sequence 1 data for given ring
```
dan-du-car, line 435: Maybe add detail explanation what sequence 1 data is , and why choose this sequence 1 not other number?

tsc_client_service/test/test_monitor_state.cpp
```diff
@@ -241,5 +242,300 @@ namespace traffic_signal_controller_service
...
+            EXPECT_CALL(*mock_snmp, process_snmp_request(concurrent_phase_oid , request_type , _) ).Times(1).WillRepeatedly(testing::DoAll(
+                SetArgReferee<2>(concurrent_phase_resp), 
+                Return(true)));
+        }
+
+        state.initialize();
+        // phase number
+        // ring 1 ->  | 1 | 2 || 3 | 4 ||
+        // ring 2 ->  | 5 | 6 || 7 | 8 ||
+        // signal group
```
dan-du-car, line 509: What are all the possible SG ids?

tsc_client_service/include/monitor_tsc_state.h
paulbourelly999: SG group of 0 is invalid

tsc_client_service/src/monitor_tsc_state.cpp
```diff
@@ -320,6 +320,34 @@ namespace traffic_signal_controller_service
         
     }
 
+    int tsc_state::get_phase_number(const int signal_group_id ) {
+        if (signal_group_id >= 1) {
+            if ( signal_group_phase_map_.find(signal_group_id) != signal_group_phase_map_.end() ) {
+                return signal_group_phase_map_[signal_group_id];
+            } 
+            else {
```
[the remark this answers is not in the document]
paulbourelly999: No we need every activate phase to have a signal group id or else we can't send configuration data to SO and can't create valid spat message

tsc_client_service/src/monitor_tsc_state.cpp
```diff
@@ -320,6 +320,34 @@ namespace traffic_signal_controller_service
...
+                throw monitor_states_exception("No phase number found for signal id " + std::to_string(signal_group_id) + "!");
+            }
+        }
+        else {
+            throw monitor_states_exception("Signal group ids less than 1 are invalid!");
+        }
+    }
+
+    int tsc_state::get_signal_group_id(const int phase_number ) {
+        if (phase_number >= 1) {
```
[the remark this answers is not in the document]
paulbourelly999: I don't think this makes sense. The caller should never call this with an INVALID  phase number ( < 1).

tsc_client_service/src/monitor_tsc_state.cpp
paulbourelly999: I do not think this is necessary. We should never all this on a signal group less that 1.

tsc_client_service/test/test_monitor_state.cpp
paulbourelly999: From ASN1 file
```
SignalGroupID ::= INTEGER (0..255) 
    -- The value 0 shall be used when the ID is
    -- not available or not known
    -- the value 255 is reserved to indicate a 
    -- permanent green movement state
    -- therefore a simple 8 phase signal controller 
    -- device might use 1..9 as its groupIDs
```

tsc_client_service/src/monitor_tsc_state.cpp
paulbourelly999: Added detail as part of javadoc comment. Please let me know if this is better.

tsc_client_service/src/monitor_tsc_state.cpp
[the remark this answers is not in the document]
paulbourelly999: image

tsc_client_service/include/monitor_tsc_state.h
dan-du-car: Maybe update the comments

tsc_client_service/src/monitor_tsc_state.cpp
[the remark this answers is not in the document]
dan-du-car: add inline comments to indicate the valid value range for this phase number.

tsc_client_service/src/monitor_tsc_state.cpp
dan-du-car: Isn;t this exeption throw when receiving invalid sg? Maybe the exception message can be   throw monitor_states_exception("Phase numbers (=" +sg_id+" less than 1 are invalid!");

PR #239, opened 2022-10-17
233 fix spat

paulbourelly999:
# PR Details
## Description

Fix SPaT JSON serialization/deserialization to account for **TimeMark** max value of 36000. (Serialization) Will convert any epoch time that exceeds this limit to 36000 and print a warning when using the set methods of TimeChangeDetails. (Deserialization)Will throw exception when reading JSON with any value about 36000 for TimeChangeDetails time mark values.

Also added configurable parameter to message_services for lanelet2 map file 
```
TimeMark ::= INTEGER (0..36001)
   -- Tenths of a second in the current or next hour
   -- In units of 1/10th second from UTC time
   -- A range of 0~36000 covers one hour
   -- The values 35991..35999 are used when a leap second occurs
   -- The value 36000 is used to indicate time >3600 seconds
   -- 36001 is to be used when value undefined or unknown
   -- Note that this is NOT expressed in GPS time
   -- or in local time
```
## Types of changes

## Checklist:


paulbourelly999:
Still working on unit test coverage.

dan-du-car (approved):

streets_utils/streets_signal_phase_and_timing/src/models/time_change_details.cpp
```diff
@@ -94,6 +111,21 @@ namespace signal_phase_and_timing {
...
+        //          -- In units of 1/10th second from UTC time
+        //          -- A range of 0~36000 covers one hour
+        //          -- The values 35991..35999 are used when a leap second occurs
+        //          -- The value 36000 is used to indicate time >3600 seconds
+        //          -- 36001 is to be used when value undefined or unknown
+        //          -- Note that this is NOT expressed in GPS time
+        //          -- or in local time
+        if ( tenth_seconds_from_current_hour >= 36000) {
+            SPDLOG_WARN("Approximating {0} as 36000! TimeMark maximum value is 36000!", tenth_seconds_from_current_hour);
+            tenth_seconds_from_current_hour = 36000;
```
adev4a, line 127: Why are we storing this value here and throwing exceptions in this case everywhere else?

tsc_client_service/src/monitor_tsc_state.cpp
```diff
@@ -113,29 +113,31 @@ namespace traffic_signal_controller_service
...
+                next_event.timing.set_start_time(current_event_end_time); 
+                next_event.timing.set_min_end_time(current_event_end_time + phase_state.red_duration);
+                next_event.timing.set_max_end_time(current_event_end_time + phase_state.red_duration);
                 break;
 
             case signal_phase_and_timing::movement_phase_state::stop_and_remain:  //Red
                 // Create next movement - green
                 next_event.event_state = signal_phase_and_timing::movement_phase_state::protected_movement_allowed;
-                next_event.timing.start_time = convert_msepoch_to_hour_tenth_secs(current_event_end_time);
-                next_event.timing.min_end_time = convert_msepoch_to_hour_tenth_secs(current_event_end_time + phase_state.green_duration);
```
adev4a, line 131: Looks like these conversion methods are no longer needed in the monitor_tsc_state class, can we remove them?

streets_utils/streets_signal_phase_and_timing/src/models/time_change_details.cpp
paulbourelly999: The J2737 documentation specifies that it is valid to set any TimeMark that is greater that 36000 to 36000
'''
The value 36000 is used to indicate time >3600 seconds
'''
If the JSON contains a value that is greater 36000 it is an invalid SPaT message so we thrown an exception. Does that make sense.

tsc_client_service/src/monitor_tsc_state.cpp
paulbourelly999: Will remove these. Good catch!

message_services/src/services/vehicle_status_intent_service.cpp
```diff
@@ -121,19 +132,15 @@ namespace message_services
...
                     while (true)
                     {
-                        // Change spdlog from debug to info for printing output in terminal
-                        SPDLOG_DEBUG("Processing the BSM list size: {0}", bsm_w_ptr->get_curr_map().size());
-                        SPDLOG_DEBUG("Processing the MobilityOperation list size: {0}", mo_w_ptr->get_curr_list().size());
-                        SPDLOG_DEBUG("Processing the MobilityPath list size: {0}", mp_w_ptr->get_curr_map().size());
                         if (mo_w_ptr && mo_w_ptr->get_curr_list().size() > 0 && bsm_w_ptr && bsm_w_ptr->get_curr_map().size() > 0 && mp_w_ptr && mp_w_ptr->get_curr_map().size() > 0)
                         {
-                            SPDLOG_DEBUG("Processing the BSM, mobilityOperation and MP from list...");
+                            SPDLOG_TRACE("Processing the BSM, mobilityOperation and MP from list...");
```
dan-du-car, line 137: Why do we prefer to use SPDLOG_TRACE to SPDLOG_DEBUG?

paulbourelly999:
Still trying to add some other bug fixes for Dockerfile and unit test. Please hold off on merging until then

message_services/Dockerfile
```diff
@@ -126,6 +108,25 @@ RUN source /opt/ros/melodic/setup.bash && \
     DEBIAN_FRONTEND=noninteractive apt-get install -y libeigen3-dev && \
     ROS_VERSION=1 LANELET2_EXTENSION_LOGGER_TYPE=1 catkin_make install
 
+COPY ./streets_utils/ /home/carma-streets/streets_utils
```
paulbourelly999, line 111: Moved internal dependencies further back to increase the amount of build layers that can be cached between rebuilds

message_services/src/message_services.cpp
```diff
@@ -3,23 +3,20 @@
...
-    if (s_v.initialize(msg_translate_ptr))
+    if (s_v.initialize())
     {
         s_v.start();
     }
 }
 
 int main(int argc, const char **argv)
 {
-    const std::string OSM_FILE_PATH = "../vector_map.osm";
```
paulbourelly999, line 17: Using configurable parameter instead of hard_coded path

dan-du-car (dismissed):

message_services/src/services/vehicle_status_intent_service.cpp
```diff
@@ -50,26 +62,25 @@ namespace message_services
...
+                        return false;
                     }
                 }
 
                 this->_vsi_producer_worker = client->create_producer(this->bootstrap_server, this->vsi_topic_name);
                 if (!_vsi_producer_worker->init())
                 {
                     SPDLOG_CRITICAL("kafka producer (_vsi_producer_worker) initialize error");
-                    exit(-1);
+                    return false;
```
paulbourelly999, line 73: Return false since initialize method returns a bool instead of exit.

message_services/test/test_message_lanelet2_translation.cpp
```diff
@@ -6,386 +6,314 @@
 TEST(test_message_lanelet2_translation, read_lanelet2_map)
 {
```
paulbourelly999, line 7: Updated unit tests to use town 1 sample OSM map.

message_services/test/test_message_lanelet2_translation.cpp
```diff
@@ -6,386 +6,314 @@
...
-    offset.offset_z = 0;
-    trajectory.offsets.clear();
-    trajectory.offsets.push_back(offset);
-
-    ASSERT_NEAR(-66.3387, clt.ecef_2_map_point(location.ecef_x + offset.offset_x, location.ecef_y + offset.offset_y, location.ecef_z + offset.offset_z).x(), 0.1);
-    ASSERT_NEAR(327.636, clt.ecef_2_map_point(location.ecef_x + offset.offset_x, location.ecef_y + offset.offset_y, location.ecef_z + offset.offset_z).y(), 0.1);
-    ASSERT_NEAR(72, clt.ecef_2_map_point(location.ecef_x + offset.offset_x, location.ecef_y + offset.offset_y, location.ecef_z + offset.offset_z).z(), 0.1);
+    ASSERT_EQ(155, clt.get_cur_lanelet_by_loc_and_direction(48.9978481, 8.0026023, 0, "left", trajectory).id());
+
+    // auto subj_lanelet = clt.get_cur_lanelet_by_point_and_direction(clt.ecef_2_map_point(location.ecef_x,location.ecef_y,location.ecef_z), turn_direction,trajectory);
```
paulbourelly999, line 187: Only updated test cases necessary to pass sonar scanner test coverage.

streets_utils/streets_signal_phase_and_timing/src/models/time_change_details.cpp
```diff
@@ -7,68 +7,85 @@ namespace signal_phase_and_timing {
...
+        if (likely_time <= 36000) {
             detail.AddMember("likely_time", likely_time, allocator);
         }
+        else {
+            SPDLOG_WARN("Skipping invalid value {0} for likely_time! Valid range (0-36000).", likely_time);
+        }
         detail.AddMember("confidence", confidence, allocator);
         // value represents unknow see J2735 spec  
-        if (next_time != 36001) {
+        if (next_time <= 36000) {
```
paulbourelly999, line 34: Checks against invalid timemark values. See PR description for details.

streets_utils/streets_vehicle_list/src/models/vehicle_list.cpp
```diff
@@ -54,9 +54,8 @@ namespace streets_vehicles {
         uint64_t timeout_time = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch()).count() - timeout;
         for ( auto it = vehicles.begin(); it != vehicles.end();  ) {
             vehicle veh =it->second;
-            SPDLOG_DEBUG("Checking Vehicle {0} timestamp {1} < timeout time {2} !", veh._id, veh._cur_time, timeout );
```
paulbourelly999, line 57: Not useful log statement.

message_services/include/message_lanelet2_translation.h
```diff
@@ -147,12 +149,27 @@ namespace message_services
...
+
             /**
             * @brief Function to convert GPS location to a 3d point in map frame
             * @param GPS location point
             * @return 3d point in map frame
             */
             lanelet::BasicPoint3d gps_2_map_point(double lat, double lon, double elev) const;
+
+            /**
+             * @brief
```
adev4a, line 169: Looks like this needs to be filled out

message_services/include/message_lanelet2_translation.h
```diff
@@ -147,12 +149,27 @@ namespace message_services
             */
             lanelet::BasicPoint3d ecef_2_map_point(std::int32_t ecef_x, std::int32_t ecef_y, std::int32_t ecef_z) const;
 
+            /**
+            * @brief Function to convert long lat location to a 2d ecef location
+            * @param lat
```
adev4a, line 154: Maybe we can add the units its expecting these arguments in

message_services/lib/message_lanelet2_translation.cpp
```diff
@@ -117,11 +114,6 @@ namespace message_services
         {                  
             std::vector<lanelet::Lanelet> current_total_lanelets = get_cur_lanelets_by_point(subj_point3d);   
             std::vector<lanelet::Lanelet> result_lanelets;
-            if (current_total_lanelets.empty())
```
adev4a, line 120: Will the rest of this method run anything if current_total_lanelets in <=1? 
Can we not exit it here?

message_services/lib/message_lanelet2_translation.cpp
```diff
@@ -338,6 +340,39 @@ namespace message_services
...
+                    // Get possible link lanelets
+                    lanelet::ConstLanelets possible_links = vehicleGraph_ptr->previous(*ll_itr);
+                    lanelet::ConstLanelet possible_link;
+                    for ( const auto &lane : possible_links ) {
+                        if ( lane.hasAttribute("turn_direction") && lane.attribute("turn_direction").value() == turn_direction ) {
+                            possible_link = lane;
+                            break;
+                        }
+                    }
+                    if ( possible_link.id() != 0 ) {
```
adev4a, line 357: Does the possible_link.id() initialize to 0? If not we could be passing this condition with an invalid lanelet

message_services/lib/message_lanelet2_translation.cpp
```diff
@@ -338,6 +340,39 @@ namespace message_services
...
+                        lanelet::ConstLanelet possible_entry = vehicleGraph_ptr->previous(possible_link).front();
+                        if (possible_entry.regulatoryElements().size() > 0)
```
adev4a, line 359: Can we change this to .empty()

message_services/lib/message_lanelet2_translation.cpp
paulbourelly999: If you look at the higher level if, it is already checking for if (current_total_lanelets.size() > 1) else if (current_total_lanelet.size() ==1) else // has to be empty() or invalid size. Just moved the log statement there

message_services/lib/message_lanelet2_translation.cpp
paulbourelly999: We check for this at the bottom of this method. I think this check is just for ensuring that a link lanelet was found and not passing the initialized variable if non is found.

adev4a: But we could be passing an uninitialized and unassigned variable here too right?

paulbourelly999: After looking into this I figured it out. So I have already covered both conditions of this if statement but there is a slight change I think is necessary. So here is the constructor for ConstLanelet:
```
explicit ConstLanelet(Id id = InvalId)
...
InvalId =0
```
This means that  checking `id == 0` and `id== InvalId` are equivalent so I will make sure we are consistent about which one we use.

Issue #247, opened 2022-10-28
Add streets_signal_optimization to signal_opt_service


Saeid-Solei:
### Types of Issue

- [ ] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [x] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

The streets_signal_optimization library contains the logic for calculating the queue length and dissipation time for each entry lane, find the list of candidate desired phase plans, and pick the one that has the highest delay ratio. This story is created to add the streets_signal_optimization library to signal_opt_service.

### Carma streets version where this issue was discovered

4.2.0

### Expected behavior
signal_opt_service can execute the streets_signal_optimization library.

### Actual behavior

The streets_signal_optimization library is not included in signal_opt_service yet.

### Steps to reproduce the actual behavior

NA

PR #248, opened 2022-11-01
247 Add streets_signal_optimization to signal_opt_service

Saeid-Solei:
# PR Details
## Description

The streets_signal_optimization library contains the logic for calculating the queue length and dissipation time for each entry lane, find the list of candidate desired phase plans, and pick the one that has the highest delay ratio. This story is created to add the streets_signal_optimization library to signal_opt_service.

## Motivation and Context

https://github.com/usdot-fhwa-stol/carma-streets/issues/247

## How Has This Been Tested?

Yes, with Unit Tests.

## Types of changes

## Checklist:


signal_opt_service/CMakeLists.txt
```diff
@@ -39,22 +39,21 @@ target_include_directories(${PROJECT_NAME}_lib PUBLIC
...
     intersection_client_api_lib
-    streets_signal_optimization_lib
-    streets_vehicle_scheduler_lib
-    streets_service_base_lib::streets_service_base_lib
-    streets_vehicle_list_lib
     streets_signal_phase_and_timing_lib
     streets_desired_phase_plan_lib
+    streets_vehicle_list_lib
+    streets_vehicle_scheduler_lib
+    streets_service_base_lib::streets_service_base_lib
```
paulbourelly999, line 54: I think you can remove all the stuff after :: we found out. Just so its consistent

signal_opt_service/include/signal_opt_processing_worker.h
```diff
@@ -15,32 +16,69 @@ namespace signal_opt_service
...
 
     public:
+
         signal_opt_processing_worker() = default;
+        
+
         ~signal_opt_processing_worker() = default;
 
+        /**
+         * @brief Configure streets_signal_optimization library.
```
paulbourelly999, line 38: **Non critical**: Comment and method name do not make sense together. Either configures signal_opt_processing_worker or configures arbitrator and generator  or rename work to dpp optimizer.

signal_opt_service/manifest.json
```diff
@@ -50,6 +68,30 @@
             "description": "Number of attempts intersection client sending the HTTP request to get response with valid signal group ids.",
             "type": "INTEGER"
         },
+        {
+            "name": "dpp_producer_sleep_time",
```
paulbourelly999, line 72: So can we name/describe this configuration parameter a little better. I assume we only have one frequency for signal optimization now. So this is the time between checking if the spat has  <= configured max fixed phases right.

signal_opt_service/manifest.json
```diff
@@ -63,16 +105,40 @@
             "type": "INTEGER"
         },
         {
-            "name": "tsc_config_consumer_topic",
-            "value": "tsc_config_state",
-            "description": "Kafka topic for CARMA Streets internal traffic signal controller phase configuration message.",
-            "type": "STRING"
+            "name": "et_inaccuracy_buffer",
+            "value": 2000,
```
paulbourelly999, line 109: Is 2 second default reasonable. Lets start with this at 0 and then with performance tune it up to try and improve behavior.

signal_opt_service/manifest.json
```diff
@@ -63,16 +105,40 @@
...
+            "description": "The configurable time interval in milliseconds defined to account for entering time (ET) estimation inaccuracy.",
+            "type": "INTEGER"
         },
         {
-            "name": "tsc_config_group_id",
-            "value": "tsc_config_group",
-            "description": "Kafka consumer group for SO tsc_config_consumer.",
-            "type": "STRING"
+            "name": "queue_max_time_headway",
+            "value": 3000,
```
paulbourelly999, line 115: Is 3 seconds reasonable here. Also will this impact our testing at all without queuing. If it won't impact our testing maybe we keep it at 0 to indicate/ensure it does not impact anything unintended.

signal_opt_service/manifest.json
```diff
@@ -63,16 +105,40 @@
...
+        },
+        {
+            "name": "so_radius",
+            "value": 200.0,
+            "description": "The configurable distance in meters defined as the radius of the signal optimization (SO) area.",
+            "type": "DOUBLE"
+        },
+        {
+            "name": "min_green",
+            "value": 5000,
```
paulbourelly999, line 127: Ok so we don't use the values from tsc config?

signal_opt_service/src/signal_opt_service.cpp
```diff
@@ -127,16 +149,64 @@ namespace signal_opt_service
...
+    void signal_opt_service::produce_dpp(const std::shared_ptr<kafka_clients::kafka_producer_worker> dpp_producer,
+                                const std::shared_ptr<OpenAPI::OAIIntersection_info> _intersection_info_ptr, 
+                                const std::shared_ptr<streets_vehicles::vehicle_list> _vehicle_list_ptr, 
+                                const std::shared_ptr<signal_phase_and_timing::spat> _spat_ptr, 
+                                const std::shared_ptr<streets_tsc_configuration::tsc_configuration_state> _tsc_config_ptr, 
+                                const std::shared_ptr<streets_signal_optimization::movement_groups> _movement_groups_ptr,
+                                const streets_signal_optimization::streets_desired_phase_plan_generator_configuration &_dpp_config) const
+    {
+        
+        SPDLOG_INFO("Starting desired phase plan producer thread.");
```
paulbourelly999, line 162: Not a useful log statement.

streets_utils/streets_signal_optimization/src/streets_desired_phase_plan_arbitrator.cpp
```diff
@@ -53,6 +65,13 @@ namespace streets_signal_optimization
                 throw(streets_desired_phase_plan_arbitrator_exception("There is no desired phase plan and delay measures mapping."));
             }
 
+            if (enable_so_logging) {
+                auto logger = spdlog::get("csv_so_logger");
+                if ( logger != nullptr ) {
+                    logger->info( "\n" );
```
paulbourelly999, line 71: Why are we logging new lines here. Also this will log 2 new lines. Is that intentional.

streets_utils/streets_signal_optimization/src/streets_desired_phase_plan_arbitrator.cpp
```diff
@@ -152,4 +182,42 @@ namespace streets_signal_optimization
...
     }
+
+    std::string streets_desired_phase_plan_arbitrator::dpp_delay_toCSV(
+                const streets_desired_phase_plan::streets_desired_phase_plan &candidate_dpp, 
+                const u_int64_t candidate_vehicle_delay, 
+                const u_int64_t TBD_delay, 
+                const float delay_measure) const 
+    {
+        
+        std::string so_info = "";
```
paulbourelly999, line 193: This is not csv. Or atleast not the way we have been using CSV. There is no need to write the field names for each entry. The optimal way to do this would be to write the first line of headers in the configure method and then just log values for every new line.

streets_utils/streets_signal_optimization/src/streets_desired_phase_plan_arbitrator.cpp
```diff
@@ -152,4 +182,42 @@ namespace streets_signal_optimization
...
+        so_info += "candidate vehicle delay";
+        so_info += ",";
+        so_info += std::to_string(candidate_vehicle_delay) + ",";
+        so_info += "TBD delay";
+        so_info += ",";
+        so_info += std::to_string(TBD_delay) + ",";
+        so_info += "delay measure";
+        so_info += ",";
+        so_info += std::to_string(delay_measure);
+        so_info += "\n";
```
paulbourelly999, line 219: A call to SPDLOG_INFO(payload) will already append an \n. If you include a additional one each log statement will be separated by a new line.

streets_utils/streets_signal_optimization/test/test_streets_desired_phase_plan_arbitrator.cpp
```diff
@@ -249,6 +253,10 @@ namespace streets_signal_optimization
                 }
                 return updates;
             }
+            else {
+                SPDLOG_ERROR("vehicle update is not array!");
```
paulbourelly999, line 257: DO NOT DO THIS! Either log and error or throw and exception. What ever calls this method or its parent will have to catch the exception and then chose to log something or do something or both but throwing an exception and logging the same thing is a log is a little redundant.

kafka_clients/include/kafka_producer_worker.h
```diff
@@ -80,11 +80,47 @@ namespace kafka_clients
...
+             * @param msg message to produce.
+             */
+            virtual void send(const std::string &msg);
+            /**
+             * @brief Is kafka_producer_worker still running?
+             * 
+             * @return true if kafka producer is still running.
+             * @return false if kafka producer is stopped.
+             */
+            virtual bool is_running() const;
```
paulbourelly999, line 110: Is this a method you added? Why is this method necessary?

kafka_clients/include/kafka_producer_worker.h
```diff
@@ -80,11 +80,47 @@ namespace kafka_clients
...
+             * 
+             * @param msg message to produce.
+             */
+            virtual void send(const std::string &msg);
+            /**
+             * @brief Is kafka_producer_worker still running?
+             * 
+             * @return true if kafka producer is still running.
+             * @return false if kafka producer is stopped.
+             */
```
dan-du-car, line 109: Is this part of your changes?

streets_utils/streets_signal_optimization/include/streets_desired_phase_plan_generator.h
```diff
@@ -45,6 +45,19 @@ namespace streets_signal_optimization {
              */
             std::unordered_map<uint8_t, std::vector<int>> signal_group_entry_lane_mapping;
 
+            /**
+             * @brief Given an intersection_info pointer and a list of possible movement groups, this method checks if all
+             * signal groups included in the movement groups list are included in the intersection_info pointer. Basically,
```
dan-du-car, line 50: Why movement group might not be in the intersection model?

streets_utils/streets_signal_optimization/src/streets_desired_phase_plan_arbitrator.cpp
```diff
@@ -36,12 +38,22 @@ namespace streets_signal_optimization
...
+                schedule_ptr->vehicle_schedules.clear();
+                for (const auto& veh_sched : all_schedule_ptr->vehicle_schedules) {
+                    if (veh_sched.state == streets_vehicles::vehicle_state::EV && veh_list_ptr->get_vehicles().at(veh_sched.v_id)._cur_distance <= so_radius) {
+                        schedule_ptr->vehicle_schedules.push_back(veh_sched);
+                    }
+                }
 
                 // Given vehicles' ET and EET from signalized intersection schedule, Calculate vehicles' delay measure for the current candidatedesired phase plan
-                float delay_measure = calculate_delay_measure(schedule_ptr, candidate_dpp);
+                float delay_measure = calculate_delay_measure(schedule_ptr, candidate_dpp, enable_so_logging);
```
dan-du-car, line 56: Why do we need to enable logging for this lib? but not other libs?

signal_opt_service/manifest.json
Saeid-Solei: Do you have any name in mind?

signal_opt_service/manifest.json
Saeid-Solei: Tsc_config does not have min_green member!

streets_utils/streets_signal_optimization/src/streets_desired_phase_plan_arbitrator.cpp
Saeid-Solei: Yes, I wanted to create one empty line gap between two consecutive iterations of the SO service. And why  two new lines?

kafka_clients/include/kafka_producer_worker.h
Saeid-Solei: yes

signal_opt_service/src/signal_opt_service.cpp
```diff
@@ -127,16 +149,64 @@ namespace signal_opt_service
...
+        int sleep_millisecs = streets_service::streets_configuration::get_int_config("dpp_producer_sleep_time");
+        auto sleep_secs = static_cast<unsigned int>(sleep_millisecs / 1000);
+        int prev_future_move_group_count = 0;
+        int current_future_move_group_count = 0;
+        while ( dpp_producer->is_running() ) {
+            
+            // First, convert the spat to desired phase plan
+            streets_desired_phase_plan::streets_desired_phase_plan spat_dpp = _so_processing_worker_ptr->convert_spat_to_dpp(_spat_ptr, _movement_groups_ptr);
```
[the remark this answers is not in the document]
Saeid-Solei: I believe it should be the latest spat.

streets_utils/streets_signal_optimization/include/streets_desired_phase_plan_generator.h
Saeid-Solei: Basically, the list of movement groups could include a signal group that is not defined in the map\intersection_info. There is no check to protect against this anywhere else. And the dpp_generator requires all signal groups used in movement group list be defined in intersection_info.

streets_utils/streets_signal_optimization/src/streets_desired_phase_plan_arbitrator.cpp
```diff
@@ -36,12 +38,22 @@ namespace streets_signal_optimization
...
+                // set_timestamp for the all_schedule_ptr current Unix timestamp
+                calculate_vehicle_schedules(all_schedule_ptr, local_spat_ptr, veh_list_ptr, intersection_info_ptr, initial_green_buffer, final_green_buffer);
+
+                // Add those vehicles that are within SO radius to schedule_ptr
                 auto schedule_ptr = std::make_shared<streets_vehicle_scheduler::signalized_intersection_schedule>();
-                // set_timestamp for the schedule_ptr current Unix timestamp
-                calculate_vehicle_schedules(schedule_ptr, local_spat_ptr, veh_list_ptr, intersection_info_ptr, initial_green_buffer, final_green_buffer);
+                schedule_ptr->timestamp = all_schedule_ptr->timestamp;
+                schedule_ptr->vehicle_schedules.clear();
+                for (const auto& veh_sched : all_schedule_ptr->vehicle_schedules) {
```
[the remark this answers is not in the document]
Saeid-Solei: I have added a comment to say why this new piece is added (Line 45: // Add those vehicles that are within SO radius to schedule_ptr)

signal_opt_service/CMakeLists.txt
Saeid-Solei: Done

signal_opt_service/include/signal_opt_processing_worker.h
Saeid-Solei: Done

signal_opt_service/CMakeLists.txt
paulbourelly999: It should be, there is a different issue if you run into a compilation issue here. All the libraries we include, we don't use the semicolon. Not sure why it would be necessary here.

signal_opt_service/include/signal_opt_processing_worker.h
```diff
@@ -15,32 +16,69 @@ namespace signal_opt_service
...
-         *
+         * @brief Configure the dpp_generator and dpp_arbitrator in signal_opt_processing_worker.
+         * 
+         * @param dpp_config desired_phase_plan_generator_configuration object
+         */
+        void configure_dpp_optimizer(const streets_signal_optimization::streets_desired_phase_plan_generator_configuration &dpp_config);
+
+        /**
+         * @brief Convert the provided intersection_state to a desired phase plan. The number of fixed future movement groups
+         * shall be at least 1.
```
paulbourelly999, line 46: **NON CRITICAL** Does this method throw an exception if less than one fixed movement group is represent in the spat? If so may want to add a @throws description of when this method will throw an exception. Also as an external person it may be helpful to know what is considered a fixed movement group. 
```
A fixed movement group is 1 or 2 signal groups in the spat having a future or current protected movement event with a start and end time.
```

signal_opt_service/manifest.json
paulbourelly999: Signal Optimization Interval or Signal Optimization Frequency or Signal Optimization Phase Decision Frequency. Sleep time tells very little about what the configuration property does. Also please update the description to include a description of what this property changes so that someone familiar with SO but maybe not our internal implementation could figure out what this property controls.

signal_opt_service/manifest.json
```diff
@@ -50,6 +68,30 @@
             "description": "Number of attempts intersection client sending the HTTP request to get response with valid signal group ids.",
             "type": "INTEGER"
         },
+        {
+            "name": "dpp_producer_sleep_time",
+            "value": 1000,
```
paulbourelly999, line 73: Is 1 second a reasonable default value?

signal_opt_service/manifest.json
```diff
@@ -63,16 +105,40 @@
             "type": "INTEGER"
         },
         {
-            "name": "tsc_config_consumer_topic",
-            "value": "tsc_config_state",
-            "description": "Kafka topic for CARMA Streets internal traffic signal controller phase configuration message.",
-            "type": "STRING"
+            "name": "et_inaccuracy_buffer",
+            "value": 0,
+            "description": "The configurable time interval in milliseconds defined to account for entering time (ET) estimation inaccuracy.",
```
paulbourelly999, line 110: Just for my own knowledge, what specifically will this buffer impact?

signal_opt_service/manifest.json
paulbourelly999: No the purpose of the parameter is that we can use it to account for queuing vehicles. Since we don't have queuing vehicle we don't need it to have a default value. My outlook on this is that a 0 default value is better than an arbitrary default value, especially one who's impact we don't clearly know yet.

signal_opt_service/manifest.json
paulbourelly999: tsc config has a min green for each phase and and a max green for each phase. I assume we are using this as a general configuration for all SG that overrides the values set on the TSC. One thing we have to keep an eye on, also may be worth mentioning in the description is that any min green set here needs to be >= the min green set for any active phase on the controller, or else it won't be achievable.

signal_opt_service/src/signal_opt_processing_worker.cpp
```diff
@@ -1,17 +1,61 @@
...
+                        const std::shared_ptr<streets_signal_optimization::movement_groups> &move_groups) const
+    {
+        streets_desired_phase_plan::streets_desired_phase_plan spat_dpp;
+    
+        signal_phase_and_timing::intersection_state int_state = spat_ptr->get_intersection();
+        if (dpp_generator_ptr) {
+            spat_dpp = dpp_generator_ptr->convert_spat_to_dpp(int_state, move_groups);
+        }
+        else {
+            SPDLOG_DEBUG("dpp_generator_ptr is not initialized!");
```
paulbourelly999, line 30: If the dpp generator is not initialized should we be throwing and exception or at least logging something higher than info. The method that calls this, would it fail if given an empty desired phase plan. If so, we may as well throw an exception here.

signal_opt_service/src/signal_opt_processing_worker.cpp
```diff
@@ -1,17 +1,61 @@
...
+        const std::shared_ptr<streets_signal_optimization::movement_groups> &move_groups,
+        const streets_signal_optimization::streets_desired_phase_plan_generator_configuration &dpp_config) const
     {
-        streets_signal_optimization::streets_desired_phase_plan_arbitrator arbitrator;
-        auto chosen_dpp = arbitrator.select_optimal_dpp(dpp_list, intersection_info_ptr, spat_ptr, tsc_config_state, veh_list_ptr, initial_green_buffer, final_green_buffer);
+        
+        bool enable_so_logging = streets_service::streets_configuration::get_boolean_config("enable_so_logging");
+        auto intersection_state = spat_ptr->get_intersection();
+        auto vehicle_map = veh_list_ptr->get_vehicles();
+        SPDLOG_DEBUG("so_processing_worker - dpp_generator start!");
```
paulbourelly999, line 48: Could we replace these log statements with something more informative. What are we trying to capture here?

signal_opt_service/src/signal_opt_processing_worker.cpp
```diff
@@ -1,17 +1,61 @@
...
+        const std::shared_ptr<signal_phase_and_timing::spat> &spat_ptr, 
+        const std::shared_ptr<streets_tsc_configuration::tsc_configuration_state> &tsc_config_state, 
+        const std::shared_ptr<streets_vehicles::vehicle_list> &veh_list_ptr,
+        const std::shared_ptr<streets_signal_optimization::movement_groups> &move_groups,
+        const streets_signal_optimization::streets_desired_phase_plan_generator_configuration &dpp_config) const
     {
-        streets_signal_optimization::streets_desired_phase_plan_arbitrator arbitrator;
-        auto chosen_dpp = arbitrator.select_optimal_dpp(dpp_list, intersection_info_ptr, spat_ptr, tsc_config_state, veh_list_ptr, initial_green_buffer, final_green_buffer);
+        
+        bool enable_so_logging = streets_service::streets_configuration::get_boolean_config("enable_so_logging");
```
paulbourelly999, line 45: **NON CRITICAL** Next time it is probably better to pass this boolean parameter in from the service level. Even though we have access to the configurations throughout the code I think it is good practice to try and organize configuration calls into a common place so that people looking through our code know where to look for stuff. It does not make any functional difference but I think since we have almost all our configuration calls in service classes for all of our CARMA-Streets Services, this is a good practice to keep up.

signal_opt_service/src/signal_opt_service.cpp
```diff
@@ -43,11 +47,22 @@ namespace signal_opt_service
...
 
+            if(!_dpp_producer->init())
+            {
+                SPDLOG_CRITICAL("kafka producer (_dpp_producer) initialize error!");
+                exit(EXIT_FAILURE);
+            }
+
+            // Create logger
+            if ( streets_service::streets_configuration::get_boolean_config("enable_so_logging") ) {
+                configure_csv_logger();
```
paulbourelly999, line 63: Can this method fail, if so, what do we want to do on failure.

signal_opt_service/src/signal_opt_service.cpp
```diff
@@ -43,11 +47,22 @@ namespace signal_opt_service
...
-                    SPDLOG_CRITICAL("kafka consumers ( _spat_consumer, _tsc_config_consumer or _vsi_consumer) is not running");
+                    SPDLOG_CRITICAL("kafka consumers ( _spat_consumer, _tsc_config_consumer or _vsi_consumer) is not running!");
                     exit(EXIT_FAILURE);
                 }
             }
 
+            if(!_dpp_producer->init())
+            {
+                SPDLOG_CRITICAL("kafka producer (_dpp_producer) initialize error!");
+                exit(EXIT_FAILURE);
```
paulbourelly999, line 58: **NON Critical** I am not a fan of initialize methods that can call exit and can return false. We should pick one. I prefer we return a boolean and allow the calling file to decide how to proceed based on the state of initialization.  I this case we would return false for kafka initialization issues and main.cpp would call exit on false returns.

signal_opt_service/src/signal_opt_service.cpp
```diff
@@ -127,16 +149,64 @@ namespace signal_opt_service
...
+    void signal_opt_service::produce_dpp(const std::shared_ptr<kafka_clients::kafka_producer_worker> dpp_producer,
+                                const std::shared_ptr<OpenAPI::OAIIntersection_info> _intersection_info_ptr, 
+                                const std::shared_ptr<streets_vehicles::vehicle_list> _vehicle_list_ptr, 
+                                const std::shared_ptr<signal_phase_and_timing::spat> _spat_ptr, 
+                                const std::shared_ptr<streets_tsc_configuration::tsc_configuration_state> _tsc_config_ptr, 
+                                const std::shared_ptr<streets_signal_optimization::movement_groups> _movement_groups_ptr,
+                                const streets_signal_optimization::streets_desired_phase_plan_generator_configuration &_dpp_config) const
+    {
+        
+        SPDLOG_DEBUG("Starting desired phase plan producer thread.");
```
paulbourelly999, line 162: Not a useful log statement

signal_opt_service/src/signal_opt_service.cpp
```diff
@@ -127,16 +149,64 @@ namespace signal_opt_service
...
+        int sleep_millisecs = streets_service::streets_configuration::get_int_config("dpp_producer_sleep_time");
```
paulbourelly999, line 163: **NON CRITICAL** See earlier comment about retrieving configuration values in a common place.

signal_opt_service/src/signal_opt_service.cpp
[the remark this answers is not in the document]
paulbourelly999: If this currently returns an empty desired phase plan we will still try to run the rest of the functions. What will happen if we try to run the other functions with an empty desired phase plan. If we want to guard against this either make `convert_spat_to_dpp` throw an exception when  dpp generator is not initialized or something fails and catch that exception here or add java doc comment that outlines that  method will return empty desired phase plan if dpp_generator is not initialized or something fails and add a check here to ignore cases in which that happens

signal_opt_service/src/signal_opt_service.cpp
```diff
@@ -191,13 +261,48 @@ namespace signal_opt_service
         return false;
     }
 
+    void signal_opt_service::configure_so_processing_worker() {
+        _so_processing_worker_ptr = std::make_shared<signal_opt_processing_worker>();
+        //Parameter config
+        dpp_config.initial_green_buffer = streets_service::streets_configuration::get_int_config("initial_green_buffer");
```
paulbourelly999, line 267: **NON CRITICAL** See earlier comment about retrieving configuration parameter in a common place. Currently that is the initialize() method for most of our service.

signal_opt_service/test/test_signal_opt_processing_worker.cpp
```diff
@@ -25,7 +25,7 @@ namespace signal_opt_service
...
...7, 29, 32], \"id\":24, \"length\":10.29555911705508, \"speed_limit\":4.4, \"signal_group_id\":3 }, { \"conflict_lanelet_ids\":[27, 32], \"id\":25, \"length\":15.85394784011176, \"speed_limit\":8.0, \"signal_group_id\":3 }, { \"conflict_lanelet_ids\":[21, 23, 27, 29, 30, 31], \"id\":26, \"length\":9.74459032026013, \"speed_limit\":4.4, \"signal_group_id\":8 }, { \"conflict_lanelet_ids\":[23, 24, 25, 26, 30, 32], \"id\":27, \"length\":13.64738192837192, \"speed_limit\":8.0, \"signal_group_id\":5 }, { \"conflict_lanelet_ids\":[23, 30], \"id\":28, \"length\":8.18273610098126, \"speed_limit\":4.4, \"signal_group_id\":5 }, { \"conflict_lanelet_ids\":[21, 22, 24, 26, 30, 32], \"id\":29, \"length\":15.85394784011176, \"speed_limit\":8.0, \"signal_group_id\":2 }, { \"conflict_lanelet_ids\":[21, 23, 26, 27, 28, 29], \"id\":30, \"length\":9.74459032026013, \"speed_limit\":4.4, \"signal_group_id\":7 }, { \"conflict_lanelet_ids\":[21, 26], \"id\":31, \"length\":13.64738192837192, \"speed_limit\":8.0, \"signal_group_id\":7 }, { \"conflict_lanelet_ids\":[21, 23, 24, 25, 27, 29], \"id\":32, \"length\":8.18273610098126, \"speed_limit\":4.4, \"signal_group_id\":4 }], \"name\":\"WestIntersection\"}";
```
paulbourelly999, line 28: **NON CRITICAL** This is fine for now since we do it in other places as well but for the future we should move away from including single line giant JSON like this. It is not readable. Either we should add it to a file and read from there or indent it in a way that makes it readable.

streets_utils/streets_signal_optimization/src/streets_desired_phase_plan_arbitrator.cpp
```diff
@@ -36,12 +38,22 @@ namespace streets_signal_optimization
                 update_spat_with_candidate_dpp(local_spat_ptr, candidate_dpp, tsc_state);
 
                 // Given spat and vehicle list, estimate current vehicles' ET and EET from signalized intersection schedule
+                auto all_schedule_ptr = std::make_shared<streets_vehicle_scheduler::signalized_intersection_schedule>();
+                // set_timestamp for the all_schedule_ptr current Unix timestamp
+                calculate_vehicle_schedules(all_schedule_ptr, local_spat_ptr, veh_list_ptr, intersection_info_ptr, initial_green_buffer, final_green_buffer);
+
+                // Add those vehicles that are within SO radius to schedule_ptr
```
paulbourelly999, line 45: **NON CRITICAL** update to
```
Add vehicle schedules from all_schedule_ptr for vehicle that are in SO area.
```

streets_utils/streets_signal_optimization/src/streets_desired_phase_plan_arbitrator.cpp
paulbourelly999: **NON CRITICAL** Ideally this would be a private member that you can set on the class in one place instead of making it a parameter that has to be passed into multiple methods. Non critical though

streets_utils/streets_signal_optimization/src/streets_desired_phase_plan_arbitrator.cpp
paulbourelly999: Why? We have to be better with formatting for CSV logs. Also this will create two empty lines which is definitely redundant. I almost prefer no empty lines and us having to use timestamps to differentiate runs because it makes the log files smaller and more concise. We can always using parsing logic to parse out runs based on some timing or state information.

streets_utils/streets_signal_optimization/src/streets_desired_phase_plan_arbitrator.cpp
```diff
@@ -152,4 +182,34 @@ namespace streets_signal_optimization
...
+                const streets_desired_phase_plan::streets_desired_phase_plan &candidate_dpp, 
+                const u_int64_t candidate_vehicle_delay, 
+                const u_int64_t TBD_delay, 
+                const float delay_measure) const 
+    {
+        
+        std::string so_info = "";
+        // timestamp
+        so_info += std::to_string(candidate_dpp.timestamp) + ",";
+        for (const auto &move_group : candidate_dpp.desired_phase_plan) {
```
paulbourelly999, line 196: So this information will change in length based on the length of the desired phase plan. Lets move all other constant length information about this for parsing. This can then be used purely as additional information we can manually use for debugging.

paulbourelly999: Alternatively we can also have this be a single comma separated value that is a string JSON payload. We still want it at the end. Something like this :

```
val, constant_length_value, constant length value , "{desired_phase_plan json}"
```

streets_utils/streets_signal_optimization/test/test_streets_desired_phase_plan_arbitrator.cpp
```diff
@@ -249,6 +253,9 @@ namespace streets_signal_optimization
                 }
                 return updates;
             }
+            else {
+                throw streets_vehicles::status_intent_processing_exception("vehicle update is not array!");
```
paulbourelly999, line 257: See earlier comment about error handling in unit tests. It is not necessary to throw exceptions in unit testing helper functions but if you do it only makes sense if you include some sort of error handling in the test case that call the helper method to fail or do something when the helper method throws and exception.

paulbourelly999 (changes requested):
Great job with the mock kafka producer. Most of this looks good. So some small stuff and then some non critical stuff. Also gave some future pointers. Looks like more work than we initially anticipated. Thanks for doing it. Feel free to use your digression for the non critical stuff. Just leave a comment for the ones you address.

signal_opt_service/manifest.json
Saeid-Solei: Done!

signal_opt_service/manifest.json
Saeid-Solei: Done!

signal_opt_service/manifest.json
Saeid-Solei: Done!

signal_opt_service/src/signal_opt_service.cpp
Saeid-Solei: Changed to DEBUG.

streets_utils/streets_signal_optimization/src/streets_desired_phase_plan_arbitrator.cpp
Saeid-Solei: Sure, I have removed this extra line.

streets_utils/streets_signal_optimization/src/streets_desired_phase_plan_arbitrator.cpp
Saeid-Solei: I have removed the field names.

streets_utils/streets_signal_optimization/src/streets_desired_phase_plan_arbitrator.cpp
Saeid-Solei: Removed "\n" line.

streets_utils/streets_signal_optimization/test/test_streets_desired_phase_plan_arbitrator.cpp
Saeid-Solei: Removed SPDLOG_ERROR.

signal_opt_service/include/signal_opt_processing_worker.h
```diff
@@ -15,32 +16,69 @@ namespace signal_opt_service
...
         ~signal_opt_processing_worker() = default;
 
         /**
-         * @brief Call streets_signal_optimization library to choose the optimized desired phase plan.
-         *
+         * @brief Configure the dpp_generator and dpp_arbitrator in signal_opt_processing_worker.
+         * 
+         * @param dpp_config desired_phase_plan_generator_configuration object
+         */
+        void configure_dpp_optimizer(const streets_signal_optimization::streets_desired_phase_plan_generator_configuration &dpp_config);
```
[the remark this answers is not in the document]
Saeid-Solei: Done!

signal_opt_service/include/signal_opt_processing_worker.h
```diff
@@ -15,32 +16,69 @@ namespace signal_opt_service
...
+
+        /**
+         * @brief Convert the provided intersection_state to a desired phase plan. The number of fixed future movement groups
```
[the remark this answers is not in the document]
Saeid-Solei: Good catch. I have copied this description from streets_desired_phase_plan_generator. Revised.

signal_opt_service/include/signal_opt_processing_worker.h
Saeid-Solei: The streets_desired_phase_plan_generator library will throw an exception if the number of fixed movement group is 0. Would adding a try\catch method in signal_opt_service fix the issue? Also, since we explained when the streets_desired_phase_plan_generator library throws an exception, do we need to include the throws description here as well?

dan-du-car (dismissed):

signal_opt_service/include/signal_opt_service.h
```diff
@@ -73,14 +82,32 @@ namespace signal_opt_service
          */
         void consume_tsc_config(const std::shared_ptr<kafka_clients::kafka_consumer_worker> tsc_config_consumer, 
                                 const std::shared_ptr<streets_tsc_configuration::tsc_configuration_state> _tsc_config_ptr) const;
+        /**
+         * @brief Method to produce desired phase plan to kafka producer.
```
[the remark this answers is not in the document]
Saeid-Solei: Done!

signal_opt_service/manifest.json
Saeid-Solei: Technically, since our min_green is around 5 secs, we could increase this value. I have no preference here, but I want to make sure that the signal optimization service has at least 2 chances to send a desired phase plan during one green phase.

signal_opt_service/manifest.json
Saeid-Solei: When calculating the required green duration for each entry lane, this buffer will be added to the estimated ET of the last vehicle from a queue. Basically, the end time of the required green will be:

end time of the green = ET of last vehicle in the queue + et_inaccuracy_buffer + final_green_buffer.

signal_opt_service/src/signal_opt_processing_worker.cpp
Saeid-Solei: So, the library will throw an exception if the desired_phase_plan is empty. But I agree that here we should also throw an exception and do not continue. I have added a throw statement here and will catch it in signal_opt_service.

signal_opt_service/src/signal_opt_processing_worker.cpp
Saeid-Solei: I have added these when debugging. I forgot to remove them. I have removed them now.
Please let me know if you think a log statement here is necessary.

signal_opt_service/src/signal_opt_service.cpp
Saeid-Solei: I am not sure if it will fail. If it does, this method will catch the failure. We did not have any separate try\catch for this method in scheduling_service. So, I thought this should be okay.

signal_opt_service/include/signal_opt_processing_worker.h
paulbourelly999: Oh I missed that. If a lower method already throws an exception for this no need to add any additional exceptions. It is always nice to have documentation that is complete and independent of other methods documentation so including is good but not required I guess.

paulbourelly999: I still think a description for fixed movement group is helpful.

signal_opt_service/src/signal_opt_service.cpp
Saeid-Solei: Removed!

signal_opt_service/manifest.json
paulbourelly999: This would give us 5 but that is OK to. Ideally I would think you want this to run a little less frequent so in between runs you are exposed to an entire queue of vehicle instead of just a single on that has newly entered the SO radius

signal_opt_service/src/signal_opt_service.cpp
Saeid-Solei: Addressed!

signal_opt_service/src/signal_opt_service.cpp
Saeid-Solei: Addressed!

signal_opt_service/manifest.json
paulbourelly999: Hmm, not sure I understand why we separated this from final buffer. It seems to be exactly the same functionality.

paulbourelly999: Thats fine I guess. Maybe we want to put in an enhancement to remove this redundant configuration property.

streets_utils/streets_signal_optimization/src/streets_desired_phase_plan_arbitrator.cpp
Saeid-Solei: Done!

signal_opt_service/src/signal_opt_processing_worker.cpp
Saeid-Solei: Done!

signal_opt_service/test/test_data/updates_signalized.json
```diff
@@ -63,38 +63,38 @@
                 "max_decel": 5.0,
                 "cur_speed": 5.0,
```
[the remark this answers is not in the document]
Saeid-Solei: addressed.

signal_opt_service/test/test_signal_opt_processing_worker.cpp
```diff
@@ -240,6 +239,10 @@ namespace signal_opt_service
                 }
                 return updates;
             }
+            else {
+                SPDLOG_ERROR("vehicle update is not array!");
+                throw streets_vehicles::status_intent_processing_exception("vehicle update is not array!");
```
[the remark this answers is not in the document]
Saeid-Solei: Removed Throw!

streets_utils/streets_signal_optimization/src/streets_desired_phase_plan_arbitrator.cpp
```diff
@@ -36,12 +38,22 @@ namespace streets_signal_optimization
...
+                auto all_schedule_ptr = std::make_shared<streets_vehicle_scheduler::signalized_intersection_schedule>();
+                // set_timestamp for the all_schedule_ptr current Unix timestamp
+                calculate_vehicle_schedules(all_schedule_ptr, local_spat_ptr, veh_list_ptr, intersection_info_ptr, initial_green_buffer, final_green_buffer);
+
+                // Add those vehicles that are within SO radius to schedule_ptr
                 auto schedule_ptr = std::make_shared<streets_vehicle_scheduler::signalized_intersection_schedule>();
-                // set_timestamp for the schedule_ptr current Unix timestamp
-                calculate_vehicle_schedules(schedule_ptr, local_spat_ptr, veh_list_ptr, intersection_info_ptr, initial_green_buffer, final_green_buffer);
+                schedule_ptr->timestamp = all_schedule_ptr->timestamp;
+                schedule_ptr->vehicle_schedules.clear();
```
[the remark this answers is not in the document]
Saeid-Solei: Removed!

streets_utils/streets_signal_optimization/src/streets_desired_phase_plan_arbitrator.cpp
Saeid-Solei: Done!

streets_utils/streets_signal_optimization/test/test_streets_desired_phase_plan_arbitrator.cpp
Saeid-Solei: Removed throw.

signal_opt_service/manifest.json
Saeid-Solei: That is a great and valid point. I agree. We should increase this value. Is 5000 a reasonable value?

streets_utils/streets_signal_optimization/src/streets_desired_phase_plan_arbitrator.cpp
Saeid-Solei: Done!

signal_opt_service/include/signal_opt_processing_worker.h
Saeid-Solei: I have added the description.

signal_opt_service/src/signal_opt_service.cpp
Saeid-Solei: Addressed!

signal_opt_service/src/signal_opt_service.cpp
paulbourelly999: Good call we have a try catch in the method. That works too.

streets_utils/streets_signal_optimization/src/streets_desired_phase_plan_arbitrator.cpp
paulbourelly999: Instead of adding the desired phase plan as comma separated values just add it as a JSON in the last comma separated value as shown above. This way the number of columns in the CSV is constant.

signal_opt_service/CMakeLists.txt
paulbourelly999: Will address this later in another PR

paulbourelly999 (approved):
Looks good to me.

PR #252, opened 2022-11-09
245 fix next green

paulbourelly999:
# PR Details
## Types of changes

## Checklist:

### DIFF
--- a/tsc_client_service/include/ntcip_oids.h
+++ b/tsc_client_service/include/ntcip_oids.h
@@ -97,9 +97,10 @@ namespace ntcip_oids {
      * @brief Phase Next Status Mask returns a 8 bit value. When a bit = 1, this phase is currently commited to be 
      * next, when a bit = 0, this phase is NOT commited to be next. The phase commited to be next is determined at
      * the end of each green interval, except if that decision cannot be made. In this case it will wait until all 
-     * vehicel change and clearance intervales.
+     * vehicel change and clearance intervales. NOTE:Uses first entry in the phaseStatusGroupTable which is assumed
+     * to be the vehicle phase group.
      * 
      */
-    static const std::string PHASE_STATUS_GROUP_PHASE_NEXT = "1.3.6.1.4.1.1206.4.2.1.1.4.1.11";
+    static const std::string PHASE_STATUS_GROUP_PHASE_NEXT = "1.3.6.1.4.1.1206.4.2.1.1.4.1.11.1";
 
 }
--- a/tsc_client_service/include/ntcip_oids.h
+++ b/tsc_client_service/include/ntcip_oids.h
@@ -93,5 +93,13 @@ namespace ntcip_oids {
      * .1 is needed at the end to set the phase control. NTCIP documentation is unclear about why its required or what it represents
      */
     static const std::string PHASE_HOLD_CONTROL = "1.3.6.1.4.1.1206.4.2.1.1.5.1.4.1";
+    /**
+     * @brief Phase Next Status Mask returns a 8 bit value. When a bit = 1, this phase is currently commited to be 
+     * next, when a bit = 0, this phase is NOT commited to be next. The phase commited to be next is determined at
+     * the end of each green interval, except if that decision cannot be made. In this case it will wait until all 
+     * vehicel change and clearance intervales.
+     * 
+     */
+    static const std::string PHASE_STATUS_GROUP_PHASE_NEXT = "1.3.6.1.4.1.1206.4.2.1.1.4.1.11";
 
 }
--- a/tsc_client_service/src/monitor_desired_phase_plan.cpp
+++ b/tsc_client_service/src/monitor_desired_phase_plan.cpp
@@ -35,8 +35,9 @@ namespace traffic_signal_controller_service
             bool is_green_present = false;
             auto state = spat_ptr->get_intersection();
             for (const auto &movement : state.states) {
-                if ( movement.state_time_speed.front().event_state == signal_phase_and_timing::movement_phase_state::protected_clearance) {
+                if ( movement.state_time_speed.front().event_state == signal_phase_and_timing::movement_phase_state::protected_movement_allowed) {
                     is_green_present = true;
+                    SPDLOG_INFO("Green is present in SPat. No need to fix upcoming green");
                 }
             }
             if (!is_green_present ) {
--- a/tsc_client_service/src/monitor_desired_phase_plan.cpp
+++ b/tsc_client_service/src/monitor_desired_phase_plan.cpp
@@ -36,12 +36,10 @@ namespace traffic_signal_controller_service
             for (const auto &movement : state.states) {
                 if ( movement.state_time_speed.front().event_state == signal_phase_and_timing::movement_phase_state::protected_movement_allowed) {
                     is_green_present = true;
-                    SPDLOG_INFO("Green is present in SPat. No need to fix upcoming green");
                     break;
                 }
             }
             if (!is_green_present ) {
-                SPDLOG_INFO("This is happening!");
                 fix_upcoming_green(spat_ptr, tsc_state_ptr);
             }
         } else {
--- a/tsc_client_service/src/monitor_desired_phase_plan.cpp
+++ b/tsc_client_service/src/monitor_desired_phase_plan.cpp
@@ -2,6 +2,9 @@
 
 namespace traffic_signal_controller_service
 {
+    monitor_desired_phase_plan::monitor_desired_phase_plan(const std::shared_ptr<snmp_client> client) :  _snmp_client(client) {
+
+    }
     void monitor_desired_phase_plan::update_desired_phase_plan(const std::string &payload)
     {
         desired_phase_plan_ptr = std::make_shared<streets_desired_phase_plan::streets_desired_phase_plan>();
@@ -13,7 +16,7 @@ namespace traffic_signal_controller_service
         return desired_phase_plan_ptr;
     }
 
-    void monitor_desired_phase_plan::update_spat_future_movement_events(std::shared_ptr<signal_phase_and_timing::spat> spat_ptr, const std::shared_ptr<tsc_state> tsc_state_ptr) const
+    void monitor_desired_phase_plan::update_spat_future_movement_events(const std::shared_ptr<signal_phase_and_timing::spat> spat_ptr, const std::shared_ptr<tsc_state> tsc_state_ptr) const
     {
         if (tsc_state_ptr == nullptr || spat_ptr == nullptr)
         {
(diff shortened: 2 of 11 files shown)


tsc_client_service/src/monitor_desired_phase_plan.cpp
```diff
@@ -24,22 +27,99 @@ namespace traffic_signal_controller_service
...
-
         if (desired_phase_plan_ptr == nullptr || desired_phase_plan_ptr->desired_phase_plan.empty())
         {
-            throw monitor_desired_phase_plan_exception("Desired phase plan is empty. No update.");
+            // If no current green -> then either yellow change or red clearance active
+            // => Fix next green phase in SPaT
+            bool is_green_present = false;
+            auto state = spat_ptr->get_intersection();
+            for (const auto &movement : state.states) {
+                if ( movement.state_time_speed.front().event_state == signal_phase_and_timing::movement_phase_state::protected_movement_allowed) {
```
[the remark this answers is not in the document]
paulbourelly999: So there can be fixed greens in the spat when the desired phase plan is empty but there can only ever be one and it would have to be the current event for a given signal group. This is because the only thing that will add future events, in the case of using our desired phase plan, is desired phase plan projection, so there can not be future events not reflected in our desired phase plan. In a scenario, where we are fixing phases based solely on TSC configuration, none of this logic is run since we don't even listen for desired phase plans and definitely don't project them on to the SPaT.

tsc_client_service/src/monitor_desired_phase_plan.cpp
```diff
@@ -24,22 +27,99 @@ namespace traffic_signal_controller_service
...
+        uint64_t cur_time_since_epoch = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch()).count();
+        auto state = spat_ptr->get_intersection();
+
+        // Check if current spat includes yellow clearance
+        for (const auto &movement : state.states) {
+            // Found yellow states
+            if ( movement.state_time_speed.front().event_state == signal_phase_and_timing::movement_phase_state::protected_clearance ) {
+                is_yellow_change = true;
+                auto yellow_phase_configuration =tsc_state->get_signal_group_state_map().find(movement.signal_group)->second;
+                uint64_t local_start_time_epoch = movement.state_time_speed.front().timing.get_epoch_min_end_time() + yellow_phase_configuration.red_clearance;
```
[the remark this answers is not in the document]
paulbourelly999: if the current spat includes a yellow we have to capture the start time for the next green. This start time will be the yellow duration plus red clearance.

tsc_client_service/src/tsc_service.cpp
```diff
@@ -200,7 +201,16 @@ namespace traffic_signal_controller_service {
                     }
                     
                     spat_producer->send(spat_ptr->toJson());
-                    
+                    if (count <= 20 ) {
```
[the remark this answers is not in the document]
paulbourelly999: This is more of a logging parameter, we probably don't want end users configuring the sample size for SPaT latency calculations since that is probably to granular of control and it does not impact functionality. It only impacts how we report spat latency.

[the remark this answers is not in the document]
paulbourelly999: I should leave a TOD to increase this value since 2s is probably to often but I wanted to keep it low for right now to be able to find out if we have any latency spikes that we are not aware of currently during integration testing.

tsc_client_service/src/monitor_desired_phase_plan.cpp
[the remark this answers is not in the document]
adev4a: Wouldn't the next green be after the (green times + yellow duration + red clearance) of all the other phases in the ring?
There is a red_duration parameter in the configuration that captures this

tsc_client_service/include/ntcip_oids.h
```diff
@@ -93,5 +93,14 @@ namespace ntcip_oids {
      * .1 is needed at the end to set the phase control. NTCIP documentation is unclear about why its required or what it represents
      */
     static const std::string PHASE_HOLD_CONTROL = "1.3.6.1.4.1.1206.4.2.1.1.5.1.4.1";
+    /**
+     * @brief Phase Next Status Mask returns a 8 bit value. When a bit = 1, this phase is currently commited to be 
+     * next, when a bit = 0, this phase is NOT commited to be next. The phase commited to be next is determined at
+     * the end of each green interval, except if that decision cannot be made. In this case it will wait until all 
+     * vehicel change and clearance intervales. NOTE:Uses first entry in the phaseStatusGroupTable which is assumed
```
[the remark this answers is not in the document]
paulbourelly999: Will try to update this comment since I see some issues here as well. Although I am not sure I understand your concern here. This description is partially taken from the NTCIP documentation for phaseStatusGroupPhaseNexts. They call it the Phase Next Status Mask.

tsc_client_service/src/monitor_desired_phase_plan.cpp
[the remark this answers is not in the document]
paulbourelly999: This is in a scenario where there are now current greens in the spat. What red duration captures is the time from yellow to the next green given fixed phase sequence for all other phases. This is phase prediction logic we added to essentially mock out fixing phases in sequence. In practice we only fix phases based on the desired phase plan or the next phase at a yellow. Red duration captures a full cycle of greens for every other phase will this will only assume the fixed green that is confirmed to be upcoming.

tsc_client_service/src/monitor_desired_phase_plan.cpp
```diff
@@ -24,22 +27,99 @@ namespace traffic_signal_controller_service
...
-            streets_tsc_configuration::signal_group_configuration sg_config;
-            sg_config.signal_group_id = (uint8_t)sg_id;
-            sg_config.red_clearance = (uint16_t)signal_group_state.red_clearance;
-            sg_config.yellow_change_duration = (uint16_t)signal_group_state.yellow_duration;
-            tsc_state_config_ptr->tsc_config_list.push_back(sg_config);
+    void monitor_desired_phase_plan::fix_upcoming_green(  const std::shared_ptr<signal_phase_and_timing::spat> spat_ptr, 
+            const std::shared_ptr<tsc_state> tsc_state)  const{
+        // Find start time of next green
+        bool is_yellow_change = false;
+        uint64_t start_time_epoch_ms = 0;
```
paulbourelly999, line 54: Leave comment that this is the start time for the fixed green.

tsc_client_service/src/monitor_desired_phase_plan.cpp
```diff
@@ -24,22 +27,99 @@ namespace traffic_signal_controller_service
...
+        uint64_t cur_time_since_epoch = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch()).count();
+        auto state = spat_ptr->get_intersection();
+
+        // Check if current spat includes yellow clearance
+        for (const auto &movement : state.states) {
+            // Found yellow states
+            if ( movement.state_time_speed.front().event_state == signal_phase_and_timing::movement_phase_state::protected_clearance ) {
+                is_yellow_change = true;
+                auto yellow_phase_configuration =tsc_state->get_signal_group_state_map().find(movement.signal_group)->second;
```
paulbourelly999, line 63: Add comment to say the start of the green will be the end of yellow plus red clearance

tsc_client_service/src/monitor_desired_phase_plan.cpp
```diff
@@ -24,22 +27,99 @@ namespace traffic_signal_controller_service
...
+        snmp_response_obj response;
+        response.type = snmp_response_obj::response_type::INTEGER;
+        _snmp_client->process_snmp_request(ntcip_oids::PHASE_STATUS_GROUP_PHASE_NEXT,traffic_signal_controller_service::request_type::GET, response );
+        
+        // Generate Desired Phase Plan with next green fixed.
+        streets_desired_phase_plan::streets_desired_phase_plan one_fixed_green;
+        one_fixed_green.timestamp = cur_time_since_epoch;
+        streets_desired_phase_plan::signal_group2green_phase_timing fixed_green;
+        fixed_green.start_time = start_time_epoch_ms;
+
```
paulbourelly999, line 97: Comment on reading bit masking.

tsc_client_service/src/monitor_desired_phase_plan.cpp
```diff
@@ -24,22 +27,99 @@ namespace traffic_signal_controller_service
...
+            }
+        }
+        if ( fixed_green.signal_groups.size() > 2 ) {
+            throw monitor_desired_phase_plan_exception( "More than two phases were found to be next!");
         }
-        spat_ptr->update_spat_with_candidate_dpp(*desired_phase_plan_ptr, tsc_state_config_ptr);
+        else if ( fixed_green.signal_groups.size() < 1) {
+            throw monitor_desired_phase_plan_exception("No signal groups were found to be next");
+        }
+        uint64_t fixed_green_end_time_epoch_ms = 0;
```
paulbourelly999, line 111: Add comment calculate end time of the next green.

tsc_client_service/include/ntcip_oids.h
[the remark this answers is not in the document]
adev4a: That makes sense, but its confusing to read here. It makes it seem like this description isn't meant for the object that's actually being referenced to.

tsc_client_service/src/monitor_desired_phase_plan.cpp
```diff
@@ -13,9 +16,9 @@ namespace traffic_signal_controller_service
         return desired_phase_plan_ptr;
     }
 
-    void monitor_desired_phase_plan::update_spat_future_movement_events(std::shared_ptr<signal_phase_and_timing::spat> spat_ptr, const std::shared_ptr<tsc_state> tsc_state_ptr) const
+    void monitor_desired_phase_plan::update_spat_future_movement_events(const std::shared_ptr<signal_phase_and_timing::spat> spat_ptr, const std::shared_ptr<tsc_state> tsc_state_ptr) const
```
adev4a, line 19: How can the spat_ptr be constant here? Aren't we modifying it?

tsc_client_service/src/monitor_desired_phase_plan.cpp
```diff
@@ -24,22 +27,121 @@ namespace traffic_signal_controller_service
...
-        // Convert tsc_state to tsc_state_configuration
-        auto tsc_state_config_ptr = std::make_shared<streets_tsc_configuration::tsc_configuration_state>();
-        for (const auto &[sg_id, signal_group_state] : tsc_state_ptr->get_signal_group_state_map())
-        {
-            streets_tsc_configuration::signal_group_configuration sg_config;
-            sg_config.signal_group_id = (uint8_t)sg_id;
-            sg_config.red_clearance = (uint16_t)signal_group_state.red_clearance;
-            sg_config.yellow_change_duration = (uint16_t)signal_group_state.yellow_duration;
-            tsc_state_config_ptr->tsc_config_list.push_back(sg_config);
+    void monitor_desired_phase_plan::fix_upcoming_green(  const std::shared_ptr<signal_phase_and_timing::spat> spat_ptr,
```
adev4a, line 50: Same comment here, why is the spat_ptr constant?

tsc_client_service/src/monitor_desired_phase_plan.cpp
```diff
@@ -24,22 +27,121 @@ namespace traffic_signal_controller_service
...
+        // Check if current spat includes yellow clearance
+        for (const auto &movement : state.states) {
+            // Found yellow states
+            if ( movement.state_time_speed.front().event_state == signal_phase_and_timing::movement_phase_state::protected_clearance ) {
+                is_yellow_change = true;
+                auto yellow_phase_configuration =tsc_state->get_signal_group_state_map().find(movement.signal_group)->second;
+                // Start time of upcoming green will be end of yellow + red clearance 
+                uint64_t local_start_time_epoch = movement.state_time_speed.front().timing.get_epoch_min_end_time() + yellow_phase_configuration.red_clearance;
+                if ( local_start_time_epoch > start_time_epoch_ms) {
+                    // If there are multiple movement groups in yellow, start time will be the largest end of yellow + red clearance time
```
adev4a, line 69: Shouldn't we be looking at the smallest yellow + red_clearance to know which phase is green next?

tsc_client_service/src/monitor_desired_phase_plan.cpp
paulbourelly999: It is a const pointer so the object it is pointing to can not be reassign.

tsc_client_service/src/monitor_desired_phase_plan.cpp
paulbourelly999: It is a const pointer meaning the object it is pointing to can't be reassign. The underlying object is not constant. We can't assign the pointer to spat 1 and then reassign it using make_shared to another thing

tsc_client_service/src/monitor_desired_phase_plan.cpp
paulbourelly999: No our limitation is the largest. So if two signal groups in a movement group have different red clearance end times, the first time we can grant the next conflicting phase green when both red clearances are over. Technically we need to check barrier stuff here too in order for this to actually work but we don't do that yet anywhere else. Right now our system can really only handle concurrent phases that have equal red clearance and yellow change configured.

adev4a: Can we add that as a comment? or TODO? Just so it can be tracked down and updated later

paulbourelly999: I am not sure I understand. The TODO is for out entire system not just this line or class. Essentially we account for different phases having different configuration in terms of minimum green , yellow change, and red clearance only partially here and in other places in the code. We account for differences in these configurations but not for the ring and barrier setup. So for two concurrent phases that have different red clearance times but are next to the barrier this logic is sufficient but for two phases that are not next to the barrier or if only one of the phases is next to the barrier this prediction logic will be inaccurate unless we exercise TSC NTCIP calls to make it accurate (essentially hold an all red for the difference). This is not appropriate in my opinion as an inline todo because it is a functionality improvement that would need to be made in several services and requires some design choices.

adev4a (approved):

Saeid-Solei (approved):

Issue #260, opened 2022-11-18
Max Green Overwrite Issue in dpp_generator Library 


Saeid-Solei:
### Types of Issue

- [x] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

To optimize SPaT and adding a new fixed future movement group, the desired_phase_plan_generator library first finds the required green duration for dissipating the queue from each entry lane. This required green duration is calculated based on the estimated entering times (ETs) of the vehicles that are part of a queue. Basically, the required green duration for an entry lane will be long enough to cover the estimated ETs of the vehicles in the queue from the entry lane. 
However, if a required green duration is greater the max green, the required green duration will be set to max green. The issue is, it is possible that the estimated ET of the first vehicle in the queue is more than max green duration far away from the beginning of this new green. In this case, after setting the required green duration to max green, it will not serve any vehicle and will not cover any estimated ET. This means that it is possible to have green phases with max duration that does not serve any vehicle. This shall not happen!

### Carma streets version where this issue was discovered

4.2.0

### Expected behavior
Expect to not have green phases with max duration that does not serve any vehicle.

### Actual behavior

It might have green phases with max duration that does not serve any vehicle.

PR #261, opened 2022-11-18
Max Green Overwrite Issue in dpp_generator Library

Saeid-Solei:
# PR Details
## Description

To optimize SPaT and adding a new fixed future movement group, the desired_phase_plan_generator library first finds the required green duration for dissipating the queue from each entry lane. This required green duration is calculated based on the estimated entering times (ETs) of the vehicles that are part of a queue. Basically, the required green duration for an entry lane will be long enough to cover the estimated ETs of the vehicles in the queue from the entry lane.
However, if a required green duration is greater than the max green, the required green duration will be set to max green. The issue is, it is possible that the estimated ET of the first vehicle in the queue is more than max green duration far away from the beginning of this new green. In this case, after setting the required green duration to max green, it will not serve any vehicle and will not cover any estimated ET. This means that it is possible to have green phases with max duration that does not serve any vehicle. This shall not happen!

## Related Issue

https://github.com/usdot-fhwa-stol/carma-streets/issues/260

## How Has This Been Tested?

Unit Tests

## Types of changes

## Checklist:


streets_utils/streets_signal_optimization/src/streets_desired_phase_plan_generator.cpp
```diff
@@ -357,8 +357,15 @@ namespace streets_signal_optimization {
                 uint64_t et_rounded = uint64_t(std::ceil(double(last_ev_in_queue.et)/100.0) * 100);
                 uint64_t green_end_compared_with_min_green = std::max(et_rounded + config.et_inaccuracy_buffer + config.final_green_buffer, tbd_start + config.min_green);
                 uint64_t green_end = std::min(green_end_compared_with_min_green, tbd_start + config.max_green);
-
-                green_end_per_entry_lane.try_emplace(entry_lane_id, green_end);   
+                // in case the green_end is overwritten by max_green, check if there is any ET within the new green duration.
+                // if not, skip this entry lane, assuming there is no queue that can be served.
+                if (green_end < green_end_compared_with_min_green && evs_in_lane.front().et >= green_end - config.final_green_buffer) {
+                    SPDLOG_DEBUG("The required green duration for dissipation the queue in entry lane " + std::to_string(entry_lane_id) + " is overwritten by max_green which cannot serve any vehicle!");
```
paulbourelly999, line 363: Lets make this a warn log and include end_green_max vs ET

Saeid-Solei: Done!

streets_utils/streets_signal_optimization/src/streets_desired_phase_plan_generator.cpp
```diff
@@ -357,8 +357,20 @@ namespace streets_signal_optimization {
                 uint64_t et_rounded = uint64_t(std::ceil(double(last_ev_in_queue.et)/100.0) * 100);
                 uint64_t green_end_compared_with_min_green = std::max(et_rounded + config.et_inaccuracy_buffer + config.final_green_buffer, tbd_start + config.min_green);
                 uint64_t green_end = std::min(green_end_compared_with_min_green, tbd_start + config.max_green);
-
-                green_end_per_entry_lane.try_emplace(entry_lane_id, green_end);   
+                // in case the green_end is overwritten by max_green, check if there is any ET within the new green duration.
+                // if not, skip this entry lane, assuming there is no queue that can be served.
+                if (green_end < green_end_compared_with_min_green && evs_in_lane.front().et >= green_end - config.final_green_buffer) {
```
paulbourelly999, line 362: Last small request, just to make sure we don't have some really unintended behavior. Can we make sure we trigger this condition in a test case and confirm the outcome is what we expect.

Saeid-Solei: Done! I found a bug in the other test cases as well. Apparently, someone changed the vehicle length assuming it is in meter now. But it is assumed to be in centimeter still. Also, the min_green values where not correct.

paulbourelly999 (approved):

Issue #267, opened 2022-11-22
TSC Service hangs before consuming SPAT


paulbourelly999:
### Types of Issue

- [x] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

TSC Service currently hangs in the last part of the initialization. This is because we are passing the `control_tsc_state` pedestrian phases  on initialization instead of vehicle phases.
### Expected behavior
TSC Service runs and produces SPaT messages
### Actual behavior

TSC Service hangs on initialization.
### Steps to reproduce the actual behavior

1. Deploy CARMA-Streets

Issue #265, opened 2022-11-22
Enable SPaT OID is not reachable with McCain controllers


paulbourelly999:
### Types of Issue

- [x] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

It seems the McCain Traffic Signal Controllers either don't expose the enable SPaT OID or are supporting a different version of the NTCIP 1202 standard since when the TSC Service tries to enable SPaT with the econolite controllers this works but for the McCain controllers it causes the TSC Service to fail.

**OID that works for Econolite 1.3.6.1.4.1.1206.3.5.2.9.44.1.0**
**OID that might work for McCain (still needs to be tested) 1.3.6.1.4.1.1206.4.2.1.16.4 or maybe 1.3.6.1.4.1.1206.4.2.1.16.5.1.1**
### Expected behavior
Remove Enable Spat call from TSC Service Initialization and require it as TSC configuration 
### Actual behavior
It seems the McCain Traffic Signal Controllers either don't expose the enable SPaT OID or are supporting a different version of the NTCIP 1202 standard since when the TSC Service tries to enable SPaT with the econolite controllers this works but for the McCain controllers it causes the TSC Service to fail.
### Steps to reproduce the actual behavior

1. Deploy CARMA-Streets with McCain TSC

Issue #258, opened 2022-11-18
Failing Unit Test 


paulbourelly999:
### Types of Issue

- [x] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

Unit test failure (https://app.circleci.com/pipelines/github/usdot-fhwa-stol/carma-streets/1425/workflows/da467df0-e5fb-482d-94e7-21583c859201/jobs/2204)
### Carma streets version where this issue was discovered

Latest
### Expected behavior
All unit test pass.
### Actual behavior

One unit test is failing 
### Steps to reproduce the actual behavior

1. Check CircleCI workflow for builds (https://app.circleci.com/pipelines/github/usdot-fhwa-stol/carma-streets/1425/workflows/da467df0-e5fb-482d-94e7-21583c859201/jobs/2204)

PR #269, opened 2022-12-01
Tsmo uc3 integration

paulbourelly999:
# PR Details
## Description

Multiple Bug fixes for TSMO UC 3 Integration Testing.
## Related Issue
#277 
#275
#273 
#271 
#267 
#265 
#258
#282

## Motivation and Context

Fix bugs discovered during integration testing

## How Has This Been Tested?

Deployed and tested at West Intersection at THFRC during TSMO UC3 Integration testing
## Types of changes

## Checklist:

### DIFF
--- a/collect_service_logs.sh
+++ b/collect_service_logs.sh
@@ -5,7 +5,7 @@
 #  License for the specific language governing permissions and limitations under
 #  the License.
 
-# script collects all service logs files and zips them into an archive.
+# script collects all service logs files and zips them into an archive and then removes the source files.
 set -e
 zip -r service_logs.zip ./scheduling_service/logs ./intersection_model/logs ./tsc_client_service/logs ./message_services/logs ./signal_opt_service/logs
 sudo rm ./scheduling_service/logs/* ./intersection_model/logs/* ./tsc_client_service/logs/* ./message_services/logs/* ./signal_opt_service/logs/*
--- a/collect_service_logs.sh
+++ b/collect_service_logs.sh
@@ -1,0 +1,11 @@
+# 
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#  License for the specific language governing permissions and limitations under
+#  the License.
+
+# script collects all service logs files and zips them into an archive.
+set -e
+zip -r service_logs.zip ./scheduling_service/logs ./intersection_model/logs ./tsc_client_service/logs ./message_services/logs ./signal_opt_service/logs
+sudo rm ./scheduling_service/logs/* ./intersection_model/logs/* ./tsc_client_service/logs/* ./message_services/logs/* ./signal_opt_service/logs/*
--- a/signal_opt_service/src/signal_opt_service.cpp
+++ b/signal_opt_service/src/signal_opt_service.cpp
@@ -107,7 +107,6 @@ namespace signal_opt_service
 
     void signal_opt_service::consume_spat(const std::shared_ptr<kafka_clients::kafka_consumer_worker> spat_consumer, 
                                             const std::shared_ptr<signal_phase_and_timing::spat> _spat_ptr ) const{
-        SPDLOG_INFO("Consuming SPat.");
         while( spat_consumer->is_running()) {
             const std::string payload = spat_consumer->consume(1000); 
             if (!payload.empty()) {
@@ -121,7 +120,6 @@ namespace signal_opt_service
 
     void signal_opt_service::consume_vsi(const std::shared_ptr<kafka_clients::kafka_consumer_worker> vsi_consumer,
                                          const std::shared_ptr<streets_vehicles::vehicle_list> _vehicle_list_ptr) const{
-        SPDLOG_INFO("Consuming VSI.");
         while( vsi_consumer->is_running()) {
             const std::string payload = vsi_consumer->consume(1000); 
             if (!payload.empty()) {
@@ -169,7 +167,7 @@ namespace signal_opt_service
         bool new_dpp_generated = false;
 
         while ( dpp_producer->is_running() ) {
-            SPDLOG_INFO("Signal Optimization iteration!");
+            SPDLOG_DEBUG("Signal Optimization iteration!");
             if ( !_vehicle_list_ptr->get_vehicles().empty() ) {
                 streets_desired_phase_plan::streets_desired_phase_plan spat_dpp;
                 try
--- a/signal_opt_service/src/signal_opt_service.cpp
+++ b/signal_opt_service/src/signal_opt_service.cpp
@@ -166,6 +166,7 @@ namespace signal_opt_service
         
         int prev_future_move_group_count = 0;
         int current_future_move_group_count = 0;
+        bool new_dpp_generated = false;
 
         while ( dpp_producer->is_running() ) {
             SPDLOG_INFO("Signal Optimization iteration!");
@@ -182,13 +183,21 @@ namespace signal_opt_service
                     break;
                 }
                 auto current_timestamp = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch()).count();
-
-                current_future_move_group_count = static_cast<int>(spat_dpp.desired_phase_plan.size());
-                SPDLOG_INFO("Current movement groups represented in SPAT : {0}!", spat_dpp.desired_phase_plan.size());
-                SPDLOG_INFO("Current green ends at time {0} and current timestamp is {1}", spat_dpp.desired_phase_plan.front().end_time, current_timestamp );
-                auto time_to_yellow = spat_dpp.desired_phase_plan.front().end_time - current_timestamp;
                 SPDLOG_INFO("Time to yellow is : {0}", time_to_yellow );
                 SPDLOG_INFO("Base DPP is {0}", spat_dpp.toJson());
+                current_future_move_group_count = static_cast<int>(spat_dpp.desired_phase_plan.size());
+                if (new_dpp_generated) {
+                    SPDLOG_INFO("Current number of movement groups represented in SPAT : {0}!", prev_future_move_group_count);
+                    if (current_future_move_group_count > prev_future_move_group_count) {
+                        new_dpp_generated = false;
+                    }
+                    else {
+                        SPDLOG_WARN("Skipping SO iteration. Previously sent DPP is not yet reflected in SPaT!");
+                        continue;
+                    }
+                }
+                auto time_to_yellow = spat_dpp.desired_phase_plan.front().end_time - current_timestamp;
+                
                 if ( time_to_yellow <= _time_to_yellow ) {
                     SPDLOG_INFO("Checking fixed phases");
                     /**
@@ -196,7 +205,7 @@ namespace signal_opt_service
                      * and it is less than or equal to the desired numNo vehicles presenter of future movement groups, run streets_signal_optimization
                      * libraries to get optimal desired_phase_plan.
                     */
-                    if (current_future_move_group_count != prev_future_move_group_count || current_future_move_group_count <= dpp_config.desired_future_move_group_count ) {
+                    if (current_future_move_group_count <= dpp_config.desired_future_move_group_count ) {
                         SPDLOG_INFO("The number of future movement groups in the spat is updated from {0} to {1}.", 
                                                                     prev_future_move_group_count, current_future_move_group_count);
                         // Current movement group count includes current fix momvement group.
@@ -212,6 +221,8 @@ namespace signal_opt_service
                             std::string msg_to_send = optimal_dpp.toJson();
                             /* produce the optimal desired phase plan to kafka */
                             dpp_producer->send(msg_to_send);
+                            prev_future_move_group_count = current_future_move_group_count;
+                            new_dpp_generated = true;
                         }
                         
                     }
@@ -227,7 +238,6 @@ namespace signal_opt_service
                 SPDLOG_DEBUG("No vehicles present");
  
             }
-            prev_future_move_group_count = current_future_move_group_count;
             std::this_thread::sleep_for(std::chrono::milliseconds(_so_sleep_time));
         }
         SPDLOG_WARN("Stopping desired phase plan producer thread!");
--- a/signal_opt_service/src/signal_opt_service.cpp
+++ b/signal_opt_service/src/signal_opt_service.cpp
@@ -143,6 +143,7 @@ namespace signal_opt_service
                     SPDLOG_CRITICAL("Failed to update TSC Configuration with update {0}!", payload);
                 }else {
                     // Stop consumer after receiving tsc_configuration information
+                    SPDLOG_INFO("TSC Configuration sucessfully loaded : \n{0}", payload );
                     tsc_config_consumer->stop();
                     break;
                 }
--- a/signal_opt_service/src/signal_opt_service.cpp
+++ b/signal_opt_service/src/signal_opt_service.cpp
@@ -185,9 +185,11 @@ namespace signal_opt_service
 
                 current_future_move_group_count = static_cast<int>(spat_dpp.desired_phase_plan.size());
                 SPDLOG_INFO("Current movement groups represented in SPAT : {0}!", spat_dpp.desired_phase_plan.size());
-
+                SPDLOG_INFO("Current green ends at time {0} and current timestamp is {1}", spat_dpp.desired_phase_plan.front().end_time, current_timestamp );
                 auto time_to_yellow = current_timestamp - spat_dpp.desired_phase_plan.front().end_time;
+                SPDLOG_INFO("Time to yellow is : {0}", time_to_yellow );
                 if ( time_to_yellow <= _time_to_yellow ) {
+                    SPDLOG_INFO("Checking fixed phases");
                     /**
                      * If the number of future movement groups in the spat has changed compared to the previous step
                      * and it is less than or equal to the desired numNo vehicles presenter of future movement groups, run streets_signal_optimization
--- a/signal_opt_service/src/signal_opt_service.cpp
+++ b/signal_opt_service/src/signal_opt_service.cpp
@@ -186,7 +186,7 @@ namespace signal_opt_service
                 current_future_move_group_count = static_cast<int>(spat_dpp.desired_phase_plan.size());
                 SPDLOG_INFO("Current movement groups represented in SPAT : {0}!", spat_dpp.desired_phase_plan.size());
                 SPDLOG_INFO("Current green ends at time {0} and current timestamp is {1}", spat_dpp.desired_phase_plan.front().end_time, current_timestamp );
-                auto time_to_yellow = current_timestamp - spat_dpp.desired_phase_plan.front().end_time;
+                auto time_to_yellow = spat_dpp.desired_phase_plan.front().end_time - current_timestamp;
                 SPDLOG_INFO("Time to yellow is : {0}", time_to_yellow );
                 SPDLOG_INFO("Base DPP is {0}", spat_dpp.toJson());
                 if ( time_to_yellow <= _time_to_yellow ) {
--- a/signal_opt_service/src/signal_opt_service.cpp
+++ b/signal_opt_service/src/signal_opt_service.cpp
@@ -284,6 +284,24 @@ namespace signal_opt_service
             for ( const auto &mg : _groups->groups) {
                 SPDLOG_DEBUG("Signal Group {0} and Signal Group {1}.", mg.signal_groups.first, mg.signal_groups.second );
             }
+            // Remove configured signal groups to ignore them
+            auto mg_itr = _groups->groups.begin();
+            while ( mg_itr != _groups->groups.end() ) {
+                bool erased = false;
+                for (const auto &ignore : ignore_signal_groups) {
+                    if ( mg_itr->signal_groups.first == ignore && mg_itr->signal_groups.second == 0) {
+                        SPDLOG_WARN("Removing movement group with signal groups {0} , {0}", mg_itr->signal_groups.first, mg_itr->signal_groups.second);
+                        mg_itr = _groups->groups.erase(mg_itr);
+                        erased = true;
+                    }
+                }
+                if (!erased) {
+                    mg_itr++;
+                }
+
+            }
+            
+            
             
         }
     }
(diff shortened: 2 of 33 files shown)


signal_opt_service/src/signal_opt_service.cpp
```diff
@@ -161,50 +164,90 @@ namespace signal_opt_service
...
+                SPDLOG_INFO("Time to yellow is : {0}", time_to_yellow );
+                if ( time_to_yellow <= _time_to_yellow ) {
+                    SPDLOG_INFO("Checking fixed phases");
+                    /**
+                     * If the number of future movement groups in the spat has changed compared to the previous step
+                     * and it is less than or equal to the desired numNo vehicles presenter of future movement groups, run streets_signal_optimization
+                     * libraries to get optimal desired_phase_plan.
+                    */
+                    if (current_future_move_group_count <= dpp_config.desired_future_move_group_count ) {
+                        SPDLOG_INFO("The number of future movement groups in the spat is updated from {0} to {1}.",
```
Saeid-Solei, line 218: This Log is not correct anymore. Please remove or change it.

signal_opt_service/src/signal_opt_service.cpp
```diff
@@ -161,50 +164,90 @@ namespace signal_opt_service
...
+                            std::string msg_to_send = optimal_dpp.toJson();
+                            /* produce the optimal desired phase plan to kafka */
+                            dpp_producer->send(msg_to_send);
+                            prev_future_move_group_count = current_future_move_group_count;
+                            new_dpp_generated = true;
+                        }
+                        
                     }
+                    else {
+                        SPDLOG_INFO("The number of future movement groups in the spat did not change from the previous check.");
```
Saeid-Solei, line 239: Please change this one as well. It probably can be:
The number of future movement groups in spat is greater than or equal to the desired number future movement groups!

signal_opt_service/test/test_signal_opt_service.cpp
```diff
@@ -597,4 +596,100 @@ TEST(signal_opt_service, populate_movement_group_nema) {
...
+
+    signal_opt_service::signal_opt_service service;
+    // Setup movement groups
+    auto movement_groups = std::make_shared<streets_signal_optimization::movement_groups>();
+    // Setup movement groups
+    streets_signal_optimization::movement_group mg1;
+    mg1.signal_groups = {1,5};
+    streets_signal_optimization::movement_group mg2;
+    mg2.signal_groups = {2,5};
+    streets_signal_optimization::movement_group mg3;
```
Saeid-Solei, line 651: Given the implemented logic, if signal group 2 is removed, then we would have both movement groups {1, 5} and {5, 0}. In this case {5, 0} is a subset of {1, 5}. So we technically don't need it. 

An improved logic would be ignoring the signal groups when creating the movement groups, or adding a logic to remove the movement groups that are a subset of other movement groups. But I believe this is an enhancement and the current logic might not cause any issue.

streets_utils/streets_signal_phase_and_timing/src/models/time_change_details.cpp
```diff
@@ -110,7 +110,16 @@ namespace signal_phase_and_timing {
         auto system_time = std::chrono::system_clock::now();
```
Saeid-Solei, line 110: This comment is for the convert_hour_tenth_secs2epoch_ts method!

Do we still have a plan to include the counted hours as an input here? (For the case where we want to get epoch start time or end time from spat)

tsc_client_service/include/control_tsc_state.h
```diff
@@ -82,14 +87,15 @@ namespace traffic_signal_controller_service
              * @param phase_to_signal_group_map A map from pedestrian phases in the traffic signal controller to the signal group ids
              * @param desired_phase_plan Pointer to the desired phase plan.
```
Saeid-Solei, line 88: Please update the parameter descriptions.

tsc_client_service/include/monitor_desired_phase_plan.h
```diff
@@ -61,7 +71,11 @@ namespace traffic_signal_controller_service
          * @param spat_ptr shared pointer to the current spat information.
          * @param tsc_state_ptr shared pointer to the tsc state information including tsc phase configuration information.
          */
-        void fix_upcoming_green(const std::shared_ptr<signal_phase_and_timing::spat> spat_ptr, const std::shared_ptr<tsc_state> tsc_state_ptr) const;
+        void fix_upcoming_green(const std::shared_ptr<signal_phase_and_timing::spat> spat_ptr, const std::shared_ptr<tsc_state> tsc_state_ptr) ;
+
+        void fix_upcoming_yell_red(const std::shared_ptr<signal_phase_and_timing::spat> spat_ptr,
```
Saeid-Solei, line 76: Description is missing!

tsc_client_service/include/tsc_service.h
```diff
@@ -129,15 +132,18 @@ namespace traffic_signal_controller_service {
              */
 
             bool initialize_kafka_producer( const std::string &bootstap_server, const std::string &producer_topic, 
-                    std::shared_ptr<kafka_clients::kafka_producer_worker>& producer);
+                    std::shared_ptr<kafka_clients::kafka_producer_worker> &producer);
             /**
              * @brief Initialize Kafka Desired phase plan consumer. 
              * @param bootstap_server for CARMA-Streets Kafka broker.
              * @param desired_phase_plan_consumer_topic name of topic to produce to.
```
Saeid-Solei, line 139: Please update the input parameter descriptions.

tsc_client_service/src/control_tsc_state.cpp
```diff
@@ -136,5 +137,23 @@ namespace traffic_signal_controller_service
         }
     }
 
+    std::string snmp_cmd_struct::get_cmd_info() const{
+
+        std::string command_type;
+        if(control_type_ == control_type::Hold){
+            command_type = "Hold";
+        }
+        else{
```
Saeid-Solei, line 146: Shall we protect against not setting the command to Omit if the control_type is something else than Hold and Omit?

tsc_client_service/src/monitor_desired_phase_plan.cpp
```diff
@@ -31,24 +34,55 @@ namespace traffic_signal_controller_service
...
+                if (last_green_served.empty() ) {
+                    throw monitor_desired_phase_plan_exception("No information on previous green!");
+                }
                 fix_upcoming_green(spat_ptr, tsc_state_ptr);
+            } else {
+                // Track which greens were served last and update this value.
+                if ( last_green_served.empty() || green_phases_present.front().signal_group != last_green_served.front().signal_group || 
+                    green_phases_present.back().signal_group != last_green_served.front().signal_group) {
+                    last_green_served = green_phases_present;
+                    SPDLOG_DEBUG("Last served greens set to {0} and {0}!",
```
Saeid-Solei, line 56: "Last served greens set to {0} and {1}!"

tsc_client_service/src/monitor_desired_phase_plan.cpp
```diff
@@ -63,24 +97,34 @@ namespace traffic_signal_controller_service
...
+                        last_green_served.size() == 1 ? 0 : last_green_served.back().signal_group );
+                }
             }
         }
         // If it does not include a green and does not include a yellow assume it is in all red clearance 
         if ( !is_yellow_change ) {
             for (const auto &movement : state.states) {
                 // Found red states
-                if ( movement.state_time_speed.front().event_state == signal_phase_and_timing::movement_phase_state::stop_and_remain ) {
+                if ( (movement.signal_group == last_green_served.front().signal_group
```
Saeid-Solei, line 119: What if last_green_served is empty (if the service starts on red clearance)? Shouldn't we have a similar check to the one for the yellow change?

tsc_client_service/src/monitor_desired_phase_plan.cpp
```diff
@@ -140,8 +184,37 @@ namespace traffic_signal_controller_service
...
+        uint64_t cur_time_since_epoch = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch()).count();
+        auto state = spat_ptr->get_intersection();
+        // Get green phase configuration to get min green.
+        auto green_phase_config =tsc_state->get_signal_group_state_map().find(green_phases.front().signal_group)->second;
+
+            
+        // Generate Desired Phase Plan with next green fixed.
+        one_fixed_green.timestamp = cur_time_since_epoch;
+        fixed_green.start_time = green_phases.front().state_time_speed.front().timing.get_epoch_start_time();
+        fixed_green.end_time =  fixed_green.start_time + green_phase_config.min_green;
```
Saeid-Solei, line 215: In a case where 2 signal groups are green at the same time, shall min_green be the max of min_greens of both signal groups? Or is it safe to assume that all min_greens are the same?

Saeid-Solei, line 215: I remember one of the issues we found is the wrong epoch_start_time and epoch_min_end_time at the hour change. I see here that the min_end_time is purely calculated based on the start_time. And I saw that you have made changes to set the start time to 0 if the hour is changed. This means that the start_time and min_end_time might have a shift at the hour change, at least until the end of the current phase. In my opinion, this might still cause some issues!

collect_service_logs.sh
```diff
@@ -0,0 +1,11 @@
+# 
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#  License for the specific language governing permissions and limitations under
+#  the License.
+
+# script collects all service logs files and zips them into an archive and then removes the source files.
+set -e
+zip -r service_logs.zip ./scheduling_service/logs ./intersection_model/logs ./tsc_client_service/logs ./message_services/logs ./signal_opt_service/logs
```
dan-du-car, line 10: maybe add a navigation to the directory where this zip command runs?

streets_utils/streets_signal_phase_and_timing/src/models/time_change_details.cpp
paulbourelly999: The SPaT fix for using message time stamp vs current time should be completely independent from any changes in this PR. Not sure if we still plan on tackling the SPaT issue but I don't think so

signal_opt_service/src/signal_opt_service.cpp
```diff
@@ -107,6 +107,7 @@ namespace signal_opt_service
 
     void signal_opt_service::consume_spat(const std::shared_ptr<kafka_clients::kafka_consumer_worker> spat_consumer, 
                                             const std::shared_ptr<signal_phase_and_timing::spat> _spat_ptr ) const{
+        SPDLOG_INFO("Consuming SPat.");
```
dan-du-car, line 110: maybe a debug level?

signal_opt_service/src/signal_opt_service.cpp
```diff
@@ -161,50 +164,94 @@ namespace signal_opt_service
...
-            try
-            {
-                spat_dpp = _so_processing_worker_ptr->convert_spat_to_dpp(_spat_ptr, _movement_groups_ptr);
-            }
-            catch(const streets_signal_optimization::streets_desired_phase_plan_generator_exception &ex)
-            {
-                SPDLOG_ERROR("dpp_generator_ptr is not initialized! : {0} ", ex.what());
-            }
-            current_future_move_group_count = static_cast<int>(spat_dpp.desired_phase_plan.size());
+            SPDLOG_INFO("Signal Optimization iteration!");
```
dan-du-car, line 172: maybe debug level? The same for all the steps INFO statements below.

tsc_client_service/src/control_tsc_state.cpp
paulbourelly999: This is just a to string method. Currently we only support Omit and Hold

tsc_client_service/src/monitor_desired_phase_plan.cpp
paulbourelly999: This logic keeps track of the last green that was served. This is used during red clearance to determine which phase was just green and therefor is currently in red clearance

tsc_client_service/src/monitor_desired_phase_plan.cpp
paulbourelly999: If last green is empty the service will not produce spat until the first green is served, at which point last green will never be empty again. For yellow change, the phase currently in yellow change is the one that was green last so this check is unnecessary

tsc_client_service/src/monitor_desired_phase_plan.cpp
paulbourelly999: Our current system will fail if these are different so it is a current simplifying assumption of our system that these will be the same.

paulbourelly999: We will have issues at the start of the hour regardless of these assumptions because start time can not be accurately represented in SPaT format at the hour transition

paulbourelly999 (commented):
Addressed all PR comments

signal_opt_service/test/test_signal_opt_service.cpp
```diff
@@ -597,4 +601,100 @@ TEST(signal_opt_service, populate_movement_group_nema) {
...
+    streets_signal_optimization::movement_group mg5;
+    mg5.signal_groups = {11,0};
+    movement_groups->groups.push_back(mg1);
+    movement_groups->groups.push_back(mg2);
+    movement_groups->groups.push_back(mg3);
+    movement_groups->groups.push_back(mg4);
+    movement_groups->groups.push_back(mg5);
+
+    // Setup Ignore Signal Groups
+    std::vector<uint> ignore_signal_groups;
```
dan-du-car, line 627: Do we need to ignore it first then remove it? Are these remove and ignore connected?

streets_utils/streets_signal_phase_and_timing/src/models/time_change_details.cpp
```diff
@@ -110,7 +110,16 @@ namespace signal_phase_and_timing {
...
         auto duration = system_time.time_since_epoch();
         auto hours_since_epoch = std::chrono::duration_cast<std::chrono::hours>(duration).count();
-        auto tenth_seconds_from_current_hour = (epoch_time_ms - (hours_since_epoch * HOUR_TO_SECONDS * SECOND_TO_MILLISECONDS))/100;
+        auto hours_since_epoch_ms = hours_since_epoch * HOUR_TO_SECONDS * SECOND_TO_MILLISECONDS;
+        uint64_t tenth_seconds_from_current_hour;
+        if ( hours_since_epoch_ms > epoch_time_ms ) {
```
dan-du-car, line 115: hours_since_epoch_ms is greater than epoch_time_ms. Does that mean hours is next or previous hour?

streets_utils/streets_signal_phase_and_timing/src/models/time_change_details.cpp
```diff
@@ -110,7 +110,16 @@ namespace signal_phase_and_timing {
...
+            SPDLOG_WARN("Epoch time provided {0} is smaller than the current UTC hour time {1}! SPaT cannot represent a time from a previous hour. Approximating time as beginning of current hour."
+                , epoch_time_ms, hours_since_epoch_ms);
+            tenth_seconds_from_current_hour = 0;
```
dan-du-car, line 118: Why do we decide to approximate it to the beginning of current hour?

tsc_client_service/include/control_tsc_state.h
```diff
@@ -82,14 +87,15 @@ namespace traffic_signal_controller_service
...
-            explicit control_tsc_state(std::shared_ptr<snmp_client> snmp_client,const std::unordered_map<int, int>& signal_group_to_phase_map);
+            explicit control_tsc_state(std::shared_ptr<snmp_client> snmp_client,std::shared_ptr<tsc_state> _tsc_state);
             
             /** 
              * @brief Method to update the queue of tsc_control
              * @param desired_phase_plan Pointer to the desired phase plan.
              * @param tsc_command_queue Queue of snmp commands to set HOLD and OMIT on the traffic signal controller
              **/
-            void update_tsc_control_queue(std::shared_ptr<streets_desired_phase_plan::streets_desired_phase_plan> desired_phase_plan, std::queue<snmp_cmd_struct>& tsc_command_queue);
+            void update_tsc_control_queue(std::shared_ptr<streets_desired_phase_plan::streets_desired_phase_plan> desired_phase_plan,
```
dan-du-car, line 97: It is kind of confusing that the method name says update meaning update the status of this instance but it is  a const function.

dan-du-car: The same comments for the create_* below.

tsc_client_service/test/test_tsc_service.cpp
```diff
@@ -2,120 +2,429 @@
...
+            // Test get max channels
+            snmp_response_obj max_channels_in_tsc;
+            max_channels_in_tsc.val_int = 16;
+            max_channels_in_tsc.type = snmp_response_obj::response_type::INTEGER;
+
+            request_type request_type = request_type::GET;
+
+            EXPECT_CALL( *mock_snmp, process_snmp_request(ntcip_oids::MAX_CHANNELS, request_type , _) ).WillRepeatedly(testing::DoAll(
+                SetArgReferee<2>(max_channels_in_tsc), 
+                Return(true)));
```
dan-du-car, line 49: Comments about what this return true means?

tsc_client_service/test/test_tsc_service.cpp
```diff
@@ -2,120 +2,429 @@
...
+                SetArgReferee<2>(seq_data), 
+                Return(true)));
+
+            seq_data.val_string = {char(5),char(6), char(7), char(8)};
+            std::string seq_data_ring2_oid = ntcip_oids::SEQUENCE_DATA + "." + "1" + "." + std::to_string(2);
+            EXPECT_CALL(*mock_snmp, process_snmp_request(seq_data_ring2_oid, request_type , _) ).WillRepeatedly(testing::DoAll(
+                SetArgReferee<2>(seq_data), 
+                Return(true)));
+
+            seq_data.val_string = {};
```
dan-du-car, line 72: Can we add brief inline command about each process_snmp_request call?

signal_opt_service/src/signal_opt_service.cpp
Saeid-Solei: I can see you have changed it already. Thanks!

streets_utils/streets_signal_phase_and_timing/src/models/time_change_details.cpp
Saeid-Solei: Ok. Just wanted to know if the fix for using current time vs message time stamp is included here or not. If not, we just want to know the fix is not added yet.

tsc_client_service/include/tsc_service.h
Saeid-Solei: The desired_phase_plan_consumer_topic also must be removed from @param.

tsc_client_service/src/monitor_desired_phase_plan.cpp
Saeid-Solei: I get the logic. My question is that whether the second {} in SPDLOG_DEBUG shall include 1 or 0?

tsc_client_service/src/monitor_desired_phase_plan.cpp
Saeid-Solei: If I understood correctly, the service will not produce spat if it starts on red clearance, but will produce spat if it starts on green or yellow change?

tsc_client_service/src/monitor_desired_phase_plan.cpp
Saeid-Solei: I see. Just for my information, when you say system will fail, does it mean that the service does not allow min_green params for different signal groups to be different and it will throw exception if they are not the same. Is that right?

Saeid-Solei: So I guess we shall try not to test at the hour change!

tsc_client_service/src/monitor_desired_phase_plan.cpp
paulbourelly999: Oh sorry, yeah you are right I updated it.

tsc_client_service/src/monitor_desired_phase_plan.cpp
paulbourelly999: yes, it will not produce spat starting on red clearance until a yellow or green phase is active

tsc_client_service/src/monitor_desired_phase_plan.cpp
paulbourelly999: We do not know exactly how the system will perform if min green is different for concurrent signal groups. It may through some exceptions, it may just incorrectly predict spat, it may core dump somewhere. Since this is a simplifying assumption of our UC I am not sure how robust we need to be in addressing when the assumptions are not true, but my inclination is not at all.

signal_opt_service/src/signal_opt_service.cpp
paulbourelly999: Removed

signal_opt_service/src/signal_opt_service.cpp
paulbourelly999: Removed

signal_opt_service/test/test_signal_opt_service.cpp
paulbourelly999: The ignore signal group vector is just a list of signal groups we will remove from the movement groups sent by the tsc_config_state json from the tsc_service

streets_utils/streets_signal_phase_and_timing/src/models/time_change_details.cpp
paulbourelly999: if we are trying to set an epoch time on a SPaT message that is less than the current hour, it can not be represented by the message structure so we approximate it as the earliest time we can set which is the beginning of the current hour. The condition checks if the time set is from a previous hour.

streets_utils/streets_signal_phase_and_timing/src/models/time_change_details.cpp
paulbourelly999: This is the earliest time we can represent in the SPaT since we can represent a time from a previous hour

tsc_client_service/test/test_tsc_service.cpp
paulbourelly999: It means the process_snmp_method returns true. This is the syntax for gmock so I don't really think this needs in line documentation. If the method documents what return true means (process_snmp_request) and gmock documents how to use the EXPECT CALL macros I don't think addition documentation is neccessary.

tsc_client_service/test/test_tsc_service.cpp
paulbourelly999: I think that is a lot of work for very marginal value in the unit test.

adev4a (approved):

PR #276, opened 2022-12-02
Fix update desired plan

adev4a:
# PR Details
Fixes an issue with updating desired phase plan, to check if the list is currently empty before removing entries.
## Types of changes

## Checklist:

### DIFF
--- a/tsc_client_service/src/monitor_desired_phase_plan.cpp
+++ b/tsc_client_service/src/monitor_desired_phase_plan.cpp
@@ -50,7 +50,7 @@ namespace traffic_signal_controller_service
             
             // Loop through the desired phase plan and remove an event if the end time is past current
             bool no_expired_events = false;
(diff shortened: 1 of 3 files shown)


tsc_client_service/src/monitor_desired_phase_plan.cpp
```diff
@@ -50,13 +50,15 @@ namespace traffic_signal_controller_service
             
             // Loop through the desired phase plan and remove an event if the end time is past current
             bool no_expired_events = false;
-            while(!no_expired_events){
+            while( !no_expired_events || desired_phase_plan_ptr != nullptr || !desired_phase_plan_ptr->desired_phase_plan.empty()){
```
paulbourelly999, line 53: Is or correct here?

adev4a: Yes, we need to stop the while loop in either of those conditions

adev4a: Yes, we need to stop the while loop if either of conditions are not true

Issue #284, opened 2022-12-20
convert_spat_to_dpp movement group sequence issue in dpp_generator


Saeid-Solei:
### Types of Issue

- [x] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

Depending on the sequence of available movement groups in the movement group list, the convert_spat_to_dpp method in streets_desired_phase_plan_generator might make a mistake in the converting spat to dpp and might not return the correct sequence of movement groups. 

### Carma streets version where this issue was discovered

4.2.0

### Expected behavior
The converted dpp from spat always have the correct sequence of movement groups. 

### Actual behavior

Depending on the sequence of movement groups in the movement group list, sometimes the converted dpp does not include the correct sequence of movement groups.

PR #285, opened 2022-12-21
convert_spat_to_dpp in dpp_generator returns wrong movement group sequence

Saeid-Solei:
# PR Details
## Description

Depending on the sequence of available movement groups in the movement group list, the convert_spat_to_dpp method in streets_desired_phase_plan_generator might make a mistake in the converting spat to dpp and might not return the correct sequence of movement groups.

## Related Issue

https://github.com/usdot-fhwa-stol/carma-streets/issues/284

## How Has This Been Tested?

Unit Tests

## Types of changes

## Checklist:

### DIFF
--- a/streets_utils/streets_signal_optimization/src/streets_desired_phase_plan_generator.cpp
+++ b/streets_utils/streets_signal_optimization/src/streets_desired_phase_plan_generator.cpp
@@ -217,6 +217,11 @@ namespace streets_signal_optimization {
                 base_desired_phase_plan.desired_phase_plan.push_back(mg_green);
             }
             else {
+                /** 
+                 * is_added parameter checks if mg_green has been added to the base_desired_phase_plan 
+                 *      somewhere before the last existing movement_group. If this parameter is false at the end of
+                 *      the for loop, mg_green must be added to the end of the list.
+                 */
                 bool is_added = false;
                 for (int index = 0; index < base_desired_phase_plan.desired_phase_plan.size(); ++index) {
                     if ( (mg_green.start_time < base_desired_phase_plan.desired_phase_plan[index].start_time && 
@@ -231,6 +236,10 @@ namespace streets_signal_optimization {
                         break;
                     }
                 }
+                /**
+                 * If is_added is false, mg_green is not added to the base_desired_phase_plan which means that mg_green
+                 *      must be added to the end of the list!
+                */
                 if (!is_added) {
                     base_desired_phase_plan.desired_phase_plan.push_back(mg_green);
                 }
--- a/streets_utils/streets_signal_optimization/src/streets_desired_phase_plan_generator.cpp
+++ b/streets_utils/streets_signal_optimization/src/streets_desired_phase_plan_generator.cpp
@@ -217,6 +217,7 @@ namespace streets_signal_optimization {
                 base_desired_phase_plan.desired_phase_plan.push_back(mg_green);
             }
             else {
+                bool is_added = false;
                 for (int index = 0; index < base_desired_phase_plan.desired_phase_plan.size(); ++index) {
                     if ( (mg_green.start_time < base_desired_phase_plan.desired_phase_plan[index].start_time && 
                                 mg_green.end_time > base_desired_phase_plan.desired_phase_plan[index].start_time) || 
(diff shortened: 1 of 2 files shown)


streets_utils/streets_signal_optimization/test/test_streets_desired_phase_plan_generator.cpp
```diff
@@ -176,6 +176,23 @@ TEST(test_streets_desired_phase_plan_generator, test_convert_spat_to_dpp) {
...
+    spat_object.fromJson(json_spat_2green);
+    intersection_state = spat_object.get_intersection();
+    base_desired_phase_plan = generator.convert_spat_to_dpp(intersection_state, move_groups);
+    ASSERT_EQ( base_desired_phase_plan.desired_phase_plan.size(), 3);
+    SPDLOG_INFO("converted spat to dpp: {0}", base_desired_phase_plan.toJson());
+    ASSERT_EQ( base_desired_phase_plan.desired_phase_plan.front().signal_groups.front(), 1);
+    ASSERT_EQ( base_desired_phase_plan.desired_phase_plan.back().signal_groups.front(), 1);
 }
```
paulbourelly999, line 197: Could we add a new unit test case with the data from yesterday as testing data

streets_utils/streets_signal_optimization/src/streets_desired_phase_plan_generator.cpp
```diff
@@ -217,6 +217,7 @@ namespace streets_signal_optimization {
                 base_desired_phase_plan.desired_phase_plan.push_back(mg_green);
             }
             else {
+                bool is_added = false;
```
paulbourelly999, line 220: Could you explain this fix with a comment. What is the is_added flag indicating

streets_utils/streets_signal_optimization/test/test_streets_desired_phase_plan_generator.cpp
Saeid-Solei: This is a similar data to what we had in testing yesterday. I was able to reproduce the exact same behavior with this example. The problem was in the logic that adds movement_groups to the desired_phase_plan as part of conversion.
Please let me know if you still want to see the exact same example.

streets_utils/streets_signal_optimization/src/streets_desired_phase_plan_generator.cpp
Saeid-Solei: Added some explanations.

paulbourelly999: To be honest I do not understand this logic at all. You have a base desired phase plan which should be empty for the first iteration of adding movement groups everytime. Then you are looping over all the entries in the desired phase plan for each new movement group to add. Why? Seems like if you just ordered the movement groups, you could loop through them once and add all in the correct order. This is why I do not understand this flag either. What is this flag guarding against that wasn't previously account for.  Maybe inline comments would help me understand this logic, but it looks like you are using some insert logic to ensure that each new entry is added at index +1 of the first dpp entry where its start time is greater than the existing entries end time. What happens if the first movement group you add is not the first in order. It seems like if the first movement group is incorrect there is no way to account for that.

Saeid-Solei: This entire for loop here is for sorting the movement groups and adding them to the desired phase plan in a sorted manner. 

Previously, for each new movement group, we looped over all the entries in the desired phase plan. At a given iteration, if the new movement group's end time is less than the start time of an entry in the desired phase plan, that means we the new movement group shall come before this entry in the desired phase plan. Similarly, at a given iteration, if the new movement group's start time is greater than the end time of an entry in the desired phase plan, we assumed that the new movement group shall come immediately after this entry. However this is not true. It is possible that the new movement group shall come after this entry, but not immediately. To fix this issue, we will not compare the new movement group's start time with the entry's end time anymore. Instead, if a new movement group could not be added before any entry in the desired phase plan, it means that is should be added at the end of the list.

If this is still confusing for you, please let me know. Maybe we can have a quick discussion.

paulbourelly999 (approved):
Looks good

Issue #281, opened 2022-12-14
Signal Optmization Service performance logging produces empty logfiles


paulbourelly999:
### Types of Issue

- [x] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Carma streets version where this issue was discovered

Develop
### Expected behavior
Signal Optimization Service is meant to log performance metrics including candidate desire phase plans considered and their delay measure to use to evaluate Signal Optimization Service performance and confirm phase sequence and duration selection decisions. 
### Actual behavior

While the Signal Optimization Service seems to be correctly producing desired phase plans for single and multiple vehicle test, it does not produce any performance logs associated with the decision.
### Steps to reproduce the actual behavior

1. Deploy CARMA-Streets
2. Deploy and engage CARMA-Platform vehicle with a route that travels through the intersection
3. CARMA-Streets will produce a Desire Phase Plan and modify Traffic Signal Controller default phase sequence but does not produce associated performance logs

PR #287, opened 2022-12-23
Add Flush_on to so csv logger!

Saeid-Solei:
# PR Details
## Description

Signal Optimization Service is meant to log performance metrics including candidate desire phase plans considered and their delay measure to use to evaluate Signal Optimization Service performance and confirm phase sequence and duration selection decisions. While the Signal Optimization Service seems to be correctly producing desired phase plans for single and multiple vehicle test, it does not produce any performance logs associated with the decision.

## Related Issue

https://github.com/usdot-fhwa-stol/carma-streets/issues/281

## Types of changes

## Checklist:


signal_opt_service/src/signal_opt_service.cpp
```diff
@@ -338,6 +338,7 @@ namespace signal_opt_service
             // Only log log statement content
             csv_logger->set_pattern("%v");
             csv_logger->set_level(spdlog::level::info);
+            csv_logger->flush_on(spdlog::level::info);
```
paulbourelly999, line 341: @adev4a @Saeid-Solei Is there a reason you are using `flush_on(spdlog::level::info)` instead of just `flush()`. I think based on this documentation (https://github.com/gabime/spdlog/wiki/7.-Flush-policy#severity-based-flush) this sets the logger to flush on every `info` level log which will be every log statement. I am not sure this is the correct behavior. Also were you able to test any of this locally?

paulbourelly999: NVM, after reading this (https://github.com/gabime/spdlog/wiki/0.-FAQ#the-log-file-remains-empty) we should include this for all of our performance logging. Can you add this to the scheduling service to. Flushing is not removing queue entries but is a performance setting the controls when message will be written to the loggers configured sink, similar to batching messages. For performance loggers which may print logs less frequently the default behavior for flushing set in libc may not be ideal and may leave messages out due to flushing policy. We want to ensure that every message gets written immediately for performance logging so maybe we add this to all of our performance loggers

paulbourelly999 (changes requested):

paulbourelly999 (approved):
looks good!

Issue #289, opened 2022-12-27
Occasional SNMP timeout on  PHASE_STATUS_GROUP_PHASE_NEXT GET requests during yellow change and red clearance


paulbourelly999:
### Types of Issue

- [ x] Anomaly report (something appears to not work correctly)
- [ ] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

Occasional SNMP timeout on  PHASE_STATUS_GROUP_PHASE_NEXT GET requests during yellow change and red clearance
### Expected behavior
`tsc_service` sends a spat message for every udp packet received from the tsc
### Actual behavior

Occasional SNMP timeout on  PHASE_STATUS_GROUP_PHASE_NEXT GET requests during yellow change and red clearance will cause a single SPaT to be dropped
### Steps to reproduce the actual behavior

1. Deploy `tsc_service`

PR #293, opened 2023-01-05
289 update snmp timeout

paulbourelly999:
# PR Details
## Description

Update `snmp_client` to make snmp timeout default to 15000 microseconds. Included updates to class to remove unnecessary class member as well as updating configuration parameter name and description to more accurately represent functionality
## Related Issue
#289 

## Motivation and Context
Occasionally NTCIP GET NEXT PHASE GROUP commands sent during yellow change and red clearance intervals will timeout due to the 10 ms set timeout. This will cause the TSC to drop a single SPaT message. Increasing the default value of this parameter and cleaning up unnecessary class member/confusing configuration name/incorrect configuration description can help reduce the number of times we drop SPaT messages under these conditions. 

## How Has This Been Tested?
Integration tested on VM Test Instance 

## Types of changes

## Checklist:

### DIFF
--- a/tsc_client_service/src/snmp_client.cpp
+++ b/tsc_client_service/src/snmp_client.cpp
@@ -4,7 +4,7 @@ namespace traffic_signal_controller_service
 {
     
     snmp_client::snmp_client(const std::string& ip, const int& port, const std::string& community, int snmp_version, int timeout)
-        : ip_(ip), port_(port), community_(community),snmp_version_(snmp_version), timeout_(timeout)
+        : ip_(ip), port_(port), community_(community),snmp_version_(snmp_version)
     {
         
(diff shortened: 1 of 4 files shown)


tsc_client_service/src/snmp_client.cpp
```diff
@@ -28,7 +28,7 @@ namespace traffic_signal_controller_service
         
         session.community = comm;
         session.community_len = community_.length();
-        session.timeout = timeout_;
```
dan-du-car, line 31: Can we add a comment to note the unit of measure this time is microsecond?

paulbourelly999: Yes, good call!

dan-du-car (approved):

Issue #259, opened 2022-11-18
Update CS Services with latency logging for important processes.


paulbourelly999:
### Types of Issue

- [ ] Anomaly report (something appears to not work correctly)
- [x] Enhancement request (describe the enhancement being requested)
- [ ] Other (please ensure the description clarifies why the issue doesn’t fall into either of the above categories)

### Descriptive summary

Implement logging for performance latency for **SPaT generation**, **SO Desired Phase Plan selection**, **Intersection Schedule generation**, and **SNMP dynamic requests**.
### Carma streets version where this issue was discovered

Latest
### Expected behavior
NA
### Actual behavior

NA
### Steps to reproduce the actual behavior

NA

PR #297, opened 2023-01-11
Issue-259: Add proccessing time logging

paulbourelly999:
+ Add processing time logging for Signal Optimization iteration
 + Add processing time logging for Scheduling iteration
 + Update Spat processing time logging 

# PR Details
## Description

Add processing time log statements for signal optimization iterations and for scheduling iterations

## Related Issue
#259 

## Motivation and Context
Log  performance metrics including Signal Optimization calculation time and scheduling calculation time 

## How Has This Been Tested?
Tested on Test Instance deployment. Start Time and End Times print. Spat lag also prints.

## Types of changes

## Checklist:

### DIFF
--- a/signal_opt_service/src/signal_opt_service.cpp
+++ b/signal_opt_service/src/signal_opt_service.cpp
@@ -177,7 +177,7 @@ namespace signal_opt_service
                 SPDLOG_DEBUG("Current spat lag is {0} ms!", spat_lag); 
             }
             catch( const signal_phase_and_timing::signal_phase_and_timing_exception &e) {
-                SPDLOG_ERROR("Cannot interpret SPAT due to exception :  {0}", e);
+                SPDLOG_ERROR("Cannot interpret SPAT due to exception :  {0}", e.what());
                 std::this_thread::sleep_for(std::chrono::milliseconds(_so_sleep_time));
                 continue;
             }
--- a/signal_opt_service/src/signal_opt_service.cpp
+++ b/signal_opt_service/src/signal_opt_service.cpp
@@ -170,11 +170,18 @@ namespace signal_opt_service
             auto current_timestamp = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch()).count();
             SPDLOG_DEBUG("Signal Optimization iteration start time {0}!", current_timestamp);
             try{
-                auto spat_lag =  current_timestamp - _spat_ptr->get_intersection().get_epoch_timestamp() ;
-                    if ( spat_lag > 200 ) {
-                        SPDLOG_WARN("Current SPat Lag exceends 200 ms!");
-                    } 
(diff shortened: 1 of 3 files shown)


signal_opt_service/src/signal_opt_service.cpp
```diff
@@ -165,9 +165,29 @@ namespace signal_opt_service
...
+                }
+                else {
+                    SPDLOG_WARN("SPaT Pointer is unset, skipping SO Iteration");
+                    std::this_thread::sleep_for(std::chrono::milliseconds(_so_sleep_time));
+                    continue;
+                }
+            }
+            catch( const signal_phase_and_timing::signal_phase_and_timing_exception &e) {
+                SPDLOG_ERROR("Cannot interpret SPAT due to exception :  {0}", e.what());
+                std::this_thread::sleep_for(std::chrono::milliseconds(_so_sleep_time));
```
adev4a, line 188: Why do we need a sleep in case of an exception?

signal_opt_service/src/signal_opt_service.cpp
```diff
@@ -165,9 +165,29 @@ namespace signal_opt_service
...
+                    auto spat_lag =  current_timestamp - _spat_ptr->get_intersection().get_epoch_timestamp() ;
+                        if ( spat_lag > 200 ) {
+                            SPDLOG_WARN("Current SPat Lag exceends 200 ms!");
+                        } 
+                    SPDLOG_DEBUG("Current spat lag is {0} ms!", spat_lag); 
+                }
+                else {
+                    SPDLOG_WARN("SPaT Pointer is unset, skipping SO Iteration");
+                    std::this_thread::sleep_for(std::chrono::milliseconds(_so_sleep_time));
+                    continue;
```
adev4a, line 183: How would this else and catch be different without the continue?

dan-du-car (approved):

signal_opt_service/src/signal_opt_service.cpp
paulbourelly999: We want this to run at a frequency regardless if there is an error condition or not. If we do not sleep and continue here, error conditions would cause this error statement to print infinitely until the spat pointer is initialized. This can happen when consumer groups on kafka are rebalancing. I.E some consumers are up and running while others are not. If the spat consumer group is rebalancing, this sleep and continue will poll the object at SO iteration frequency and resolve once the rebalance is complete.

adev4a (approved):

signal_opt_service/src/signal_opt_service.cpp
paulbourelly999: As stated in the above response, we want the SO service to attempt a signal optimization calculation at the configured frequency regardless of failure. Without this sleep and continue, when the spat is set but empty, (can happen due to consumer group rebalancing) this statement will print out indefinitely which out a interval in between for the spat to be populated by the spat consumer.

Issue #303, opened 2023-01-24
SO Service unrecoverable error state.


paulbourelly999:
### Summary

Currently it is possible for the SO Service to enter an unrecoverable error state depending on when it sends it’s last DPP for a given green phase. This is because to confirm a DPP has been received by the TSC Service, on the SO service we confirm that the size of the current spat has increased. If we send a DPP in the final SO iteration for a given green ( time to yellow is < so frequency), the next time we check the spat, it may have the added phase but its size will seem unchanged since the previous green is now no longer present.  Once in this state, the SO service has no way to recover.

![image](https://user-images.githubusercontent.com/77466294/214436255-8ae8edb6-0e89-4250-95aa-c931120c6dc2.png)
![image](https://user-images.githubusercontent.com/77466294/214436306-2acefb76-ea89-4cae-ae2c-3926f725d478.png)

### Commit

be46553cfd8669e16f9828d70bd80c37c50c94ba

### Expected Behavior

SO Service is able to accurately determine :
- Whether its last sent DPP was received
- Whether it needs to run SO calculations again

and has functionality to recover incase its last DPP sent was not received.

### Actual Behavior

If the SO Service creates and sends a DPP on its last iteration in a given green it will enter an unrecoverable error state where it will no longer produce any DPPs for any future vehicles. The service won't crash but will continually skip running the SO calculations because it still believes its last DPP was not yet excepted by the TSC Service.

### Steps to Reproduce the Actual Behavior

1. Deploy CARMA-Streets.
2. Repeatedly engage CARMA-Streets at the end of a green to attempt to produce this behavior.
3. If the SO service is no longer creating DPPs for new vehicles and instead you see the following log statement you are in the error state:
`"Skipping SO iteration. Previously sent DPP is not yet reflected in SPaT!"`

PR #304, opened 2023-01-24
303 dpp producer reset

paulbourelly999:
# PR Details
## Description

Currently it is possible for the SO Service to enter an unrecoverable error state depending on when it sends it’s last DPP for a given green phase. This is because to confirm a DPP has been received by the TSC Service, on the SO service we confirm that the size of the current spat has increased. If we send a DPP in the final SO iteration for a given green ( time to yellow is < so frequency), the next time we check the spat, it may have the added phase but its size will seem unchanged since the previous green is now no longer present.  
## Related Issue
#303 

## Motivation and Context

Make SO service robust to edge case scenarios

## How Has This Been Tested?
NA

## Types of changes

## Checklist:

### DIFF
--- a/signal_opt_service/src/signal_opt_service.cpp
+++ b/signal_opt_service/src/signal_opt_service.cpp
@@ -165,6 +165,10 @@ namespace signal_opt_service
         int prev_future_move_group_count = 0;
         int current_future_move_group_count = 0;
         bool new_dpp_generated = false;
+        // Keeps track of sequential SO iteration skips caused by percieved missing DPP data in spat.
+        // TODO: More robust fix could use check last entry in sent DPP vs last entry in received spat 
+        // instead of future movement group counts/
+        int sequential_skips_for_missing_dpp = 0;
                
         while ( dpp_producer->is_running() ) {
             auto current_timestamp = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch()).count();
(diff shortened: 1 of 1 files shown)


signal_opt_service/src/signal_opt_service.cpp
```diff
@@ -213,11 +217,21 @@ namespace signal_opt_service
                         new_dpp_generated = false;
                     }
                     else {
-                        SPDLOG_WARN("Skipping SO iteration. Previously sent DPP is not yet reflected in SPaT!");
+                        // Increment sequential skip count for each skip
+                        sequential_skips_for_missing_dpp++;
+                        SPDLOG_WARN("Skipping SO iteration {0}. Previously sent DPP is not yet reflected in SPaT!", sequential_skips_for_missing_dpp);
+                        if (sequential_skips_for_missing_dpp > 5 ) {
```
Saeid-Solei, line 223: In our successful runs, are we sure that SO will receive the updated SPaT within next 5 iteration after sending the dpp? Do we want to make it a configurable parameter?

paulbourelly999: Yes we are sure, in fact I have never seen a missed DPP in all data collection. 5 missed DPPs would equate to approximately 2 seconds of missed information. The only behavior I have see is ~30ms latency causing a DPP to be invalid once it is received.

paulbourelly999: Additionally even though this is impossible, if we sent a DPP and it hasn't been received for 2 seconds we can likely assume it had been dropped some how or is now invalid so coming up with a new one is the appropriated response I think.

Saeid-Solei: You are saying that 5 iterations means 2 seconds of missed information. I guess this depends on the frequencies we define. I am fine with leaving 5 here for now. But we probably should be careful when changing the SO frequency.

Issue #324, opened 2023-05-01
TSC Service failed to link librdkafka library


paulbourelly999:
### Summary

Running TSC Service image would result in the following core dump:
```
tsc_service           | /home/carma-streets/tsc_client_service/build/traffic_signal_controller_service: error while loading shared libraries: librdkafka.so.1: cannot open shared object file: No such file or directory
```

### Commit

9b0f608

### Expected Behavior

TSC Service runs without exception or core dump when configured correctly.

### Actual Behavior

Running TSC Service image would result in the following core dump:
```
tsc_service           | /home/carma-streets/tsc_client_service/build/traffic_signal_controller_service: error while loading shared libraries: librdkafka.so.1: cannot open shared object file: No such file or directory
```

### Steps to Reproduce the Actual Behavior

1. Pull tsc_service docker image.
2. Run tsc_service docker image

PR #325, opened 2023-05-01
Fixes for tsc_service failed initilaization

paulbourelly999:
# PR Details
## Description

+ Executable could not find librkafka library (fixed with ldconfig call)
+ Added find package to find kafka-clients library
+ Update main to not attempt to initialize logger since that now happens inside service base (This was triggering singleton exception)
+ Updated docker-compose with correct environment variables for V2X-Hub and tsc_service
+ Added javadoc comments to service base classes
+ Added try catch to initialize method to ensure no exceptions are uncaught
+ Update streets_service to launch time consumer as detached thread to avoid blocking base class start method or thread to go out of scope in inherited start method.

## Related Issue
#324

## Motivation and Context
XIL CARMA-Streets Simulation Integration

Simulation Integration with CARMA-Streets under XIL
## How Has This Been Tested?

Integration tested by deploying V2X-Hub and CARMA Streets.
1) Deploy V2X-Hub in non-simulation mode and forward MAP message to kafka using the CARMA-Streets Plugin
2) Deploy Intersection Model with an OSM map that shares an intersection id with the MAP message
3) Deploy EVC with no spat forwarding configurations (1 2 7 menu option all set to zeros)
4) Send snmpset -v1 -c administrator <ip-address>:SNMP-Port  1.3.6.1.4.1.1206.3.5.2.9.44.1.0 i 2 to enable spat ( do not use local host as IP)
5) Configure tsc_service to listen for spat at 6053 and send NTCIP commands to ip-address:SNMP-Port. ( do not use local host as IP)
6) Deploy TSC Service in simulation mode with the following environment variables set
      SIMULATION_MODE: TRUE
      CONFIG_FILE_PATH: ../manifest.json
      TIME_SYNC_TOPIC: time_sync
7) Use docker exec -it kafka bash to ssh into kafka container
8) Use kafka-console-producer --bootstrap-server=127.0.0.1:9092 --topic=time_sync to send sample time sync messages
{ "timestep":1231,"seq":1231322}
9) This should trigger tsc_service to produce SPaT JSON on kafka topic modified_spat

## Types of changes

## Checklist:


tsc_client_service/src/tsc_service.cpp
```diff
@@ -183,16 +184,9 @@ namespace traffic_signal_controller_service {
...
-                        if (count <= 20 ) {
-                            uint64_t timestamp = streets_clock_singleton::time_in_ms();
-                            spat_processing_time += timestamp - spat_ptr->get_intersection().get_epoch_timestamp();
-                            count++;
-                        } else {
-                            double total_processing_time = ((double)(spat_processing_time))/20.0;
-                            SPDLOG_INFO("SPat average processing time over 20 messages is {0} ms and total processing time for 20 messages is {1} ms!", total_processing_time, spat_processing_time);
-                            spat_processing_time = 0;
-                            count = 0;
+                        // No utility in calculating spat latency in simulation mode
```
paulbourelly999, line 187: Refactored into separate method to resolve sonar scanner code smell

tsc_client_service/test/test_tsc_service.cpp
```diff
@@ -424,6 +424,32 @@ namespace traffic_signal_controller_service
 
     }
 
+    TEST_F(tsc_service_test, test_log_spat_latency){
```
paulbourelly999, line 427: New test for refactored method

tsc_client_service/test/test_tsc_service.cpp
```diff
@@ -454,9 +480,9 @@ namespace traffic_signal_controller_service
         auto logger = spdlog::get("snmp_cmd_logger"); 
         EXPECT_TRUE(logger != nullptr);
     }
-    // TODO: Restore test by fixting streets_configuration_singleton
-    // TEST_F(tsc_service_test, test_initialization_no_mock_response_from_snmp_client) {
-    //     ASSERT_FALSE(service.initialize());
-    // }
+    // Test tsc_service initialize without mocking snmp responses
```
paulbourelly999, line 483: Fixed singleton destructor for streets_configuration singleton and uncommented test case.

streets_utils/streets_service_configuration/test/test_streets_configuration.cpp
```diff
@@ -10,147 +10,164 @@
 #include "streets_configuration.h"
```
paulbourelly999, line 11: All Changes here are just making the test cases into test fixtures by creating a test class

streets_utils/streets_service_configuration/include/internal/streets_singleton.tpp
```diff
@@ -43,10 +42,9 @@ namespace streets_service {
     template <typename T, typename... Args>   
     streets_singleton<T,Args...>::~streets_singleton() {
       SPDLOG_WARN("Deleting Singleton of type {0}!", typeid(instance).name());
-      // Avoid loop recursive call of destructor
-      if ( typeid(T) != typeid(instance) ){
-          delete instance;
-      }
-      instance = nullptr;
+      // Zero out first to avoid recursive call to desctructor
+      // https://www.codeproject.com/Articles/987473/Unit-Testing-a-Singleton-in-Cplusplus
```
paulbourelly999, line 46: Probably still some work to do here since the only way the unit test currently trigger this destructor is on exceptions.

streets_utils/streets_service_base/test/test_streets_service.cpp
```diff
@@ -70,5 +70,20 @@ namespace streets_service{
         ASSERT_THROW(serv.get_system_config("NON_EXISTANT"), std::runtime_error);
         ASSERT_THROW(serv.get_system_config(nullptr), std::runtime_error);
     };
+    TEST_F(test_streets_service, test_start) {
+        ASSERT_TRUE(serv.initialize());
+        serv.start();
+    }
+
+    TEST_F(test_streets_service, test_initialize_exception) {
```
paulbourelly999, line 78: Unit tests added for catch statements

streets_utils/streets_service_base/test/test_streets_service.cpp
```diff
@@ -70,5 +70,20 @@ namespace streets_service{
         ASSERT_THROW(serv.get_system_config("NON_EXISTANT"), std::runtime_error);
         ASSERT_THROW(serv.get_system_config(nullptr), std::runtime_error);
     };
+    TEST_F(test_streets_service, test_start) {
```
paulbourelly999, line 73: Unit test add for start method

streets_utils/streets_service_base/test/test_streets_clock_singleton.cpp
```diff
@@ -3,44 +3,59 @@
 
 using namespace streets_service;
 using namespace std::chrono;
-
```
paulbourelly999, line 6: Added test class and test fixture.

streets_utils/streets_service_base/test/test_files/invalid.json
```diff
@@ -0,0 +1 @@
+{"service_name":"test_service","loglevel":"trace","configurations":[{"name":"param1","value":123,"description":"Test Parameter 1","type":"INTEGER"},{"name":"param2","value":true,"description":"Test Parameter 2","type":"BOOL"},{"name":"bootstrap_server","value":"127.0.0.1:9092","description"asd:"Test Parameter 3","type":"STRING"},asd{"name":"param4","value":24.2,"description":"Test Parameter 4","type":"DOUBLE"}]}
```
paulbourelly999, line 1: Invalid json to test configuration exception handling

streets_utils/streets_service_base/src/streets_service.cpp
```diff
@@ -62,14 +67,19 @@ namespace streets_service {
...
-        {
+        {   
             const std::string payload = _time_consumer->consume(1000);
             if (payload.length() != 0)
             {
-                SPDLOG_DEBUG("Consumed: {0}", payload);
-                simulation::time_sync_message msg;
-                msg.fromJson(payload);
-                streets_clock_singleton::update(msg.timestep);
+                try {
```
paulbourelly999, line 74: Exception handling for invalid time sync messages.

streets_utils/streets_service_base/include/streets_service.h
```diff
@@ -29,22 +36,61 @@ namespace streets_service {
...
+             * @return Returns service name set in configuration file.
+             */
             std::string get_service_name() const;
-
+            /**
+             * @brief Returns boolean if service is currently configured to be in simulation_mode. This corresponsds
+             * to the value of the environment variable SIMULATION_MODE. Simulation mode indicates whether the streets_clock_singleton
+             * uses simulation time for kafka or system time.
+             * @return true if in simulation mode, false if in real time more
+             */
```
dan-du-car, line 93: What is the default mode? What is diff between simulation time and system time? e.g.?

streets_utils/streets_service_base/include/time_sync_message.h
```diff
@@ -7,12 +7,30 @@
...
+     * during simulation. This message is sent to all CARMA Streets services
+     * via kafka and consumed to manage simulation time.
+    */
     struct time_sync_message{
+        /**
+         * @brief Current time in ms.
+         */
         uint64_t timestep;
+        /**
+         * @brief Sequential number of timestep.
```
dan-du-car, line 21: How is this seq number used by the timestep?

streets_utils/streets_service_base/src/streets_clock_singleton.cpp
```diff
@@ -21,6 +25,7 @@ namespace streets_service {
 
     void streets_clock_singleton::sleep_until(uint64_t ms) {
         auto &inst = get_singleton();
+        inst.wait_for_initialization();
```
dan-du-car, line 28: inline comment why wait for initialization?

streets_utils/streets_service_base/src/streets_service.cpp
```diff
@@ -13,7 +13,8 @@ namespace streets_service {
         try {
             std::string config_file_path = get_system_config("CONFIG_FILE_PATH");
             streets_configuration::create(config_file_path);
-            _simulation_mode = get_system_config("SIMULATION_MODE").compare("TRUE") == 0;
+            std::string sim_mode_string = get_system_config("SIMULATION_MODE");
+            _simulation_mode = sim_mode_string.compare("true") == 0 || sim_mode_string.compare("TRUE") == 0 ;
```
dan-du-car, line 17: can we lower the sim_mode_string then compare? So we can remove the || and use one comparison?

streets_utils/streets_service_base/src/streets_service.cpp
```diff
@@ -62,14 +67,19 @@ namespace streets_service {
...
             {
-                SPDLOG_DEBUG("Consumed: {0}", payload);
-                simulation::time_sync_message msg;
-                msg.fromJson(payload);
-                streets_clock_singleton::update(msg.timestep);
+                try {
+                    SPDLOG_DEBUG("Consumed: {0}", payload);
+                    simulation::time_sync_message msg;
+                    msg.fromJson(payload);
+                    streets_clock_singleton::update(msg.timestep);
```
dan-du-car, line 78: Why do we need to update the timestamp after consuming a payload? Is this payload generated by sim environment?

streets_utils/streets_service_base/src/streets_service.cpp
```diff
@@ -62,14 +67,19 @@ namespace streets_service {
     void streets_service::consume_time_sync_message() const  {
         _time_consumer->subscribe();
         while (_time_consumer->is_running())
```
dan-du-car, line 69: Is there a design doc for this street service?

streets_utils/streets_service_configuration/include/internal/streets_singleton.tpp
```diff
@@ -19,11 +19,10 @@ namespace streets_service {
...
       if (instance != nullptr){
         SPDLOG_WARN("Recreating Singleton of type {0}!", typeid(instance).name());
-        // Avoid loop recursive call of destructor
-        if ( typeid(T) != typeid(instance) ){
-          delete instance;
-        }
-        instance = nullptr;
+        // Remove old instance
+        
+        delete instance;
```
adev4a, line 24: Just so i understand this correctly, we were earlier deleting the instance object and then initializing it to a nullptr, but now we dont need to initialize it?

streets_utils/streets_service_configuration/include/internal/streets_singleton.tpp
```diff
@@ -43,10 +42,9 @@ namespace streets_service {
...
     streets_singleton<T,Args...>::~streets_singleton() {
       SPDLOG_WARN("Deleting Singleton of type {0}!", typeid(instance).name());
-      // Avoid loop recursive call of destructor
-      if ( typeid(T) != typeid(instance) ){
-          delete instance;
-      }
-      instance = nullptr;
+      // Zero out first to avoid recursive call to desctructor
+      // https://www.codeproject.com/Articles/987473/Unit-Testing-a-Singleton-in-Cplusplus
+      instance = 0;
```
adev4a, line 47: Why is this instance = 0 required here?

streets_utils/streets_service_base/include/streets_service.h
paulbourelly999: So there is not really a default. If the SIMULATION_MODE environment variable is set to false (more accurately anything but TRUE or true) the system functions in real time using chronos system time calls. If set to True or true the system will use time_sync messages to update its "current time". This is used so that we can deploy this service in simulation where time may progress slower than real time. Time Sync messages are sent by our XIL software which manages all the simulators and our software and only sends the next time step when all software has completed calculation of the previous time step and reports it is ready. Check out https://usdot-carma.atlassian.net/wiki/spaces/CRMSRT/pages/2425946175 and https://usdot-carma.atlassian.net/wiki/spaces/CRMSIM/pages/2444263441/XIL+Time+Synchronization+Design+for+V2X-Hub+CARMA-Streets

streets_utils/streets_service_base/include/time_sync_message.h
paulbourelly999: This is part of the Mosiac Time Message. It is not used by V2X-Hub or CARMA-Streets yet but I included since the design docs Kyle made included it and we may want to have access to it in the future.

streets_utils/streets_service_base/src/streets_clock_singleton.cpp
paulbourelly999: This is a blocking method that essentially registers a listener for the calling thread for the first time update. So if running in simulation mode we need to wait for the first time sync message and confirm the clock has been initialize at a time before we are able to do any time related functionality. For more information about how carma-clock works please check out https://github.com/usdot-fhwa-stol/carma-time-lib which is the library that defines the underlying object this singleton handles.

streets_utils/streets_service_base/src/streets_service.cpp
paulbourelly999: Its really the same number of lines of code so I didn't do it since it is really largely equivalent. The only case it does not cover is one with mixed capitalization (TrUe) which to me is not one we need to cover. Not super opinionated about this one but I don't think it there is much reason to change it.

streets_utils/streets_service_base/src/streets_service.cpp
paulbourelly999: Streets service is not a new service but a base class for services that will now use this time wrapper (carma-clock) -> streets_clock_singleton instead of direct calls to system time. The design docs I sent earlier should have information about this. The idea is now we can use an environment variable to change whether our carma-streets service operate in real time or in simulation time so we can also test the same code in simulation. The service base class also reads the configuration file and has helper methods to setup kafka consumers and producers since most carma-streets services will need this functionality and should not have to reimplement it like is currently done.

streets_utils/streets_service_base/src/streets_service.cpp
paulbourelly999: Yes see early comments and documentation

streets_utils/streets_service_configuration/include/internal/streets_singleton.tpp
adev4a, line 46: Is instance here a pointer? What does assigning it a value of 0 mean?

streets_utils/streets_service_configuration/include/internal/streets_singleton.tpp
paulbourelly999: This is a part I am not super sure about. Delete calls the deconstructor. There was some online implementation that was handling singleton clean up by calling delete and then assigning it null pointer. Since singletons essentially have global scope this was never an issue that impacted running the service. It did however occasionally cause unit tests to fail when the singleton destructor ran. We would get a double free exception assumingly because somehow over multiple unit tests an instance of this singleton would have delete called on it multiple times. To be honest I just stole this implementation from the link in the inline comment.

streets_utils/streets_service_configuration/include/internal/streets_singleton.tpp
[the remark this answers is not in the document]
paulbourelly999: This is a solution I found online for avoiding exceptions when unit testing singletons although I agree that this may be a memory leak if we were recreating singletons in our services. Will try to find a better way to address this.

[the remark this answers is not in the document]
paulbourelly999: @kjrush and @adev4a, removed all constructor and deconstructors and used smart pointer instead to avoid deletes. Also found from gdb and core dumps that some of the segfaults in unit tests seemed to be caused by calls to logger in singleton destructor. This may be because singleton scope exceeds logger scope. I think this implementation is a lot cleaner. What do you guys think.

streets_utils/streets_service_configuration/include/internal/streets_singleton.tpp
paulbourelly999: Please check out new logic

streets_utils/streets_service_base/src/streets_clock_singleton.cpp
```diff
@@ -5,10 +5,14 @@ namespace streets_service {
...
         auto &inst = get_singleton();
         inst.update(timestep);
+        SPDLOG_TRACE("streets clock is updated to {0}", timestep);
     }
 
     uint64_t streets_clock_singleton::time_in_ms() {
         auto &inst = get_singleton();
+        SPDLOG_TRACE("Calling thread is waiting on streets clock.");
+        inst.wait_for_initialization();
+        SPDLOG_TRACE("Thread is released after initializeation.");
```
dan-du-car, line 15: Are these two logs needed in this function. I see other similar functions also called the wait for initialization but logs not in those functions. Maybe add similar logs as the above SPDLOG_TRACE("streets clock now in milliseconds {0}", inst.nowInMilliseconds());

dan-du-car:
Thanks for clarifying the questions in the call.

dan-du-car (dismissed):

docker-compose.yml
```diff
@@ -189,6 +199,9 @@ services:
       WAIT_HOSTS_TIMEOUT: 300
       WAIT_SLEEP_INTERVAL: 30
       WAIT_HOST_CONNECT_TIMEOUT: 30
+      SIMULATION_MODE: TRUE
+      CONFIG_FILE_PATH: ../manifest.json
```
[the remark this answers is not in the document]
paulbourelly999: Its used by the streets_service base class.

streets_utils/streets_service_base/src/streets_clock_singleton.cpp
paulbourelly999: We can not add it directly to the carma-clock object in any neat way since the carma-time-lib is being used by both V2X-Hub and CARMA-Streets and both use a different logger.

streets_utils/streets_service_configuration/include/internal/streets_singleton.tpp
```diff
@@ -6,45 +6,30 @@ namespace streets_service {
...
       
       }
-      instance = new T(args...);
-      return *instance;
+      else {
+        SPDLOG_INFO("Initializing Singleton of type {0}!", typeid(instance.get()).name());
+        instance = std::unique_ptr<T>( new T(args...) );
+      }
+      
+      return *instance.get();
```
kjrush, line 31: This looks weird, but probably is fine? If I understand the order-of-operations here, this should be `*(instance.get())` effectively, where `.get()` returns a pointer to the contained object and it is then dereferenced (before then being implicitly re-referenced by the `T&` return signature?) If that's correct, is there a reason to not just `return *instance`? `std::unique_ptr` overloads the dereference operator directly. I think the net result is the same, so not going to block merging over this if you want to do so, but just reads a little clunky.

paulbourelly999: I did not know that smart pointers had overloads for the dereference operator. Honestly still one of the parts of C++ I am not super good with but good catch. Addressed in both the get_singleton and create methods

PR #345, opened 2023-09-11
Add defaults for environment variables

paulbourelly999:
# PR Details
## Description

Adding default values for environment variables to make CARMA Street services more deployable. 
## Related Jira Key
[CDAR-332
](https://usdot-carma.atlassian.net/browse/CDAR-332
)

## Motivation and Context
Improve deployment process for CARMA-Streets

## How Has This Been Tested?
Tested using AWS XIL-Cloud deployment

## Types of changes

## Checklist:

### DIFF
--- a/streets_utils/streets_service_base/include/streets_environment_variables.h
+++ b/streets_utils/streets_service_base/include/streets_environment_variables.h
@@ -6,10 +6,19 @@ namespace streets_service {
 
     static const std::string LOGS_DIRECTORY_ENV = "LOGS_DIRECTORY";
 
+    static const std::string DEFAULT_LOGS_DIRECTORY = "../logs/";
+
     static const std::string SIMULATION_MODE_ENV = "SIMULATION_MODE";
 
+    static const std::string DEFAULT_SIMULATION_MODE = "FALSE";
+
     static const std::string CONFIG_FILE_PATH_ENV = "CONFIG_FILE_PATH";
+    
+    static const std::string DEFAULT_CONFIG_FILE_PATH = "../manifest.json";
 
     static const std::string TIME_SYNC_TOPIC_ENV = "TIME_SYNC_TOPIC";
+
+    static const std::string DEFAULT_TIME_SYNC_TOPIC = "time_sync";
+
     
 }
--- a/streets_utils/streets_service_base/include/streets_environment_variables.h
+++ b/streets_utils/streets_service_base/include/streets_environment_variables.h
@@ -4,21 +4,21 @@
 
 namespace streets_service {
 
-    static const std::string LOGS_DIRECTORY_ENV = "LOGS_DIRECTORY";
+    inline static const std::string LOGS_DIRECTORY_ENV = "LOGS_DIRECTORY";
 
-    static const std::string DEFAULT_LOGS_DIRECTORY = "../logs/";
+    inline static const std::string DEFAULT_LOGS_DIRECTORY = "../logs/";
 
-    static const std::string SIMULATION_MODE_ENV = "SIMULATION_MODE";
+    inline static const std::string SIMULATION_MODE_ENV = "SIMULATION_MODE";
 
-    static const std::string DEFAULT_SIMULATION_MODE = "FALSE";
+    inline static const std::string DEFAULT_SIMULATION_MODE = "FALSE";
 
-    static const std::string CONFIG_FILE_PATH_ENV = "CONFIG_FILE_PATH";
+    inline static const std::string CONFIG_FILE_PATH_ENV = "CONFIG_FILE_PATH";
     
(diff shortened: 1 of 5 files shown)


streets_utils/streets_service_base/src/streets_service.cpp
```diff
@@ -110,16 +110,16 @@ namespace streets_service {
...
                 return config;
             }
             catch(const std::logic_error &e) {
-                std::string config_name_str = config_name;
-                throw std::runtime_error("System config " + config_name_str + " not set!");
+                SPDLOG_WARN("System config {0} was not set! Using default value {1}!" ,config_name, default_val);
+                return default_val;
             }
         } else {
             throw std::runtime_error(" Systme config param name is null pointer!");
```
adamlm, line 125: This function's implementation needs refactoring. Right now, if the caller passes `nullptr` for `config_name`, the function will throw an error instead of returning `default_val`. Is this the intended behavior? That seems unexpected to me.

We can check for `config_name == nullptr` at the function's beginning and return `default_val` value if `true`. This allows us to avoid throwing a `std::runtime_error`. We can also clean up unnecessary nesting the function's main logic in an `if` statement.

The `try...catch` block pushes the error to the wrong place: the error should be checked on the result of `std::getenv`, not when trying to construct the `std::string`. Function `std::getenv` returns a `nullptr` is the environment variable is not found, so not checking for it is poor practice. See [cppreference](https://en.cppreference.com/w/cpp/utility/program/getenv) for details. We can refactor the main logic into something like this:

```cpp
if (const auto config = std::getenv(config_name) {
    SPDLOG_DEBUG("Reading system config {0} as : {1}!", config_name, config);
    return config;
}

SPDLOG_WARN("System config {0} was not set! Using default value {1}!" ,config_name, default_val);
return default_val;
```
With the above changes, we can also declare `get_system_config` as `noexcept`.

adamlm (changes requested):

willjohnsonk (approved):
Looks good

willjohnsonk (commented):
See: Adam's suggestions (was not finished with review when approved)

streets_utils/streets_service_base/src/streets_service.cpp
```diff
@@ -111,20 +111,21 @@ namespace streets_service {
...
+        }
+        // If std::getenv(config_name) returns value, use this value
+        if (const auto config = std::getenv(config_name)) {
+            SPDLOG_DEBUG("Reading system config {0} as : {1}!", config_name, config);
+            return config;
+        }
+        // If std::getenv(config_name) returns nullptr, environment variable was not set so use default
+        else {
+            SPDLOG_WARN("System config {0} was not set! Using default value {1}!" ,config_name, default_val);
+            return default_val;
```
adamlm, line 127: The comments above each part of the `if` statements are unnecessary and should be removed. Comments should explain "why" instead of "what." In my opinion, the code here is sufficiently self documenting.

The `else` statement should be removed because it unnecessarily clutters the code. If either `if` statement evaluates to `true`, the contained `return` statements will exit the function early when necessary, so there is no need to scope the final `reutrn`.

There is a bit of code duplication between the `nullptr` and the unset `config_name` cases. I believe we can change the first `if` case's implementation to something like this:

```cpp
if (config_name == nullptr) {
    config_name = "";
}
```

Then we can check `config_name` as if it had been set. The `SPDLOG_WARN` about `nullptr` can then be removed or at least changed to `SPDLOG_DEBUG`.

streets_utils/streets_service_base/include/streets_environment_variables.h
```diff
@@ -4,21 +4,21 @@
...
+    inline static const std::string CONFIG_FILE_PATH_ENV = "CONFIG_FILE_PATH";
     
-    static const std::string DEFAULT_CONFIG_FILE_PATH = "../manifest.json";
+    inline static const std::string DEFAULT_CONFIG_FILE_PATH = "../manifest.json";
 
-    static const std::string TIME_SYNC_TOPIC_ENV = "TIME_SYNC_TOPIC";
+    inline static const std::string TIME_SYNC_TOPIC_ENV = "TIME_SYNC_TOPIC";
 
-    static const std::string DEFAULT_TIME_SYNC_TOPIC = "time_sync";
+    inline static const std::string DEFAULT_TIME_SYNC_TOPIC = "time_sync";
```
adamlm, line 21: Remove the `static` declarations. Declaring `inline static` essentially defeats the purpose of using `inline` variables. Check out these two StackOverflow posts for more information: [1](https://stackoverflow.com/questions/50515591/c17-inline-variable-vs-inline-static-variable), [2](https://stackoverflow.com/questions/38043442/how-do-inline-variables-work/47502744#47502744)

adamlm (changes requested):

adamlm (approved):

PR #352, opened 2023-09-29
Added JSON Utility Library to carma-streets

paulbourelly999:
# PR Details
## Description

Added new json utility library to CARMA Streets.
## Related Jira Key
[CDAR-298
](https://usdot-carma.atlassian.net/browse/CDAR-298)

## Motivation and Context
Provide utility methods for parsing json. Possibly in the future this library could abstract away the dependency on third-party json parsing library (rapidjson) to make code more modular and allow for easier transition to other alternative JSON parsing libraries.

## How Has This Been Tested?
Unit testing 

## Types of changes

## Checklist:


streets_utils/json_utils/README.md
```diff
@@ -0,0 +1,28 @@
...
+`std::optional<rapidjson::Value::ConstObject> parse_object_member(const std::string &member_name, const rapidjson::Value &doc, bool required )`
+`std::optional<rapidjson::Value::ConstArray> parse_array_member(const std::string &member_name, const rapidjson::Value &doc, bool required)`
+
+Functions to retrieve member values from JSON object [RapidJSON](https://miloyip.github.io/rapidjson/index.html) with [DOM  parsing](https://miloyip.github.io/rapidjson/md_doc_dom.html). Functions will return unintialized `std::optional` for optional members that are not found in JSON object and will throw `json_utils_exception` for required members that are not found in JSON object.
+
+## Include Library
+To include after installing with `make install` simply add find_package call to `CMakeLists.txt` and link the library as follows. Installed CMake configuration files should handle finding and linking depedencies `RapidJSON` and `spdlog`. The library along with it's dependencies can then be included by simply using the find_package() instruction.
+```
+find_package(streets_utils COMPONENT json_utils)
```
paulbourelly999, line 25: @adamlm I mentioned this to you a while back but I not been able to figure out the namespace stuff for CMake. If you could help me with this that would be awesome. Currently CMake install installs this package as `json_utils_lib::json_utils_lib". I probably want to move to something more like 'streets_utils::json_utils_lib'. Is it possible to make this change without making it for all the other packages inside streets_utils yet?

streets_utils/json_utils/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+    EXPORT ${LIBRARY_NAME}_libTargets
+    LIBRARY DESTINATION lib
+    INCLUDES DESTINATION include
+    ARCHIVE DESTINATION lib
+)
+install(
+    EXPORT ${LIBRARY_NAME}_libTargets 
+    FILE ${LIBRARY_NAME}_libTargets.cmake
+    DESTINATION lib/cmake/${LIBRARY_NAME}_lib/
+    NAMESPACE ${LIBRARY_NAME}_lib::
```
paulbourelly999, line 48: Would changing this to using a different namespace be as simple as just changing this namespace parameter? Would it be best practice to make the destination include a parent directory to match the namespace like for example change destination to `lib/cmake/streets_utils/${library_name}_lib/` if we plan on using a namespace like `streets_utils`?

streets_utils/json_utils/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+    PUBLIC
+        rapidjson
+        spdlog::spdlog
+
+)  
+
+message( STATUS "Target INCLUDE DIR : " ${INCLUDE_DIRECTORIES})
+
+########################################################
+# Install sensor_data_sharing_package package.
```
paulbourelly999, line 33: Correct comment

streets_utils/json_utils/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+        $<INSTALL_INTERFACE:include>
+)
+target_link_libraries( ${LIBRARY_NAME}_lib 
+    PUBLIC
+        rapidjson
+        spdlog::spdlog
+
+)  
+
+message( STATUS "Target INCLUDE DIR : " ${INCLUDE_DIRECTORIES})
```
paulbourelly999, line 30: Remove message

streets_utils/json_utils/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
+cmake_minimum_required(VERSION 3.10.2)
+project(streets_json_utils)
+# Set Flags
+set(CMAKE_CXX_STANDARD 20)
```
adamlm, line 4: CARMA targets C++17. I've run into issues trying C++20 with other CARMA packages. We should probably stick with C++17 for now.

streets_utils/json_utils/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+        spdlog::spdlog
+
+)  
+
+message( STATUS "Target INCLUDE DIR : " ${INCLUDE_DIRECTORIES})
+
+########################################################
+# Install sensor_data_sharing_package package.
+########################################################
+file(GLOB files ${CMAKE_CURRENT_SOURCE_DIR}/include/*.hpp)
```
adamlm, line 35: Globbing here is unnecessary. This library has only two header files, so they should be explicitly listed.

streets_utils/json_utils/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+install(
+    FILES ${CMAKE_CURRENT_BINARY_DIR}/${LIBRARY_NAME}_libConfig.cmake
+    DESTINATION  lib/cmake/${LIBRARY_NAME}_lib/
+)
+install(FILES ${files} DESTINATION include)
+
+########################
+# googletest for unit testing
+########################
+set(BINARY ${LIBRARY_NAME}_test)
```
adamlm, line 64: This is unnecessary. It increases mental overhead without significantly saving on typing. The variable name is also too vague. We should replaced `${BINARY}` with `json_utils_test`.

streets_utils/json_utils/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+file(GLOB_RECURSE TEST_SOURCES LIST_DIRECTORIES false test/*.hpp test/*.cpp)
```
adamlm, line 65: Unnecessary globbing.

streets_utils/json_utils/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+set(SOURCES ${TEST_SOURCES} WORKING_DIRECTORY  ${PROJECT_SOURCE_DIR}/test)
```
adamlm, line 66: Unnecessary variable.

streets_utils/json_utils/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+add_executable(${BINARY} ${TEST_SOURCES})
+add_test(NAME ${BINARY} COMMAND ${BINARY})
+target_include_directories(${BINARY} PUBLIC ${PROJECT_SOURCE_DIR}/include)
+target_link_libraries(${BINARY} 
+    PUBLIC
```
adamlm, line 71: This should be marked `PRIVATE` because no downstream entities will link against this.

streets_utils/json_utils/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+
+########################
+# googletest for unit testing
+########################
+set(BINARY ${LIBRARY_NAME}_test)
+file(GLOB_RECURSE TEST_SOURCES LIST_DIRECTORIES false test/*.hpp test/*.cpp)
+set(SOURCES ${TEST_SOURCES} WORKING_DIRECTORY  ${PROJECT_SOURCE_DIR}/test)
+add_executable(${BINARY} ${TEST_SOURCES})
+add_test(NAME ${BINARY} COMMAND ${BINARY})
+target_include_directories(${BINARY} PUBLIC ${PROJECT_SOURCE_DIR}/include)
```
adamlm, line 69: This should be removed. Linking against `json_utils` will provide CMake with information on how to find those headers.

streets_utils/json_utils/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+install(FILES ${files} DESTINATION include)
+
+########################
+# googletest for unit testing
+########################
+set(BINARY ${LIBRARY_NAME}_test)
+file(GLOB_RECURSE TEST_SOURCES LIST_DIRECTORIES false test/*.hpp test/*.cpp)
+set(SOURCES ${TEST_SOURCES} WORKING_DIRECTORY  ${PROJECT_SOURCE_DIR}/test)
+add_executable(${BINARY} ${TEST_SOURCES})
+add_test(NAME ${BINARY} COMMAND ${BINARY})
```
adamlm, line 68: This should be replaced with `gtest_discover_tests()`.

streets_utils/json_utils/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+        ${LIBRARY_NAME}_lib
+        rapidjson
+        spdlog::spdlog
+
+ )
+if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.20") 
+    target_link_libraries( ${BINARY} PUBLIC GTest::gtest_main )
+else() 
+    target_link_libraries( ${BINARY} PUBLIC gtest_main )
+endif()
```
adamlm, line 81: This is unnecessary. We should either require CMake version 3.20+ and link against `GTest::gtest_main` or use `GTest::Main` for backwards compatibility.

streets_utils/json_utils/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+    NAMESPACE ${LIBRARY_NAME}_lib::
+)
+include(CMakePackageConfigHelpers)
+configure_package_config_file(
+    cmake/${LIBRARY_NAME}_libConfig.cmake.in 
+    ${CMAKE_CURRENT_BINARY_DIR}/${LIBRARY_NAME}_libConfig.cmake
+    INSTALL_DESTINATION  lib/${LIBRARY_NAME}_lib/ )
+install(
+    FILES ${CMAKE_CURRENT_BINARY_DIR}/${LIBRARY_NAME}_libConfig.cmake
+    DESTINATION  lib/cmake/${LIBRARY_NAME}_lib/
```
adamlm, line 57: This should be `${CMAKE_INSTALL_LIBDIR}/cmake/json_utils`. You will need to `include(GNUInstallDirs)`.

streets_utils/json_utils/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+)
+install(FILES ${files} DESTINATION include)
```
adamlm, line 59: The `DESTINATION` should be `${CMAKE_INSTALL_INCLUDEDIR}/cmake/json_utils`.

streets_utils/json_utils/include/json_utils.hpp
```diff
@@ -0,0 +1,149 @@
+#pragma once
```
adamlm, line 1: We should use `#include` guards instead of `#pragma once` because the latter is not standard, even though many compilers support it.

streets_utils/json_utils/include/json_utils_exception.hpp
```diff
@@ -0,0 +1,26 @@
...
+     * @brief Runtime error related to processing signal_phase_and_timing JSON updates.
+     * 
+     * @author Paul Bourelly 
+     */ 
+    class json_utils_exception : public std::runtime_error{
+        public:
+            /**
+             * @brief Destructor.
+             */ 
+            ~json_utils_exception() = default;
```
adamlm, line 18: I believe this should be declared `override` since it's overriding `std::runtimte_error`'s destructor.

streets_utils/json_utils/test/json_utils_test.cpp
```diff
@@ -0,0 +1,289 @@
+#include <gtest/gtest.h>
+#include <json_utils.hpp>
+#include <json_utils_exception.hpp>
+
+using namespace streets_utils::json_utils;
+
+//---------------------test validate_json---------------------
+TEST(json_utils_test, test_validate_invalid_json) {
```
adamlm, line 8: This violates GTest's naming requirements. Relevant quote from the [documentation](http://google.github.io/googletest/primer.html):

>TEST() arguments go from general to specific. The first argument is the name of the test suite, and the second argument is the test’s name within the test suite. Both names must be valid C++ identifiers, and they should not contain any underscores (_). A test’s full name consists of its containing test suite and its individual name. Tests from different test suites can have the same individual name.

Issue applies to all test cases in this file.

streets_utils/json_utils/src/json_utils.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+    rapidjson::Document validate_json(const std::string &json) {
+        rapidjson::Document obj;
+        obj.Parse(json);
+        if (obj.HasParseError())
+        {
+            throw json_utils_exception("Message JSON is misformatted. JSON parsing failed!");
+        }
+        return obj;
+    }
```
adamlm, line 13: This function should be refactored to something like `is_valid`. The current implementation does not take any validation steps, it only throws an error if the JSON string is invalid.

adamlm (changes requested):

streets_utils/json_utils/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+# Set Flags
+set(CMAKE_CXX_STANDARD 20)
+# Find Packages 
+find_package(RapidJSON REQUIRED)
+find_package(GTest REQUIRED)
+find_package(spdlog REQUIRED)
+
+# Add definition for rapidjson to include std::string
+add_definitions(-DRAPIDJSON_HAS_STDSTRING=1)
+set(LIBRARY_NAME json_utils)
```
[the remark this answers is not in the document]
paulbourelly999: I think this reduces unintentional errors introduces with library name changes or install configuration changes. I also do not think it makes this at all harder to read.

streets_utils/json_utils/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+########################################################
+# Build Library
+######################################################## 
+file(GLOB SOURCES ${CMAKE_CURRENT_SOURCE_DIR}/src/*.cpp)
```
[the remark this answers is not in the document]
paulbourelly999: My thought process for CMakeLists.txt in carma-streets is to attempt to have a similar template that works for any library created. If we were to add another cpp file in the future, a setup like this would avoid any time spent, setting up the CMake config for it.

streets_utils/json_utils/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+find_package(RapidJSON REQUIRED)
+find_package(GTest REQUIRED)
+find_package(spdlog REQUIRED)
+
+# Add definition for rapidjson to include std::string
+add_definitions(-DRAPIDJSON_HAS_STDSTRING=1)
+set(LIBRARY_NAME json_utils)
+########################################################
+# Build Library
+########################################################
```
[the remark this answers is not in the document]
paulbourelly999: I think it is helpful to have some kind of separation between the different build/install/test portions of the CMakeLists.txt . I am open to alternative comments or ways of separating this but I don't think there is any disadvantage in having this here

streets_utils/json_utils/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+
+message( STATUS "Target INCLUDE DIR : " ${INCLUDE_DIRECTORIES})
+
+########################################################
+# Install sensor_data_sharing_package package.
+########################################################
+file(GLOB files ${CMAKE_CURRENT_SOURCE_DIR}/include/*.hpp)
+
+install(
+    TARGETS ${LIBRARY_NAME}_lib
```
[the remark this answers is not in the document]
paulbourelly999: This is a remnant of using project_name as target_name. Will remove and add to library name directly.

streets_utils/json_utils/CMakeLists.txt
paulbourelly999: I agree that this is vague but not that this variable should be replaced with direct calls to a library name. When setting up these CMakeLists.txt files for different CARMA-Streets libraries I want all the libraries to follow a similar approach for installing, naming convention for targets. This helps a great deal with CI integration later one because instead of 30 independent libraries that each have a CMakeLists.txt that handles installing, naming convention differently, using these variables makes it a lot easier to standup CI to execute tests and do other CI tasks.

With the previous comment, one advantage of using project name to set target names for library and test executable was that it made following a common naming convention easy. With that in mind I kind of what to do something like:

```
project (<project_name)
...
set(LIBRARY_NAME ${PROJECT_NAME}_lib)
...
set(TEST_NAME ${PROJECT_NAME}_test)
```

This would mean our library name and test name are based on our project name but still separate. This way when someone adds project with some name CI processes can rely on the libraries and executables named a certain way. What do you think?

streets_utils/json_utils/CMakeLists.txt
paulbourelly999: I think this is helpful instead of explicitly naming files since then we have to make CMakeLists.txt changes on every change/addition to test files. Also it allows this CMakeLists.txt to be easily portable to a new streets library.

streets_utils/json_utils/CMakeLists.txt
paulbourelly999: I think its useful. See other Globbing comments

streets_utils/json_utils/include/json_utils.hpp
paulbourelly999: Since we deliver this code as a runnable docker image unless `#pragma once` is being depreciated or we plan on switching to a compiler where this is not supported I do not see the purpose of removing this. In what kind of context would we need to support compiling this library with a compiler that does not support pragma once.

streets_utils/json_utils/src/json_utils.cpp
paulbourelly999: Good call.

streets_utils/json_utils/include/json_utils.hpp
paulbourelly999: After discussion I think we should table this change until we have a plan for how we want to handle styling. Using pragma once would potentially conflict with ROS standard clang-format which may be the eventual formatting standard we want all our code to adhere to but that is still outstanding.

streets_utils/json_utils/CMakeLists.txt
adamlm: With your suggestion, a library developer can still (unintentionally) set `LIBRARY_NAME` and `TEST_NAME` to something that doesn't follow naming conventions. Having consistent naming across projects is good, but this is not the mechanism to enforce it.

streets_utils/json_utils/CMakeLists.txt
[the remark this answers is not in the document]
adamlm: I disagree with the first part of that argument. CMake will throw an error if the target name is undefined. Also, the library name will not change frequently (if ever) after the project is established.

streets_utils/json_utils/CMakeLists.txt
adamlm: I'm not sure what you mean by "easily portable." `CMakeLists.txt` files are specific to the project. If you want some common functionality that can be reused by other projects, it should be wrapped in a function. Adding test files is relatively infrequent compared to changing code implementations within files. Therefore, I don't see a significant utility in globbing here. Globbing prevents you from easily excluding tests for debugging or other purposes.

Another thing I noticed, header files should not added to `TEST_SOURCES`.

streets_utils/json_utils/CMakeLists.txt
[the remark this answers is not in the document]
paulbourelly999: Not if it is undefine but say if you type the library name correctly for target but slightly incorrectly for the directory to install in there is no error. Or you type the cmake config in file name slightly different than the library name or any other small change

streets_utils/json_utils/CMakeLists.txt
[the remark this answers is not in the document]
paulbourelly999: I do not think explicit sources makes it more readable. In fact I think it makes it less readable. All the files in source should automatically be included. If you have a package with a single source that compiles into several dependent targets you should just separate them into separate directories. I think that is the most readable approach. Although I think calling it library_name sources is a good improvement to avoid name conflicts.

streets_utils/json_utils/include/json_utils_exception.hpp
```diff
@@ -6,21 +6,14 @@
...
-             */ 
-            ~json_utils_exception() = default;
-            /**
-             * @brief Constructor. 
-             * @param msg String exception message.
-             */  
-            explicit json_utils_exception(const std::string &msg ) : std::runtime_error(msg){};
-
+    class json_parse_exception : public std::runtime_error{
+        using std::runtime_error::runtime_error;
```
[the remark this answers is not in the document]
paulbourelly999: This seems less verbose and was a recommend change from sonar scanner for extending the runtime_error

streets_utils/json_utils/CMakeLists.txt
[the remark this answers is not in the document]
adamlm: Configuration issues would/should be discovered when doing a test install locally. That part of the configuration is unlikely to change, so I don't think that part of the argument is strong. I've gone through CMake projects before that use a lot of variables like this, and it made understanding all of the configurations/targets challenging. For a rather simple project like this, it probably does not make a difference. However, if we are trying to establish common conventions across all our projects, the variable approach is not scalable.

streets_utils/json_utils/include/json_utils_exception.hpp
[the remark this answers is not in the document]
adamlm: Makes sense. I was not aware of this technique before, but it appears to be a standard/best practice.

adamlm (approved):

PR #354, opened 2023-10-10
Update json_utils methods to use C++ primitives

paulbourelly999:
# PR Details
## Description

Replace std library fix with types with C++ primitives to avoid implicit or explicit casting of returned value.

## Related Jira Key
[CDAR-298
](https://usdot-carma.atlassian.net/browse/CDAR-298)

## Motivation and Context
Cleaner implementation also avoid sonar code smells 

## How Has This Been Tested?
Unit testing

## Types of changes

## Checklist:

### DIFF
--- a/streets_utils/json_utils/include/json_utils.hpp
+++ b/streets_utils/json_utils/include/json_utils.hpp
@@ -41,9 +41,9 @@ namespace streets_utils::json_utils {
      * @param obj JSON object with member
      * @param required bool flag to indicate whether member is required by calling code.
      * @throws json_parse_exception if member is required but not found.
-     * @return std::optional<uint> 
+     * @return std::optional<unsigned int> 
      */
-    std::optional<uint> parse_uint_member(const std::string &member_name,  const rapidjson::Value &obj, bool required );
+    std::optional<unsigned int> parse_uint_member(const std::string &member_name,  const rapidjson::Value &obj, bool required );
      /**
      * @brief Functions to retrieve bool member values from JSON object [RapidJSON](https://miloyip.github.io/rapidjson/index.html)
      * with [DOM  parsing](https://miloyip.github.io/rapidjson/md_doc_dom.html). Functions will return unintialized
--- a/streets_utils/json_utils/include/json_utils.hpp
+++ b/streets_utils/json_utils/include/json_utils.hpp
@@ -29,7 +29,7 @@ namespace streets_utils::json_utils {
      * @param obj JSON object with member
      * @param required bool flag to indicate whether member is required by calling code.
      * @throws json_parse_exception if member is required but not found.
-     * @return std::optional<int64_t> 
+     * @return std::optional<int> 
      */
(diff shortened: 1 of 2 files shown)


streets_utils/json_utils/include/json_utils.hpp
```diff
@@ -41,9 +41,9 @@ namespace streets_utils::json_utils {
      * @param obj JSON object with member
      * @param required bool flag to indicate whether member is required by calling code.
      * @throws json_parse_exception if member is required but not found.
-     * @return std::optional<uint64_t> 
+     * @return std::optional<uint> 
      */
-    std::optional<uint64_t> parse_uint_member(const std::string &member_name,  const rapidjson::Value &obj, bool required );
+    std::optional<uint> parse_uint_member(const std::string &member_name,  const rapidjson::Value &obj, bool required );
```
adamlm, line 46: `uint` is a non-standard type. It should be `unsinged int`.

streets_utils/json_utils/src/json_utils.cpp
```diff
@@ -24,7 +24,7 @@ namespace streets_utils::json_utils {
         return member;
     }
 
-    std::optional<uint64_t> parse_uint_member(const std::string &member_name,  const rapidjson::Value &obj, bool required ) {
+    std::optional<uint> parse_uint_member(const std::string &member_name,  const rapidjson::Value &obj, bool required ) {
```
adamlm, line 27: `uint` is a non-standard type. It should be `unsigned int`.

adamlm (changes requested):

streets_utils/json_utils/include/json_utils.hpp
paulbourelly999: Fixing this but trying to understand why I have access to these non standard types in the first place. It seems they are define in a types.h file under `/usr/include/x86_64-linuz-gnu/sys/types.h` as follows :
```
#ifdef __USE_MISC
/* Old compatibility names for C types.  */
typedef unsigned long int ulong;
typedef unsigned short int ushort;
typedef unsigned int uint;
#endif

/* These size-specific names are used by some of the inet code.  */

#include <bits/stdint-intn.h>

/* These were defined by ISO C without the first `_'.  */
typedef __uint8_t u_int8_t;
typedef __uint16_t u_int16_t;
typedef __uint32_t u_int32_t;
typedef __uint64_t u_int64_t;
```

Is this a compiler specific header that has typedef short hands for the c++ standard types? Is it bad practice to use  any typedefs define by compiler specific headers like this? Then why have the compiler specific header file in the first place. The license at the to seems to suggest its part of the GNU C Library, which when looking online is just an implementation of the std  C library plus some helpful utilities. I can see why using types from here is bad practice. I guess I just don't understand why some compilers would do a typedef like this if it then makes your code inherently non portable between compilers.

adamlm (approved):

PR #361, opened 2023-11-11
Initial working comment

paulbourelly999:
# PR Details
## Related Jira Key
CDAR-389

## Motivation and Context
Create base image for Sensor Data Sharing Service and any other carma-streets services that require lanelet2

## How Has This Been Tested?
Locally built base image

## Types of changes

## Checklist:

- [ ] I have read the [**CONTRIBUTING**](https://github.com/usdot-fhwa-stol/carma-platform/blob/develop/Contributing.md) document.

### DIFF
--- a/build_scripts/install_lanelet2_dependencies.sh
+++ b/build_scripts/install_lanelet2_dependencies.sh
@@ -1,0 +1,66 @@
+#!/bin/bash
+# Script assumes base image of carma-builds-64x
+set -e
+# Get ubuntu distribution code name. All STOL APT debian packages are pushed to S3 bucket based on distribution codename.
+. /etc/lsb-release
+# add the STOL APT repository
+echo "deb [trusted=yes] http://s3.amazonaws.com/stol-apt-repository ${DISTRIB_CODENAME} main" > /etc/apt/sources.list.d/stol-apt-repository.list
+echo "deb http://packages.ros.org/ros/ubuntu ${DISTRIB_CODENAME} main" > /etc/apt/sources.list.d/ros-latest.list
+DEBIAN_FRONTEND=noninteractive apt install --no-install-recommends --yes --quiet gnupg
+wget -qO- https://raw.githubusercontent.com/ros/rosdistro/master/ros.asc | apt-key add -
+apt-get update
+DEPENDENCIES=(
+    autotools-dev 
+    automake 
+    sqlite3 
+    libsqlite3-dev 
+    curl 
+    libpugixml-dev 
+    libgeographic-dev
+    ros-melodic-catkin
+    python-rospkg
+    libeigen3-dev
+    libtool
+    libboost-all-dev
+)
+DEBIAN_FRONTEND=noninteractive apt install --no-install-recommends --yes --quiet "${DEPENDENCIES[@]}"
+# Install PROJ, a package for coordinate transformations
+git clone https://github.com/OSGeo/PROJ.git /tmp/PROJ --branch 6.2.1 
+cd /tmp/PROJ  
+./autogen.sh 
+./configure 
+make 
+make install
+cd ..
+rm -r PROJ
+        
+# Download a cmake module for PROJ
+wget -P /usr/local/share/cmake-3.27/Modules https://raw.githubusercontent.com/mloskot/cmake-modules/master/modules/FindPROJ4.cmake
+
+
+# Once catkin is installed only the required lanelet2 packages will be pulled in from carma
+# NOTE: The lanelet2_python package requires additional dependencies that have not yet been installed so it is removed for now
+mkdir carma_lanelet2
+cd carma_lanelet2 
+mkdir src 
+cd src 
+git init 
+echo "temp" 
+git remote add origin -f https://github.com/usdot-fhwa-stol/autoware.ai.git 
+git config core.sparsecheckout true	&& \
+echo "common/hardcoded_params/*" >> .git/info/sparse-checkout 
+echo "common/lanelet2_extension/*" >> .git/info/sparse-checkout 
+echo "lanelet2/*" >> .git/info/sparse-checkout 
+echo "mrt_cmake_modules/*" >> .git/info/sparse-checkout 
+git pull --depth 1 origin refactor_lanelet2_extension 
+git checkout refactor_lanelet2_extension 
+rm -r lanelet2/lanelet2_python
+rm -r lanelet2/lanelet2_examples
+cd ..
+
+# In order to trick lanelet2 into building the ROS_VERSION environment variable must be set
+# In order to fully decouple lanelet2_extension from ros the LANELET2_EXTENSION_LOGGER_TYPE environment variable must be set
+source /opt/ros/melodic/setup.bash  
+ROS_VERSION=1 LANELET2_EXTENSION_LOGGER_TYPE=1 catkin_make install
+cd ..
+rm -r carma_lanelet2
--- a/streets_utils/streets_service_base/CMakeLists.txt
+++ b/streets_utils/streets_service_base/CMakeLists.txt
@@ -80,7 +80,7 @@ add_test(NAME ${BINARY} COMMAND ${BINARY})
 target_link_libraries(${BINARY} 
     PUBLIC
         ${PROJECT_NAME}_lib                                   
-        GTest::gtest
+        GTest::GTest
 )
 target_include_directories(${BINARY} 
     PUBLIC 
--- a/streets_service_base_lanelet_aware/Dockerfile
+++ b/streets_service_base_lanelet_aware/Dockerfile
@@ -1,0 +1,4 @@
+FROM ghcr.io/usdot-fhwa-stol/carma-builds-x64:bionic
+COPY ./build_scripts /home/carma-streets/build_scripts
+ENV LANELET2_MAP="/home/carma-streets/MAP/Intersection.osm"
+RUN /home/carma-streets/build_scripts/install_lanelet2_dependencies.sh
--- a/streets_service_base_lanelet_aware/README.md
+++ b/streets_service_base_lanelet_aware/README.md
@@ -1,0 +1,18 @@
+# Streets Service Base Lanelet Aware
+## Introduction
+This base image is intented for **CARMA Streets** services that require [lanelet2](https://github.com/fzi-forschungszentrum-informatik/Lanelet2) for spatial understanding of the surrouding area. Currently both **Message Service** and **Intersection Model** both depend on [lanelet2](https://github.com/fzi-forschungszentrum-informatik/Lanelet2) for this purpose. The new propose **Sensor Data Sharing Service** will also need lanelet2 for map transforms and for eventual perception. To allow all three services and any future services easy access to this dependency, this new **Streets Service Base Lanelet Aware** image exist.
+> [!IMPORTANT]\
+> Currently this base image is only being used for the **Sensor Data Sharing Service** but both the **Intersection_Model** and **Message Service** should be ported to this base image as well.
+
+## Usage
+Simply use the following to extend this image
+```
+FROM usdotfhwastoldev/message_services:<tag>
+```
+The image also contains the environment variable **LANELET2_MAP**, which is assigned a default value of `/home/carma-streets/MAP/Intersection.osm`, but can be overwritten using the docker-compose environment variable section of the service description. An example is shown belo
+```
+ lanlet_aware_service:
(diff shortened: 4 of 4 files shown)


streets_utils/streets_service_base/CMakeLists.txt
```diff
@@ -80,7 +80,7 @@ add_test(NAME ${BINARY} COMMAND ${BINARY})
 target_link_libraries(${BINARY} 
     PUBLIC
         ${PROJECT_NAME}_lib                                   
-        GTest::gtest
+        GTest::GTest
 )
```
adamlm, line 84: These should be `PRIVATE` since nothing is going to link against the test executable.

build_scripts/install_lanelet2_dependencies.sh
```diff
@@ -0,0 +1,70 @@
...
+git checkout refactor_lanelet2_extension 
+rm -r lanelet2/lanelet2_python
+rm -r lanelet2/lanelet2_examples
+cd ..
+
+# In order to trick lanelet2 into building the ROS_VERSION environment variable must be set
+# In order to fully decouple lanelet2_extension from ros the LANELET2_EXTENSION_LOGGER_TYPE environment variable must be set
+# shellcheck source=/dev/null
+source /opt/ros/melodic/setup.bash  
+ROS_VERSION=1 LANELET2_EXTENSION_LOGGER_TYPE=1 catkin_make install
```
adamlm, line 68: Does that have to be `catkin` here, or can we use `colcon`?

build_scripts/install_streets_service_base_dependencies.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
+set -e
+# Get ubuntu distribution code name. All STOL APT debian packages are pushed to S3 bucket based on distribution codename.
+# shellcheck source=/dev/null
+source /etc/lsb-release
+# add the STOL APT repository
+echo "deb [trusted=yes] http://s3.amazonaws.com/stol-apt-repository ${DISTRIB_CODENAME} main" > /etc/apt/sources.list.d/stol-apt-repository.list
```
adamlm, line 7: I think it would be cleaner and more readable to use the actual `lsb_release` command to get the distribution's codename:

```shell
lsb_release --codename --short
```

The apt config would then look like:

```shell
echo "deb [trusted=yes] http://s3.amazonaws.com/stol-apt-repository $(lsb_release --codename --short) main" > /etc/apt/sources.list.d/stol-apt-repository.list

File: build_scripts/install_streets_service_base_dependencies.sh [hunk 4]
```diff
@@ -0,0 +1,48 @@
...
+apt-get update
```
adamlm, line 8: Apparently `apt-get` is more a lower-level backend for installing packages. `apt` is the frontend and seems to be the recommended command to use. I suggest we use that for consistency.

More information here: https://askubuntu.com/questions/445384/what-is-the-difference-between-apt-and-apt-get

build_scripts/install_streets_service_base_dependencies.sh
```diff
@@ -0,0 +1,48 @@
...
+)
+
+DEBIAN_FRONTEND=noninteractive apt install --no-install-recommends --yes --quiet "${DEPENDENCIES[@]}"
+
+# Install librdkafka
+echo " ------> Install librdkafka..."
+cd /tmp
+git clone https://github.com/confluentinc/librdkafka.git -b v2.2.0
+cd librdkafka/
+cmake -H. -B_cmake_build
```
adamlm, line 21: Why are we using the `-H.` flag here? It seems to be an undocumented flag, so I suggest we avoid using it unless we have a compelling reason.

More info here: https://stackoverflow.com/questions/31090821/what-does-the-h-option-means-for-cmake

build_scripts/install_streets_service_base_dependencies.sh
```diff
@@ -0,0 +1,48 @@
...
+cmake --build _cmake_build
```
adamlm, line 22: Any particular reason we are using `_cmake_build`? If not, I suggest we use `build` to stay consistent with other CMake projects.

build_scripts/install_streets_service_base_dependencies.sh
```diff
@@ -0,0 +1,48 @@
...
+cmake --build _cmake_build --target install
+cd ../
+rm -r librdkafka
+
+# Install rapidjson
+echo " ------> Install rapidjson..."
+cd /tmp
+git clone https://github.com/Tencent/rapidjson
+cd rapidjson/
+git checkout a95e013b97ca6523f32da23f5095fcc9dd6067e5
```
adamlm, line 32: Maybe add a short justification on why we are using a specific commit hash and not a release version/branch.

adamlm (changes requested):
General comment/question: Why do we have multiple `Dockerfile`s for `streets_service_base` related images? They should be merged into a single, mutli-stage `Dockerfile` unless there is a compelling reason not to.

paulbourelly999:
> General comment/question: Why do we have multiple `Dockerfile`s for `streets_service_base` related images? They should be merged into a single, mutli-stage `Dockerfile` unless there is a compelling reason not to.

My thought process is there are two different base images created here. Most streets services will only need `streets_service_base` but some services need an understanding of the geometry of the intersection. This is where the lanelet aware base image comes into play. Just to illustrate this I can outline what current carma-streets services should have which base image.

Scheduling Service : FROM streets_service_base
Signal Optimization Service: FROM streets_service_base
TSC Service: FROM streets_service_base
Intersection Model: FROM streets_service_base_lanelet_aware
Message Service: FROM streets_service_base_lanelet_aware
Sensor Data Sharing Service: FROM streets_service_base_lanelet_aware

I could do this with a single docker file and a multi-staged build, but it may be more readable to have 2 separate docker files. The only other thing I am concerned about is if I can get docker hub automated builds to push intermediate stages as images as well.  If I can get that to work I will combine the two into a single docker file

build_scripts/install_streets_service_base_dependencies.sh
paulbourelly999: OK there have not been any release even though there has been development since august 20116 but Ill add a comment here to say that

build_scripts/install_streets_service_base_dependencies.sh
paulbourelly999: This is from the cmake build instructions from the librdkafka (https://github.com/confluentinc/librdkafka/tree/master/packaging/cmake). I can leave a comment here with this link

build_scripts/install_streets_service_base_dependencies.sh
paulbourelly999: See comment below

build_scripts/install_streets_service_base_dependencies.sh
paulbourelly999: This means we need to install the lsb_release command. Not sure how much of a difference this makes personally although I am not opinionated either way.

build_scripts/install_lanelet2_dependencies.sh
paulbourelly999: I do not want to mess with the lanelet2 install we have working here. Eventually we probably want to move this entire process to a newer ubuntu and therefore ros distro. We can make any build optimization then in my opinion. What would the benefit of using colcon be here?

streets_utils/streets_service_base/CMakeLists.txt
paulbourelly999: Good catch.

build_scripts/install_streets_service_base_dependencies.sh
paulbourelly999: My one reasoning for not doing this is to avoid 2 separate calls to apt update and apt install since installing the lsb_release has to happen before adding the stol-apt-repository. After that we have to call update and install again to actually install dependencies.

build_scripts/install_lanelet2_dependencies.sh
adamlm: Uniformity across all our ROS related build commands, and it would make it easier to migrate away from ROS 1 in the future. Your argument makes sense, so we can leave it as is.

build_scripts/install_streets_service_base_dependencies.sh
adamlm: I thought the `lsb-release` package installed `/etc/lsb-release`, but I am wrong. I agree we should avoid extra `apt install` commands, so let's keep this.

adamlm:
> I could do this with a single docker file and a multi-staged build, but it may be more readable to have 2 separate docker files. The only other thing I am concerned about is if I can get docker hub automated builds to push intermediate stages as images as well. If I can get that to work I will combine the two into a single docker file

Are you recommend we push that to a future issue, or do you want me to wait on the review until you get that working?

paulbourelly999:
> mmend we push that to a future issue, or do you want me to wait on the review until you get that working?

I looked into it, dockerhub automated builds and it does not offer any way that I have found to specific `--target` in a multi-stage build, which means that a multi-stage dockerfile in this case would only be able to push the final stage image. Alternatively we could use github actions to build the base images instead, which I am ok with. I am still in favor of keeping these as separate Dockerfiles since it makes the architecture easier to handle. Multi-stage build does not allow you to specify the tag of a build stage so I can no longer make streets_service_base_lanelet_aware use streets_service_base:bionic. I want to be able to push streets_service_base for bionic, focal and jammy and then build off those. I think for that reason it makes more sense to keep these separate.

build_scripts/install_lanelet2_dependencies.sh
```diff
@@ -0,0 +1,70 @@
+#!/bin/bash
+# Script assumes base image of carma-builds-64x
```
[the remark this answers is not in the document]
paulbourelly999: Sure

adamlm (changes requested):

.github/workflows/build_streets_base_images.yml
```diff
@@ -0,0 +1,40 @@
...
+    docker-build:
+        strategy:
+            matrix:
+                include:
+                 - ubuntu-codename: bionic
+                   build_lanelet_aware: true
+                 - ubuntu-codename: focal
+                   build_lanelet_aware: false
+                 - ubuntu-codename: jammy
+                   build_lanelet_aware: false
```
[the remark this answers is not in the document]
paulbourelly999: We can't yet. Updating ubuntu distro, forces update to ros version and we do not have scripts yet working for that. Made this so that when we eventually do solve this we can easily fix it here.

.github/workflows/build_streets_base_images.yml
```diff
@@ -0,0 +1,40 @@
...
+                username: ${{ secrets.DOCKERHUB_USERNAME }}
+                password: ${{ secrets.DOCKERHUB_TOKEN }}
+          - name: Docker Build Streets Service Base
+            uses: docker/build-push-action@v3
+            with:
+                push: true
+                build-args: |
+                    UBUNTU_CODENAME=${{ matrix.ubuntu-codename }}
+                tags: usdotfhwastoldev/streets_service_base:${{ matrix.ubuntu-codename }}
+                file: ./streets_service_base/Dockerfile
```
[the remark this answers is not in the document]
paulbourelly999: Ok I can move the directory to streets_utils if we want. Alternatively we can create a Docker folder and I can move all my Dockerfiles and eventually all Dockerfiles under this folder. My only hold up about moving this to streets_utils is that it kind of broadens the scope of streets_utils to include more than just utility libraries. What do you think?

build_scripts/install_lanelet2_dependencies.sh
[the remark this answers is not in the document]
paulbourelly999: I added a top level readme with references

.github/workflows/build_streets_base_images.yml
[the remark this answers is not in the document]
adamlm: Makes sense. Let's add a comment capturing that.

.github/workflows/build_streets_base_images.yml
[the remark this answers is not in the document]
adamlm: After an internal discussion, we are going to keep the directory structure as it is. To help provide context for newcomers, we will add a directory structure information section within the `README.md` file in a future issue.

adamlm (approved):

dan-du-car (approved):

PR #378, opened 2023-12-14
Fix failing tests

paulbourelly999:
# PR Details
## Description
After some code quality investigations done by Sai, we found that some of the unit tests in CARMA Streets are failing silently. This PR prevents that silently failing unit tests by setting the `set -e` option. It also fixes the unit tests that were failing.

Most were failing due to missing initialization of singleton classes 'streets_configuration` or `streets_clock_singleton`. One library was creating duplicate failures of unit tests by test cases in the actual library. This was a cmake bug with the `streets_utils::streets_signal_optimization` library being included in the `signal_optimization_service`.  These unit tests would fail once during library unit tests and once during service unit test. This CMake bug has also been addressed.

## Related Jira Key

CDAR-300
## Motivation and Context
Fix CI/CD for unit test coverage

## How Has This Been Tested?
CI/CD and locally running of the unit tests

## Types of changes

## Checklist:

### DIFF
--- a/streets_utils/streets_vehicle_scheduler/test/test_signalized_scenario.cpp
+++ b/streets_utils/streets_vehicle_scheduler/test/test_signalized_scenario.cpp
@@ -31,6 +31,7 @@ namespace {
          */
         void SetUp() override {
             
+            streets_service::streets_clock_singleton::create(false);
             schedule = std::make_shared<signalized_intersection_schedule>();
             // the schedule timestamp is set to 10000 tenths of seconds from the current hour.
             uint64_t hour_tenth_secs = 10000;
--- a/coverage.sh
+++ b/coverage.sh
@@ -15,7 +15,7 @@
 
 # script to run tests, generate test-coverage, and store coverage reports in a place
 # easily accessible to sonar. Test names should follow convention run<pluginName>Tests
-
+set -e
 # For lanelet aware streets services like message_services and intersection_model
 source /opt/ros/melodic/setup.bash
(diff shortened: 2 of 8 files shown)


streets_utils/streets_vehicle_scheduler/test/test_signalized_scenario.cpp
```diff
@@ -31,6 +31,7 @@ namespace {
          */
         void SetUp() override {
             
+            streets_service::streets_clock_singleton::create(false);
```
kjrush, line 34: Unrelated to this PR, but I wonder if for the future (or maybe just as a lesson learned for future attempts at similar) thinking of other ways to define this function might be good.

Just seeing this in isolation doesn't tell me what "true" or "false" mean, is it referencing "simulation mode = false" or "real time mode = false". IIRC here it means "simulation mode = false", but a different interface design could have used an enum `..._clock_singleton::create(Clock::SIMULATION_TIME)` vs `..._clock_singleton::create(Clock::REAL_TIME)`. Or maybe different create functions? `..._clock_singleton::create_simulation_clock()` `..._clock_singleton::create_realtime_clock()`?

Minor and unrelated to this PR, but my $0.02

kjrush (approved):

streets_utils/streets_vehicle_scheduler/test/test_signalized_scenario.cpp
paulbourelly999: Noted.

PR #384, opened 2023-12-19
Fix Sensor Data Sharing Serivice Docker Image

paulbourelly999:
# PR Details
## Description
Updates to create working docker image for sensor data sharing service. Added :
- ldconfig call to build.sh script to allow linker to find proj4 library
- added source /opt/carma_lanelet2/setup.bash call to entry point to allow sensor data sharing service to find lanelet2 dependencies

## Related Jira Key
CDAR-300

## How Has This Been Tested?
Local integration testing

## Types of changes

## Checklist:

### DIFF
--- a/sensor_data_sharing_service/Dockerfile
+++ b/sensor_data_sharing_service/Dockerfile
@@ -3,6 +3,8 @@ COPY ./sensor_data_sharing_service/ /home/carma-streets/sensor_data_sharing_serv
 COPY ./streets_utils/ /home/carma-streets/streets_utils
 COPY ./kafka_clients/ /home/carma-streets/kafka_clients
 RUN /home/carma-streets/sensor_data_sharing_service/build.sh
+RUN chmod u+x /opt/carma_lanelet2/setup.bash
+
 LABEL org.label-schema.schema-version="1.0"
 LABEL org.label-schema.name="sensor_data_sharing_service"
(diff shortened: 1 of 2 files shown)


sensor_data_sharing_service/Dockerfile
```diff
@@ -3,6 +3,8 @@ COPY ./sensor_data_sharing_service/ /home/carma-streets/sensor_data_sharing_serv
 COPY ./streets_utils/ /home/carma-streets/streets_utils
 COPY ./kafka_clients/ /home/carma-streets/kafka_clients
 RUN /home/carma-streets/sensor_data_sharing_service/build.sh
+RUN chmod u+x /opt/carma_lanelet2/setup.bash
```
kjrush, line 6: Is this file generated as part of the build process? Or is there a reason it doesn't have these flags in the repo naturally?

paulbourelly999: Yes, the catkin build of lanelet2 creates this file. I have no idea why but for some reason I get permission denied when trying to source it from the dockerfile. It is also sources in the previous step in the build.sh script and that works fine without any chmod changes. It does not persist to the next docker step since I think docker runs separate shells for each command or something. Open to alternatives or if I missed something.

kjrush (approved):

Issue #391, opened 2024-01-24
Streets Service logs to a non configurable relative path


paulbourelly999:
### Summary

Currently the streets_configuration singleton creates the service logger on initialization for streets_services. The logger uses a hard coded relative path to generate log files. This issue is related to https://github.com/usdot-fhwa-stol/carma-streets/issues/342 and would be fixed by addressing #342 and updating all streets services to extend streets_service_base class

### Version

4.3.0 (Current)

### Expected Behavior

All log files are generated in a configurable common directory.

### Actual Behavior

Service log files are generated under a hardcoded relative path

### Steps to Reproduce the Actual Behavior

NA

PR #392, opened 2024-01-24
Adding working directory to sensor_data_sharing_service

### LINKED ISSUE #342: Move Service Logger configuration to streets_service_base
### Summary

Move the configuration of a streets service log file to the streets_service_base class since all services need file logger and streets_service_base has helper methods to create additional loggers which can be used for the service file logger.

### Reasoning for new functionality

Improve extensibility and framework

paulbourelly999:
# PR Details
## Description

Due to #342 and #391 **sensor_data_sharing_service** would generate logs files in the incorrect directory and these would not be exposed via the volume. By adding a working directory, the hard coded relative path for generating the log file points to the correct log directory. 

**NOTE: This PR does not fix #342 or #391. It only address the problem it creates for the sensor_data_sharing_service which is that logs are not published to the logs volume.**

## Related GitHub Issue
#342 and #391

## Related Jira Key
[CDAR-743
](https://usdot-carma.atlassian.net/browse/CDAR-743)

## Motivation and Context
Fix logging for sensor data sharing service

## How Has This Been Tested?
local integration testing 

## Types of changes

## Checklist:

### DIFF
--- a/sensor_data_sharing_service/Dockerfile
+++ b/sensor_data_sharing_service/Dockerfile
@@ -4,6 +4,9 @@ COPY ./streets_utils/ /home/carma-streets/streets_utils
 COPY ./kafka_clients/ /home/carma-streets/kafka_clients
 RUN /home/carma-streets/sensor_data_sharing_service/build.sh
 RUN chmod u+x /opt/carma_lanelet2/setup.bash
+# Required due to log file hardcoded relative path "../logs" 
+# (see https://github.com/usdot-fhwa-stol/carma-streets/issues/391 & 
+# https://github.com/usdot-fhwa-stol/carma-streets/issues/342)
 WORKDIR /home/carma-streets/sensor_data_sharing_service/build
 
 LABEL org.label-schema.schema-version="1.0"
--- a/sensor_data_sharing_service/Dockerfile
+++ b/sensor_data_sharing_service/Dockerfile
@@ -4,6 +4,7 @@ COPY ./streets_utils/ /home/carma-streets/streets_utils
 COPY ./kafka_clients/ /home/carma-streets/kafka_clients
 RUN /home/carma-streets/sensor_data_sharing_service/build.sh
 RUN chmod u+x /opt/carma_lanelet2/setup.bash
+WORKDIR /home/carma-streets/sensor_data_sharing_service/build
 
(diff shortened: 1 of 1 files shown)


sensor_data_sharing_service/Dockerfile
```diff
@@ -4,6 +4,7 @@ COPY ./streets_utils/ /home/carma-streets/streets_utils
 COPY ./kafka_clients/ /home/carma-streets/kafka_clients
 RUN /home/carma-streets/sensor_data_sharing_service/build.sh
 RUN chmod u+x /opt/carma_lanelet2/setup.bash
+WORKDIR /home/carma-streets/sensor_data_sharing_service/build
```
dan-du-car, line 7: Is it possible to create volume map the logs to a centralized location like: /opt/streets/logs, then each service has their own folder of log under this logs folder. eg: /opt/streets/logs/sensor_data_sharing_service.
I am not sure how this WORKDIR fix the relative path of the log file. It seems the volume to mapped to absolute path inside the container:  - ./sensor_data_sharing_service/logs/:/home/carma-streets/sensor_data_sharing_service/logs/

paulbourelly999: This issue is that streets_configuration singleton creates the log file at a hardcoded relative that '../logs' ( #391 #342). This should be replaced by the streets_service_base log creation that allows you to generate a log file at a configurable location based on the environment variable **LOGS_DIR**. I can not fix this yet because not all services extend the streets_service_base class so removing log creation from streets_configuration singleton will make old services like intersection_model and message_service not generate logs files.

dan-du-car: Can you add a comment before line 7 to note about this relative path required by the singleton?

MishkaMN (commented):
I don't have much comments except what Dan suggested

sensor_data_sharing_service/Dockerfile
paulbourelly999: Addressed

paulbourelly999:
> I don't have much comments except what Dan suggested

addressed

MishkaMN (approved):

dan-du-car (approved):

PR #400, opened 2024-02-28
CDA-819: Add timestamp information to detection data in SDSM

paulbourelly999:
# PR Details
## Description
This PR adds measurement time offset to detection data in the sdsm. This allows for each detection timestamp information to be passed to consumers of the SDSM. Without this information the interpreted error of detection to SDSM was large

## Related Jira Key
CDAR-819

## Motivation and Context
Fix SDSM bug

## How Has This Been Tested?
Integration tested in CDASim deployment

## Types of changes

## Checklist:

### DIFF
--- a/sensor_data_sharing_service/src/detected_object_to_sdsm_converter.cpp
+++ b/sensor_data_sharing_service/src/detected_object_to_sdsm_converter.cpp
@@ -68,7 +68,7 @@ namespace sensor_data_sharing_service{
         }
         // Used to convey an offset in time relative to the sDSMTimeStamp associated with the reference position. Negative values indicate 
         // the provided detected object characteristics refer to a point in time after the sDSMTimeStamp
-        detected_object._detected_object_common_data._time_measurement_offset = static_cast<int64_t>(sdsm_message_timestamp) - static_cast<int64_t> (msg._epoch_time_ms);
+        detected_object._detected_object_common_data._time_measurement_offset = static_cast<int64_t>(sdsm_message_timestamp) - static_cast<int64_t> (msg._timestamp);
         detected_object._detected_object_common_data._classification_confidence = static_cast<unsigned int>(msg._confidence*100);
         // TODO: Change Detected Object ID to int
         detected_object._detected_object_common_data._object_id = msg._object_id;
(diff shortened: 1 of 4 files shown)


sensor_data_sharing_service/src/detected_object_to_sdsm_converter.cpp
```diff
@@ -66,6 +66,10 @@ namespace sensor_data_sharing_service{
             detected_object._detected_object_optional_data = optional_data;
 
         }
+        // Used to convey an offset in time relative to the sDSMTimeStamp associated with the reference position. Negative values indicate 
+        // the provided detected object characteristics refer to a point in time after the sDSMTimeStamp
+        detected_object._detected_object_common_data._time_measurement_offset = 
+            static_cast<int> (static_cast<int64_t>(sdsm_message_timestamp) - static_cast<int64_t> (msg._timestamp));
```
adamlm, line 72: We don't need to do so many casts. This should suffice:

```cpp
detected_object._detected_object_common_data._time_measurement_offset = 
  static_cast<int>(sdsm_message_timestamp - msg._timestamp);
```

adamlm (changes requested):

sensor_data_sharing_service/src/detected_object_to_sdsm_converter.cpp
```diff
@@ -33,7 +33,7 @@ namespace sensor_data_sharing_service{
         return sdsm_timestamp;
     }
 
-    streets_utils::messages::sdsm::detected_object_data to_detected_object_data(const streets_utils::messages::detected_objects_msg::detected_objects_msg &msg) {
+    streets_utils::messages::sdsm::detected_object_data to_detected_object_data(const streets_utils::messages::detected_objects_msg::detected_objects_msg &msg, const uint64_t sdsm_message_timestamp ) {
```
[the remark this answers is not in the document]
paulbourelly999: Good catch

sensor_data_sharing_service/src/detected_object_to_sdsm_converter.cpp
paulbourelly999: No it won't, both sdsm timestamp and detection timestamp are unsigned ints so the result would also be an unsigned int. This is an issue because the result needs to be a signed int. The code you have would give incorrect results for values where `msg._timestamp` is larger than 'sdsm_message_timestamp`

adamlm: Here's a Compiler Explorer example showing otherwise: https://godbolt.org/z/rW57Pn78T. Am I missing something?

adamlm (approved):

PR #417, opened 2024-04-10
Added workflows for release and master builds for base imags

paulbourelly999:
# PR Details
## Description
Adding CI processes to correct tag base images

## How Has This Been Tested?
CI/CD (https://github.com/usdot-fhwa-stol/carma-streets/actions/runs/8632866000/job/23664804594) Just look at the tags. Build failed because carma-system-4.5.0 does not exist yet for streets_base images.

## Types of changes

## Checklist:

- [ ] I have read the [**CONTRIBUTING**](https://github.com/usdot-fhwa-stol/carma-platform/blob/develop/Contributing.md) document.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -1,7 +1,7 @@
-name: develop_master
+name: master
 on:
   push:
-    branches: [develop, master]
+    branches: [master]
 jobs:
   docker-build:
     strategy:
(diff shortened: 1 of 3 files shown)


.github/workflows/master.yml
```diff
@@ -0,0 +1,134 @@
+name: master
+on:
+  push:
+    branches: [master]
```
MishkaMN, line 4: release and master files don't have difference in terms of content. Can we just add the release tag trigger here?

MishkaMN (changes requested):

.github/workflows/master.yml
paulbourelly999: No because its hard to tell in github action which event triggers a workflow. For tags the {{ git.refname}} is tagname and for branch pushes its branch name

MishkaMN (approved):

PR #421, opened 2024-07-31
CS-57: Fix threading for TSC and fix CI for bionic nodejs version incompatibility

paulbourelly999:
# PR Details
## Description
This PR addresses to regressions in the **TSC Service**.
The first of which is related to launching the threads necessary for consuming a desired phase plan and making the appropriate NTCIP 1202 calls to a traffic signal controller (TSC). We use `std::thread` s for this and call the `join()` method on each instead of `detach()` to launch all our threads ( See [sonar scanner rule](https://sonarcloud.io/organizations/usdot-fhwa-ops/rules?open=cpp%3AS5962&rule_key=cpp%3AS5962&tab=why) and [CPP Core Guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/e49158a/CppCoreGuidelines.md#cp26-dont-detach-a-thread) to see why we use join instead of detach). The [std::thread::join()](https://en.cppreference.com/w/cpp/thread/thread/join) call blocks until the referenced thread completes execution. Due to this , all `join()` calls need to be at the end of their scope. Inserting a `join()` call before some other logic will prevent that logic from executing until the joined thread completes execution, which for many of these threads will only happen when the service shuts down.
As part of some MMITSS development this join was moved into the incorrect location causing the `control_tsc_phases` thread never to be created. Moving all the `join()` calls to the end of the start method allows all threads to run and no logic to be blocked by thread execution completion.

The second more benign regression was that we changed the **SNMP COMMAND LOGGER** to take it's name from a constant. In doing so, the name changes and calls to retrieve this logger did not use the correct name.  This was also fixed in this PR 

The final change is exclusively to fix issues with the CI process related to bionic (ubuntu distribution) incompatibility with many github actions due to its incompatibility with newer versions of nodejs. This environment variable allow github runners to use deprecated versions of nodesJS. Eventually we plan to remove bionic from CARMA Streets.

## Related Jira Key
[CS-57
](https://usdot-carma.atlassian.net/browse/CS-58)

## Motivation and Context
Hot fix for carma-system-4.5.0 for TSC Service. Fixing functionality to control phases given a desired phase plan.

## How Has This Been Tested?
Tested in CDASim deployment

## Types of changes

## Checklist:


.github/workflows/feature_branch.yml
```diff
@@ -2,6 +2,11 @@ name: feature_branch
 on:
   pull_request:
     types: [opened, synchronize, reopened]
+# make GHA actions use node16 which still works with bionic
+# See https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/
+# Unclear how long this will work though. Remove after rebase carma-streets to newer Ubuntu release
+env:
+  ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: true
```
dan-du-car, line 9: Found this link: https://github.com/nodesource/distributions/issues/1392 It seems that installing glibc-2.28 can be a potential fix to the Node 20 compatibility issue.

paulbourelly999: It looks like install glibc-2.28 is either very difficult or not possible on bionic. https://community.nodebb.org/topic/16918/nodejs-18-glibc-2-28-in-ubuntu-18-04-cannot-be-installed. This is the core of the issue. Bionic packages are old and often incompatible with newer tools. Either way ensuring that we use node 16 for our CI is way lower risk that making any involved attempts to update the version of glibc we install in our bionic images.

dan-du-car (dismissed):

dan-du-car (approved):

SaikrishnaBairamoni (approved):
PR looks good to me. As long as we have comments and are tracking them, we can update it accordingly.

PR #426, opened 2024-08-27
sensor_data_sharing_service: Cannot parse confidence

dan-du-car:
# PR Details
## Description
Update parse_double_member function in the json_utils library to parse any numeric data. Previously it would fail to parse integers as doubles.

## Related GitHub Issue
https://github.com/usdot-fhwa-stol/carma-streets/issues/425

## Related Jira Key
[FCP-30](https://usdot-carma.atlassian.net/browse/FCP-30)

## Motivation and Context
FCP

## How Has This Been Tested?
Unit testing

## Types of changes

- [x] Defect fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that cause existing functionality to change)

## Checklist:

- [ ] I have added any new packages to the sonar-scanner.properties file
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the [**CONTRIBUTING**](https://github.com/usdot-fhwa-stol/carma-platform/blob/develop/Contributing.md) document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

[FCP-30]: https://usdot-carma.atlassian.net/browse/FCP-30?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #427, opened 2024-08-27
Update Sensor Configuration File

paulbourelly999:
# PR Details
## Description

The changes recommended in this PR are to modify the Sensor Configuration file to be able to provide either a cartesian offset or a WGS84 location based on a type enumeration. The new format of the sensor configuration file is described below
> [!NOTE]
> Sensor reference or "ref" can either be defined as a "CARTESIAN" offset from a provided lanelet2 osm map (first example) or a "WGS84" location (second example). The "CARTESIAN" offset is only recommended for simulation testing as it adds a dependency on a lanelet2 map. For real-world deployments it is recommended that the "WGS84" location is provided.

**CARTESIAN(SIMULATION)**:

```json
[
  {
    "sensorId": "sensor_1",
    "type": "SemanticLidar",
    "ref": {
      "type": "CARTESIAN",
      "location": {
        "x": 1.0, /* in meters */
        "y": 2.0, /* in meters */
        "z": -3.2 /* in meters */
      },
       "orientation": { 
        "yaw": 0.0,
        "pitch": 0.0,
        "roll": 0.0
        }
    }
  } 
]
```
The orientation information provided here is not used by the **Sensor Data Sharing Service**, but it used by other parts for the **CARMA Streets** system (V2X-Hub/CDASimAdapterPlugin) in simulation to provide orientation information for spawning a requested sensor.

or

**WGS84(REAL-WORLD)**:

```json
[
  {
    "sensorId": "sensor_1",
    "type": "SemanticLidar",
    "ref": {
      "type": "WGS84",
      "location": {
        "latitude": 38.9549716548523, /* in degrees */
        "longitude": -77.14935313519123, /* in degrees */
        "altitude": 5.5 /* in meters */
      }
    }
  } 
]
```
I opted to use **WGS84** coordinates since we do not need the orientation or offsets included in a geo refence projection string in real world deployment. The detection data has a assumed ENU coordinate frame so all we need is the reference sensor location from which this detection data is measured.

> [!WARNING]
> The changes to the Sensor Configuration file format do break some of the existing CARMA-Streets/Simulation integration. An additional PR will be introduced into CDASim to account for the format difference in the Sensor Configuration information

## Related Jira Key
[FCP-37
](https://usdot-carma.atlassian.net/browse/FCP-37)

## Motivation and Context
Currently the Sensor Data Sharing Services reads geo-reference information from lanelet2 osm map and then reads cartesian position for a sensor from the sensor configuration file to generate an SDSM WSG84 reference location.
**Example Geo reference string included in lanelet2 osm map**
```
+proj=tmerc +lat_0=0 +lon_0=0 +k=1 +x_0=0 +y_0=0 +datum=WGS84 +units=m +geoidgrids=egm96_15.gtx +vunits=m +no_defs 
```
**Example Sensor Config**
```
[
  {
    "sensorId": "sensor_1",
    "type": "SemanticLidar",
    "location": {
      "x": 1.0,
      "y": 2.0,
      "z": -3.2
     },
     "orientation": {
       "yaw": 0.0,
       "pitch": 0.0,
       "roll": 0.0
      }
  } 
]
```

**Geo reference + sensor config cartesian offset = sdsm reference location**

This is a problem because it unnecessarily introduces a dependency on an lanlet2 OSM map for SDSM functionality. It is also not practical to attempt to configure a real-world sensor based on a cartesian offset in a lanelet2 map. This implementation was created because conversely it is not practical to generate a sensor in simulation with WGS 84 coordinate location since many simulation environments do not have real world counter parts. 
As a result we offer two different ways to configure sensor's in simulation and for real world outlined above. This removes unnecessary dependency on a lanelet2 osm map for real world sensors, while still allowing easy configuration of sensor in simulation. 

## How Has This Been Tested?
Current testing is limited to unit testing. It is recommended that this change set is integration tested #426 PR is closed.

## Types of changes

- [x] Breaking change (fix or feature that cause existing functionality to change)

## Checklist:


docker-compose.yml
```diff
@@ -1,4 +1,3 @@
-version: '3.7'
```
dan-du-car, line 1: Why do we remove the docker-compose yaml version?

sensor_data_sharing_service/README.md
```diff
@@ -1,18 +1,82 @@
...
+```json
+[
+  {
+    "sensorId": "sensor_1",
+    "type": "SemanticLidar",
+    "ref": {
+      "type": "CARTESIAN",
+      "location": {
+        "x": 1.0,
+        "y": 2.0,
```
dan-du-car: This JSON is the sensor location in cartesian coordinate system?

sensor_data_sharing_service/README.md
```diff
@@ -1,18 +1,82 @@
...
+  } 
+]
+```
+
+or
+
+```json
+[
+  {
+    "sensorId": "sensor_1",
```
dan-du-car, line 52: This JSON is the physical MUST sensor location in WGS84?

dan-du-car: Can we Provide title and description for each of this sensor configuration before providing the JSON example?

sensor_data_sharing_service/README.md
```diff
@@ -1,18 +1,82 @@
...
+
+**LANELET2_MAP**: Only required for "CARTESIAN" sensor references. This is a file path to a lanelet2 osm map for which the cartesian offset is relavent.
+
+### Service Configuration
+
+**sdsm_producer_topic**: Kafka topic to which the sensor data sharing service will produce SDSMs.
+**detection_consumer_topic**: Kafka topic from which the sensor data sharing service will consume Detected Objects.
+**sensor_configuration_file_path**: Path to sensor configuration information that includes sensor location and orientation.
+**bootstrap_server**: Kafka Broker Server Address.
+**sensor_id**: Unique id of sensor for which to publish SDSMS
```
dan-du-car, line 74: Can we dynamically generate this sensor ID?

sensor_data_sharing_service/include/sensor_configuration_parser.hpp
```diff
@@ -37,11 +37,26 @@
 #include <streets_utils/json_utils_lib/json_utils.hpp>
 
 namespace sensor_data_sharing_service{
-    /**
+
+    enum class ReferenceType {
```
dan-du-car: Can we provide more description about what this reference type is, how and when it is used?

sensor_data_sharing_service/include/sensor_configuration_parser.hpp
```diff
@@ -37,11 +37,26 @@
...
+        GEO_REF = 0,
+        CARTESIAN = 1,
+    };
+    struct SensorReference {
+        std::string sensorId = "";
+        std::string map_projection = "";
+        lanelet::BasicPoint3d location;
```
dan-du-car: Can we make a note here that location is a cartesian coordinate for readers who are not aware of lanelet2 implementation?

sensor_data_sharing_service/manifest.json
```diff
@@ -31,12 +31,6 @@
             "value": "sensor_1",
             "description": "Unique id of sensor for which to publish SDSMS",
             "type": "STRING"
-        },
```
dan-du-car, line 34: This is removed because we are getting the reference string from the detected object sent by V2xHub?

sensor_data_sharing_service/include/sensor_configuration_parser.hpp
```diff
@@ -37,11 +37,26 @@
...
+        ReferenceType reference_type;
+    };
+
+
+     /**
      * @brief Method to parse sensor configuration json file.
      * @param filepath absolute or relative file path.
      * @param sensor_id ID of sensor to get location from.
      * @return cartesian location of sensor described in json configuration file.
```
dan-du-car: Please add a link to there JSON configuration file for easier navigation to the file.

sensor_data_sharing_service/include/sensor_configuration_parser.hpp
```diff
@@ -37,11 +37,26 @@
...
      */
-    lanelet::BasicPoint3d parse_sensor_location( const std::string &filepath, const std::string &sensor_id );
+    SensorReference parse_sensor_ref( const std::string &filepath, const std::string &sensor_id );
```
dan-du-car, line 76: Is this only used in simulation mode? Can we make a note in the @brief?

sensor_data_sharing_service/README.md
```diff
@@ -1,18 +1,82 @@
...
+> [!NOTE]
+> Sensor reference or "ref" can either be defined as a "CARTESIAN" offset from a provided lanelet2 osm map (first example) or a "GEOREF" projection string (second example). The "CARTESIAN" offset is only recommended for simulation testing as it adds a dependency on a lanelet2 map. For real-world deployments it is recommended that the "GEOREF" reference is provided.
+ 
+```json
+[
+  {
+    "sensorId": "sensor_1",
+    "type": "SemanticLidar",
+    "ref": {
+      "type": "CARTESIAN",
```
dan-du-car, line 29: Can we make a note what available types? Is this case sensitive?

docker-compose.yml
paulbourelly999: https://docs.docker.com/reference/compose-file/version-and-name/#version-top-level-element-obsolete

sensor_data_sharing_service/src/sensor_configuration_parser.cpp
```diff
@@ -15,31 +15,42 @@
...
-                auto location = streets_utils::json_utils::parse_object_member("location",  itr->GetObject(), true ).value();
-                sensor_location = lanelet::BasicPoint3d{
-                    streets_utils::json_utils::parse_double_member("x",  location, true ).value(),
-                    streets_utils::json_utils::parse_double_member("y",  location, true ).value(),
-                    streets_utils::json_utils::parse_double_member("z",  location, true ).value()
-                    };
+                found = true; 
+                auto ref = streets_utils::json_utils::parse_object_member("ref",  itr->GetObject(), true ).value();
+                auto ref_type = streets_utils::json_utils::parse_string_member("type", ref, true).value();
+                if ( ref_type == "CARTESIAN") {
```
dan-du-car, line 35: Can we provide a log to tell which reference type is currently used?

sensor_data_sharing_service/README.md
paulbourelly999: In simulation we spawn a sensor base on cartesian offset from a lanelet2 map. This allows us to spawn a sensor easily for locations that may not exist the in real world like location in CARLA Town04. For these sensors it is substantially easier to get osm map cartesian offsets rather than having to convert those to a WGS 84 absolute location.

sensor_data_sharing_service/README.md
paulbourelly999: Yes, we include an entire projection string to ensure we capture orientation but according to their diagram they report all detection offsets NED (Northx Easty Downz) which should alleviate need for a projection string 
See code book for reference diagram

sensor_data_sharing_service/src/sensor_configuration_parser.cpp
```diff
@@ -15,31 +15,42 @@
...
+                    sensor_ref.reference_type = ReferenceType::CARTESIAN;
+
+                    auto location = streets_utils::json_utils::parse_object_member("location",  ref, true ).value();
+                    sensor_ref.location = lanelet::BasicPoint3d{
```
dan-du-car: This implies that we use the map_projection string in the OSM map? Can we make a note here that we do not need to populate the map_projection here?

sensor_data_sharing_service/src/sensor_configuration_parser.cpp
```diff
@@ -15,31 +15,42 @@
...
+                        streets_utils::json_utils::parse_double_member("x",  location, true ).value(),
+                        streets_utils::json_utils::parse_double_member("y",  location, true ).value(),
+                        streets_utils::json_utils::parse_double_member("z",  location, true ).value()
+                        };
+                }
+                else { 
+                    sensor_ref.reference_type = ReferenceType::GEO_REF;
```
dan-du-car: Also add a note here the reference location (in cartesian system) is assuming the origin (0,0,0)?

sensor_data_sharing_service/src/sensor_data_sharing_service.cpp
```diff
@@ -39,21 +39,25 @@ namespace sensor_data_sharing_service {
...
-            return false;
-        }
+       
         // Read sensor configuration file and get WSG84 location/origin reference frame.
         const std::string sensor_config_file = streets_service::get_system_config("SENSOR_JSON_FILE_PATH", "/home/carma-streets/sensor_configurations/sensors.json");
         const std::string sensor_id = ss::streets_configuration::get_string_config("sensor_id");
-        const lanelet::BasicPoint3d pose = parse_sensor_location(sensor_config_file, sensor_id);
-        this->sdsm_reference_point =  this->map_projector->reverse(pose);
+        auto sensor_ref = parse_sensor_ref(sensor_config_file, sensor_id);
+        if ( sensor_ref.reference_type == ReferenceType::CARTESIAN ) {
```
dan-du-car: Can we add a comment here about populating the map_projector?

sensor_data_sharing_service/README.md
```diff
@@ -1,18 +1,82 @@
...
+
+The **Sensor Data Sharing Service** uses the following environment variables to retreive CARMA Streets system level configuration information.
+
+**SIMULATION_MODE** : Boolean flag indicating whether the service is being run in a simulation or real-world enviroment. In simulation mode, this service use an external signal from the kafka `time_sync` topic as time.
+
+**SENSOR_JSON_FILE_PATH**: Path to a sensor configuration file with the following format. Used for retrieving a reference location and projection for received cartesian detection data
+> [!NOTE]
+> Sensor reference or "ref" can either be defined as a "CARTESIAN" offset from a provided lanelet2 osm map (first example) or a "GEOREF" projection string (second example). The "CARTESIAN" offset is only recommended for simulation testing as it adds a dependency on a lanelet2 map. For real-world deployments it is recommended that the "GEOREF" reference is provided.
+ 
+```json
```
dan-du-car, line 23: Do we have a specification doc about this sensor configuration?

sensor_data_sharing_service/README.md
paulbourelly999: This is the sensor ID for which the service will read detections for. Why would we want this to be dynamic?

sensor_data_sharing_service/manifest.json
paulbourelly999: It was an unused configuration parameter.

sensor_data_sharing_service/src/sensor_configuration_parser.cpp
paulbourelly999: I don't understand what you mean. The map projection is populated from the lanelet2 osm map file. I think that is pretty clearly stated in the PR and in the README right.

sensor_data_sharing_service/README.md
dan-du-car: Would it be possible that users can misconfigure this with duplicated ID? If we could automatically make this unique, we offload some configuration consideration effort? Is there a reason we have to make this manual configuration and control what specific ID we use?

sensor_data_sharing_service/src/sensor_configuration_parser.cpp
dan-du-car: I understand it is in README. I guess I was asking about reiterating here to make it more accessible for readers.

dan-du-car (approved):

PR #456, opened 2026-08-06
Add support for proj conversions with detections

willjohnsonk:
# PR Details
## Description

The sensor_data_sharing_service now leverages a new proj string field in the plugin to allow for geolocation "anchors" that can be referenced by connected sensors. SDSMs produced should now use the anchor proj string reference and update the cartesian offsets appropriately so that a shared reference can be used across multiple sensors in the future. 

## Related Jira Key

HASS-618

## Motivation and Context

Improves geodetic data handling with proj and supports multiple sensor detections in a shared frame.

## How Has This Been Tested?

Tested with detections_offset_from_ref_converter_test.cpp. Can be run in the dev container with in /home/$USER/carma-streets/sensor_data_sharing_service with:

source /opt/carma_lanelet2/setup.bash
./build.sh
./sensor_data_sharing_service_test

## Types of changes

## Checklist:


sensor_data_sharing_service/src/detection_offset_from_ref_converter.cpp
```diff
@@ -0,0 +1,71 @@
...
+            ref_detection._position._y = 0.0;
+            return ref_detection;
+        }
+
+        // Convert proj strings to char arr
+        const char* detection_proj = msg._proj_string.c_str();
+        const char* reference_proj = reference_proj_string.c_str();
+
+        // Creation of proj transformation object between the two proj strings
+        PJ *detection_to_ref = proj_create_crs_to_crs(ctx, reference_proj, detection_proj, NULL);
```
MishkaMN: proj_create_crs_to_crs(ctx, source, target, area) takes source first. Here source = reference_proj, target = detection_proj, whcih means backwards from what the variable name detection_to_ref claims. Combined with PJ_FWD, this computes "where does the reference origin sit in the detection's local tangent plane," the reverse of the intended "where does the detection point sit in the reference frame.

sensor_data_sharing_service/test/detection_offset_from_ref_converter_test.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+        detection._position._x = pos_x;
+        detection._position._y = pos_y;
+        detection._position._z = pos_z;
+        detection._proj_string = "+proj=tmerc +lat_0=38.95509339558609 +lon_0=-77.14738889727099 +k=1 +x_0=0 +y_0=0 +datum=WGS84 +units=m +geoidgrids=egm96_15.gtx +vunits=m +no_defs";
+
+        auto updated_detection = detected_object_local_to_ref(detection, ref_string);
+        // Cartesian delta actuals:
+        // x = -0.12576974393503
+        // y = -0.79408763431725
```
MishkaMN, line 70: I would double check these numbers

sensor_data_sharing_service/src/detection_offset_from_ref_converter.cpp
```diff
@@ -0,0 +1,71 @@
...
+        // Transform cartesian points into the shared reference crs
+        PJ_COORD ref_transformed = proj_trans(detection_to_ref, PJ_FWD, ref_cartesian);
+        PJ_COORD detection_transformed = proj_trans(detection_to_ref, PJ_FWD, detection_cartesian);
+
+        // Calculate cartesian offset
+        // Check for zero vals to prevent equivalent transforms cancelling out with ref_cartesian
+        if(msg._position._x == 0.0 && msg._position._y == 0.0){
+            ref_detection._position._x = detection_transformed.xy.x;
+            ref_detection._position._y = detection_transformed.xy.y;
+        }
```
MishkaMN: even the 0.0 case should be the same as normal case, no need for special case

sensor_data_sharing_service/src/detection_offset_from_ref_converter.cpp
```diff
@@ -0,0 +1,71 @@
...
+        else{
+            ref_detection._position._x = detection_transformed.xy.x - ref_transformed.xy.x;
+            ref_detection._position._y = detection_transformed.xy.y - ref_transformed.xy.y;
```
MishkaMN: I think it should be just  ref_detection._position._x = detection_transformed.xy.x. 

Overall, there is only 1 transform happening and all other code is unnecessary:
```
PJ *detection_to_ref_local = proj_create_crs_to_crs(ctx, detection_proj, reference_local_proj, NULL);
PJ_COORD detection_original = proj_coord(msg._position._x, msg._position._y, 0.0, 0.0);
PJ_COORD transformed_to_local = proj_trans(detection_to_ref_local, PJ_FWD, detection_original);
detection_in_ref._position._x = transformed_to_local.xy.x;
detection_in_ref._position._y = transformed_to_local.xy.y;
```

MishkaMN (changes requested):

sensor_data_sharing_service/test/detection_offset_from_ref_converter_test.cpp
MishkaMN:
```
ref  → lat_0=38.95508624257387, lon_0=-77.14738744642074
det  → lat_0=38.95509339558609, lon_0=-77.14738889727099
```
and
```
dlon = -0.0000014509°  (detection west of reference, menaing -negative x)
dlat = +0.0000071530°  (detection north of reference, meaning positive y)
```
Detection is north of reference (higher latitude) and west of reference (more negative longitude).
And lat lon difference is approx directly similar to the standard ENU x,y cartesian differences at small distances.

That's (-0.1258, +0.7941):

sensor_data_sharing_service/src/sensor_data_sharing_service.cpp
```diff
@@ -38,28 +38,36 @@ namespace sensor_data_sharing_service {
...
+            SPDLOG_WARN("No latitude value found in reference proj string.");
         }
+        if (proj_string_pairs.count("lon_0")){
+            ref_proj_gps.lon = std::stod(proj_string_pairs["lon_0"]);
+        }
+        else{
+            SPDLOG_WARN("No longitude value found in reference proj string.");
+        }
+
+        const std::string lanlet2_map =  streets_service::get_system_config("LANELET2_MAP", "/home/carma-streets/MAP/Intersection.osm");
```
MishkaMN: typo: lanelet2_map

sensor_data_sharing_service/src/sensor_data_sharing_service.cpp
```diff
@@ -234,14 +243,15 @@ namespace sensor_data_sharing_service {
...
+        // Populate equipment type
         msg._equipment_type = sdsm::equipment_type::RSU;
-        // Polulate ref position
+        // Populate ref position
         msg._ref_positon = to_position_3d(this->sdsm_reference_point);
         std::shared_lock lock(detected_objects_lock);
         for (const auto &[object_id, object] : detected_objects){
             auto ned_object = detected_object_enu_to_ned(object);
-            auto detected_object_data = to_detected_object_data(ned_object,timestamp);
+            auto transformed_object_data = detected_object_local_to_ref(ned_object, ref_proj_string);
```
MishkaMN: I think we first apply the object frame change on enu before turning it into ned object

MishkaMN (changes requested):

sensor_data_sharing_service/include/detection_offset_from_ref_converter.hpp
```diff
@@ -0,0 +1,37 @@
...
+#include <sstream>
+#include <unordered_map>
+#include <optional>
+
+namespace sensor_data_sharing_service { 
+    /**
+     * @brief Parses a proj string into an unordered map for value retrieval
+     * @param msg 
+     * @return Unordered map of string value pairs
+     */
```
MishkaMN, line 33: /**
     * @brief Parses a whitespace-separated PROJ string (e.g. "+proj=utm +zone=10 +datum=WGS84")
     * into a map of parameter name to value. Each token has any leading '+' stripped; tokens
     * containing '=' are split into a key/value pair, while flag-only tokens (no '=') are
     * stored with an empty string value.
     * @param proj_str PROJ definition string to parse
     * @return Unordered map of PROJ parameter names to their values (empty string if the
     * parameter is a flag with no value)
     */

sensor_data_sharing_service/include/detection_offset_from_ref_converter.hpp
```diff
@@ -0,0 +1,37 @@
...
+    std::unordered_map<std::string, std::string> parse_proj_string(const std::string &proj_str);
+    /**
+     * @brief Updates a detected_object_msg to have the reference proj string and appropriate cartesian offset
+     * @param msg 
+     * @return Positionally transformed detected_object_msg
+     */
```
MishkaMN, line 45: /**
     * @brief Transforms a detected object's cartesian position from its local/detection PROJ
     * frame into the given reference PROJ frame, using the PROJ library to build and apply the
     * CRS-to-CRS transformation. Returns a copy of msg with the updated position and its
     * _proj_string set to reference_proj_string. If the PROJ context or transformation cannot
     * be created (e.g. invalid or incompatible proj strings), the returned object's position is
     * set to (0.0, 0.0) instead.
     * @param msg Detected object with a local position and its originating PROJ string
     * @param reference_proj_string PROJ string of the shared reference frame to transform into
     * @return Copy of msg repositioned into the reference frame (or zeroed out on transform failure)
     */

Just to be more descriptive

MishkaMN (changes requested):

sensor_data_sharing_service/src/sensor_data_sharing_service.cpp
willjohnsonk: I don't think it matters. In actuality what we should probably do is use the proj library to do this instead of having a function dedicated to it. Proj supports enu/ned fields and could be used to transform them, but I haven't looked into this much. @paulbourelly999

MishkaMN: talked offline the order was need to be changed. I am okay either way for proj's enu ned usage

MishkaMN (dismissed):

sensor_data_sharing_service/src/detection_offset_from_ref_converter.cpp
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        // Convert proj strings to char arr
+        const char* detection_proj = msg._proj_string.c_str();
+        const char* reference_proj = reference_proj_string.c_str();
+
+        // Creation of proj transformation object between the two proj strings
+        PJ *detection_to_ref = proj_create_crs_to_crs(ctx, detection_proj, reference_proj, NULL);
+        if(!detection_to_ref){
+            proj_context_destroy(ctx);
```
paulbourelly999: Does this need a call to proj_destroy(detection_to_ref) to avoid a leak

sensor_data_sharing_service/src/detection_offset_from_ref_converter.cpp
```diff
@@ -0,0 +1,82 @@
...
+            ref_detection._position._x = 0.0;
```
paulbourelly999: Instead of returning 0.0.0 can we just throw and exception. 0.0.0 is technical not an invalid location and throwing an exception would make it easier and more clear to handle in the sensor_data_sharing_service

sensor_data_sharing_service/src/sensor_data_sharing_service.cpp
paulbourelly999: I there is a proj function we can use for this cartesian transform that works for me. The challenge is we need to be able to indicate that the incoming data is enu. Technically it sound we should be able to do both in one fell swoop but I am not sure if there order matters or not

paulbourelly999 (changes requested):
Some small comments about error handling

sensor_data_sharing_service/manifest.json
```diff
@@ -37,6 +31,12 @@
             "value": true,
             "description": "Bool flag to indicate whether to record detection metrics for the sensor data sharing service.",
             "type": "BOOL"
+        },
+        {
+            "name": "reference_proj_string",
+            "value": "+proj=tmerc +lat_0=38.95197911150576 +lon_0=-77.14835128349988 +k=1 +x_0=0 +y_0=0 +datum=WGS84 +units=m +geoidgrids=egm96_15.gtx +vunits=m +no_defs",
+            "description": "Positional Proj string that describes the WGS84 reference location of the infrastructure.",
```
paulbourelly999: Lets make clear this is the reference position used in the published SDSM

sensor_data_sharing_service/src/sensor_data_sharing_service.cpp
```diff
@@ -234,14 +242,15 @@ namespace sensor_data_sharing_service {
...
+        // Populate equipment type
         msg._equipment_type = sdsm::equipment_type::RSU;
-        // Polulate ref position
+        // Populate ref position
         msg._ref_positon = to_position_3d(this->sdsm_reference_point);
         std::shared_lock lock(detected_objects_lock);
         for (const auto &[object_id, object] : detected_objects){
-            auto ned_object = detected_object_enu_to_ned(object);
-            auto detected_object_data = to_detected_object_data(ned_object,timestamp);
+            auto transformed_object_data = detected_object_local_to_ref(object, ref_proj_string);
```
paulbourelly999: There is currently no error handling here. Lets through an exception from this method on failure, catch it, log the error and skip the detection

sensor_data_sharing_service/src/detection_offset_from_ref_converter.cpp
willjohnsonk: It is destroyed in line 76 assuming it is created successfully. This code should only trigger if it isn't created so no memory should be allocated to be cleared.

sensor_data_sharing_service/src/sensor_data_sharing_service.cpp
willjohnsonk: I was looking into this and it definitely has the axis metadeta for the string. Not sure if it processes it as part of the transform though, future work potentially. Reference from proj:

> Axis orientation[](https://proj.org/en/stable/usage/projections.html#axis-orientation)
> Starting in PROJ 4.8.0, the +axis argument can be used to control the axis orientation of the coordinate system. The default orientation is "easting, northing, up" but directions can be flipped, or axes flipped using combinations of the axes in the +axis switch. The values are:
> 
> "e" - Easting
> 
> "w" - Westing
> 
> "n" - Northing
> 
> "s" - Southing
> 
> "u" - Up
> 
> "d" - Down
> 
> They can be combined in +axis in forms like:
> 
> +axis=enu - the default easting, northing, elevation.
> 
> +axis=neu - northing, easting, up - useful for "lat/long" geographic coordinates, or south orientated transverse mercator.
> 
> +axis=wnu - westing, northing, up - some planetary coordinate systems have "west positive" coordinate systems

sensor_data_sharing_service/src/detection_offset_from_ref_converter.cpp
willjohnsonk: Now using `throw std::runtime_error("Failed to create PROJ context.")`; since proj does NULL returns instead of having exceptions

MishkaMN (approved):

paulbourelly999 (approved):

PR #458, opened 2026-09-04
fix: explicitly consume message payload size instead of guess in kafka

liuji1031:
# PR Details
## Description

This PR fixes the following issue:

`CARMAStreetsPlugin` fails to parse SDSM JSON consumed from the `v2xhub_sdsm_sub` topic. Every failure truncates the payload at 688 bytes, with the JSON parser reporting `Missing ',' or '}'` at column 687 and a trailing garbage byte (`P\x03`) where JSON should continue.

An example console log is as follows:
```bash
[2026-09-04 18:43:47.381] /kafka_consumer_worker.cpp (153) - DEBUG1 : rdkafka#consumer-5 message Consumed: 691 bytes : {"msg_cnt":122,"source_id":"rsu_1234","equipment_type":1,"sdsm_time_stamp":{"second":47374,"minute":43,"hour":18,"day":4,"month":9,"year":2026,"offset":0},"ref_pos":{"long":-771483512,"lat":389519791,"elevation":0},"ref_pos_xy_conf":{"semi_major":0,"semi_minor":0,"orientation":0},"objects":[{"detected_object_data":{"detected_object_common_data":{"obj_type":2,"object_id":13684,"obj_type_cfd":100,"measurement_time":87,"time_confidence":0,"pos":{"offset_x":3254,"offset_y":-99,"offset_z":0},"pos_confidence":{"pos":15,"elevation":15},"speed":0,"speed_confidence":7,"speed_z":0,"speed_confidence_z":4,"heading":0,"heading_conf":4},"detected_object_optional_data":{"detected_vru_data":{}}}}]}

[2026-09-04 18:43:47.381] src/CARMAStreetsPlugin.cpp (673) - DEBUG  : consumed message payload: {"msg_cnt":122,"source_id":"rsu_1234","equipment_type":1,"sdsm_time_stamp":{"second":47374,"minute":43,"hour":18,"day":4,"month":9,"year":2026,"offset":0},"ref_pos":{"long":-771483512,"lat":389519791,"elevation":0},"ref_pos_xy_conf":{"semi_major":0,"semi_minor":0,"orientation":0},"objects":[{"detected_object_data":{"detected_object_common_data":{"obj_type":2,"object_id":13684,"obj_type_cfd":100,"measurement_time":87,"time_confidence":0,"pos":{"offset_x":3254,"offset_y":-99,"offset_z":0},"pos_confidence":{"pos":15,"elevation":15},"speed":0,"speed_confidence":7,"speed_z":0,"speed_confidence_z":4,"heading":0,"heading_conf":4},"detected_object_optional_data":{"detected_vru_data":{}P

[2026-09-04 18:43:47.381] sonToJ3224SDSMConverter.cpp (17) - ERROR  : Parse error: * Line 1, Column 687
  Missing ',' or '}' in object declaration
```

## Root cause

`msg_consume()` returned the Kafka payload as a `const char*` and threw away `message->len()`.

Since the payload is handed downstream as a C string, whoever reads it walks the bytes until it hits a null character — which it won't, because Kafka payloads are not null terminated. So the walk keeps going past the end of the message and stops at whatever null character it happens to find in the memory that follows. Depending on where that null character sits, the message either comes out truncated or comes out longer than it really is, with unrelated bytes attached. If no null character turns up before the end of the mapped page, the process crashes instead.

The length is the only reliable record of where a Kafka message ends, and dropping it means every caller has to guess.

A second problem on the same path: `consume()` never deleted the `RdKafka::Message` it got back, so one message is leaked on every call for the life of the process.

## Changes

- `msg_consume()` now returns a `std::string` built from the payload pointer **and** `message->len()`, so the exact number of bytes is copied and nothing walks looking for a null character.
- `consume()` returns that `std::string` and holds the `RdKafka::Message` in a `std::unique_ptr`, so the message is released properly instead of leaked.
- Updated the `SPDLOG_TRACE` line that was printing the same unterminated pointer.
- Updated the `consume()` signature and its Doxygen comment; added `<memory>`.
- Updated `MOCK_METHOD(const char*, consume, ...)` → `std::string`.
- Updated `examples/kafka_examples`, the one caller that held the result as a `const char*` and checked it with `strlen`, to use `std::string` / `empty()`.

## Types of changes

## Checklist:


examples/kafka_examples/src/main.cpp
```diff
@@ -28,8 +28,8 @@ void call_consumer_thread()
         
         while (consumer_worker->is_running()) 
         {
-            const char* paylod= consumer_worker->consume(1000);
-            if(strlen(paylod) > 0)
+            const std::string paylod = consumer_worker->consume(1000);
```
MishkaMN: paylod typo

MishkaMN (changes requested):

MishkaMN:
Nit: we can try adding unit test but not requirement as we can test at production

examples/kafka_examples/src/main.cpp
liuji1031: fixed

MishkaMN (approved):

MishkaMN (approved):

PR #461, opened 2026-09-11
Try stop SDSM producer from dropping detections and publish at up to 10 Hz

MishkaMN:
# PR Details
## Description

Stops the sensor data sharing service's SDSM producer from dropping detections, and publishes SDSMs at up to 10 Hz as detections arrive:

1. **Rate-limited, detection-driven publishing.** The detection consumer notifies the producer after each detection. If the last SDSM was at least 100 ms ago (`SDSM_PUBLISH_PERIOD_MS`), the producer publishes right away; otherwise it holds the detections until 100 ms after the last SDSM and publishes everything that arrived meanwhile. No two SDSMs are less than 100 ms apart, and no detection waits for a fixed loop period.
image

2. **Atomic take of detections.** The producer swaps `detected_objects` out under the write lock and builds the SDSM from that snapshot (`create_sdsm` now takes the objects map), so detections consumed while an SDSM is built and sent are kept for the next SDSM.
3. A newer detection of an object replaces one still waiting to be published, as before; the replaced detection is now counted in `Detection Drop Count`.
4. The producer's wait is bounded to 100 ms, so `is_running()` and the detection metrics (written on a 1 s timer) are still checked when no detections arrive. The hold uses `streets_clock_singleton`, so it follows simulation time.
5. New tests: a detection consumed during a send is published in the next SDSM ≥ 100 ms later; a detection is published immediately after the period and held within it.

## Related GitHub Issue

N/A

## Related Jira Key

N/A

## Motivation and Context

The producer slept 100 ms *after* each cycle's work (~101 ms per cycle) next to a 10 Hz sensor, with nothing keeping them aligned. With detection arrival jitter, a cycle sometimes found no new detection and the next found two, and `detected_objects` keeps only the newest per object. A detection consumed between building the SDSM and `detected_objects.clear()` was also erased unsent.

Old logic (actual service output on `release/wrangler`) vs. new logic (replayed on the same detection consume times from the service log), for the 2026-09-10 data-verification test session, 13:00–14:05 ET, FLIR camera at 10 Hz:

| | Old | New |
|---|---|---|
| Detections never published, test session (6,541) | 4.8% | 0.6% |
| Detections never published, 4 engaged test runs (832) | 8.9% | 0.2% |
| SDSM rate while detecting | 9.5 Hz | 9.9 Hz |
| SDSM gaps of 100–110 ms | 95.6% | 98.9% |
| Skipped 100 ms slots (gap > 140 ms) | 4.4% | 0.7% |
| SDSM gaps under 100 ms | 0% | 0% |
| Detection age at publish, median / p95 | 66 / 113 ms | 49 / 97 ms |

## How Has This Been Tested?

- `sensor_data_sharing_service.cpp` and `sensor_data_sharing_service_test.cpp` compile (`g++ -std=c++17 -fsyntax-only` against the repo headers).
- The unit tests have **not** been run: the full build needs the carma-streets Docker toolchain, which wasn't available. CI on this branch is the check.
- The old-vs-new numbers compare the old service's logged SDSMs with a replay of the new publishing rule on the same logged consume times.

## Types of changes

## Checklist:

### DIFF
--- a/sensor_data_sharing_service/include/sensor_data_sharing_service.hpp
+++ b/sensor_data_sharing_service/include/sensor_data_sharing_service.hpp
@@ -50,6 +50,18 @@
 
 namespace sensor_data_sharing_service {
 
+    /**
+     * @brief Period in milliseconds of the SDSM producer loop. An SDSM is only sent in a cycle that has new detections,
+     * so with a 10 Hz sensor SDSMs are still produced at 10 Hz. Checking at twice the sensor rate makes sure every
+     * detection gets its own cycle despite detection arrival jitter; a 10 Hz loop running alongside a 10 Hz sensor
+     * sometimes finds no new detection in one cycle and two in the next, and drops the older one.
+     */
+    inline constexpr uint64_t SDSM_PRODUCER_PERIOD_MS = 50;
+    /**
+     * @brief Number of SDSM producer cycles between detection metrics writes (once per second).
+     */
+    inline constexpr unsigned int DETECTION_METRICS_WRITE_CYCLES = 1000 / SDSM_PRODUCER_PERIOD_MS;
+
     class sds_service : public streets_service::streets_service {
         private:
             /**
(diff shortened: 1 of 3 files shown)


MishkaMN:
Docker builds are failing currently only due to some network degradation issue on ubuntu.com since July, but the CI passes

liuji1031 (approved):

